PR #1, opened 2022-05-25
initial lib impl

maxjeffos:


JackuB (approved):
All good for a PoC! Comments I'd have would be solved by employing CLI (Cobra) framework

PR #4, opened 2022-10-14
Feat: Introduce workflow engine

PeterSchafer:
Implement the initial version of the WorkflowEngine

### DIFF
--- a/pkg/workflow/engine_test.go
+++ b/pkg/workflow/engine_test.go
@@ -1,0 +1,132 @@
+package workflow
+
+import (
+	"fmt"
+	"net/url"
+	"testing"
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/spf13/pflag"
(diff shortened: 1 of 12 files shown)


pkg/workflow/engine_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"net/url"
+	"testing"
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_EngineBasics(t *testing.T) {
+	configuration := configuration.New()
```
Avishagp, line 14: Can we maybe divide this function to make it shorter and clearer?

PeterSchafer: I can move out the callback functions which blow the test up that much.

Avishagp (approved):

PR #5, opened 2022-10-21
Chore/hmmr 576 introduce cobra

PeterSchafer:


pkg/configuration/configuration.go
```diff
@@ -107,7 +103,7 @@ func New() Configuration {
 }
 
 func (ev *extendedViper) Clone() Configuration {
-	fmt.Println("extendedViper.Clone() to implement")
+	//fmt.Println("extendedViper.Clone() to implement")
```
Avishagp, line 106: Should we remove this line?

PeterSchafer: yes I will, thanks!

PR #9, opened 2022-12-02
Integrate local sbom extension

PeterSchafer:
Adding the local sbom extension


internal/utils/determine_default_org.go
```diff
@@ -0,0 +1,48 @@
...
+		Id   string `json:"id,omitempty"`
+		Type string `json:"type,omitempty"`
+	}
+}
+
+func GetDefaultOrgID(client *http.Client, apiUrl string) (orgID string, err error) {
+	url := apiUrl + "/rest/self?version=" + constants.SNYK_API_VERSION
+	res, err := client.Get(url)
+	if err != nil {
+		return "", fmt.Errorf("unable to retrieve org ID: %w", err)
```
Avishagp, line 29: Should the org ID just reside in the error like that? What does err look like?

Avishagp: Resolved by explanation in DMs

Avishagp (approved):

internal/utils/determine_default_org.go
```diff
@@ -0,0 +1,48 @@
...
+			Name              string `json:"name,omitempty"`
+			Username          string `json:"username,omitempty"`
+		} `json:"attributes,omitempty"`
+		Id   string `json:"id,omitempty"`
+		Type string `json:"type,omitempty"`
+	}
+}
+
+func GetDefaultOrgID(client *http.Client, apiUrl string) (orgID string, err error) {
+	url := apiUrl + "/rest/self?version=" + constants.SNYK_API_VERSION
```
michelkaporin, line 26: @PeterSchafer out of curiosity, can you link me to the rest api spec for this endpoint? We had lots of problems in determining user org, wonder if there's an endpoint finally :-)

PR #12, opened 2022-12-13
fix: repeated header fields added

PeterSchafer:
In GetDefaultHeader() instead of a copy of the data structure a reference was used which caused header fields to get multiple times added.


pkg/networking/networking_test.go
```diff
@@ -30,10 +30,11 @@ func Test_GetDefaultHeader_WithAuth(t *testing.T) {
 		"Authorization": {"token " + token},
 	}
 
-	url, _ := url.Parse(config.GetString(configuration.API_URL))
-	actualHeader := net.GetDefaultHeader(url)
-
-	assert.Equal(t, expectedHeader, actualHeader)
+	for i := 0; i < 3; i++ {
```
Avishagp, line 33: Why the magic number 3?

PeterSchafer: I just wanted to run the method under test multiple times to ensure it behaves the same each time. 2 was to obvious, 3 felt reasonable. shall I add a comment?

Avishagp: Yeah, I'd do that just to make it clear

Avishagp (approved):

PR #15, opened 2023-01-01
[HMMR-620] Take both org-ID and slug name

asaf92:


pkg/app/app.go
```diff
@@ -38,23 +39,25 @@ func initConfiguration(config configuration.Configuration) {
...
 
-	config.AddDefaultValue(configuration.ORGANIZATION, func(existingValue interface{}) interface{} {
-		useDefaultValue := true
-		orgid := existingValue
-		if existingValue != nil {
-			if len(existingValue.(string)) > 0 {
-				useDefaultValue = false
+	config.AddDefaultValue(configuration.ORGANIZATION, func(existingValue any) any {
+		client := networking.NewNetworkAccess(config).GetHttpClient()
+		apiClient := networking.NewApi(configuration.API_URL, client)
```
PeterSchafer, line 44: chore: `configuration.API_URL` is a key and must actually be used like this `config.GetString(configuration.API_URL)` to actually retrieve the URL, which is a configurable value. See also https://github.com/snyk/go-application-framework/pull/15/files#diff-63facfa8b028c9f397524787fac074737e5534ae2591640c1a029549ec3d7450L51

pkg/networking/api.go
```diff
@@ -0,0 +1,75 @@
+package networking
```
PeterSchafer, line 1: suggestion: I would like to move this file into "internal" since I'm not sure yet how we want to support developers that use this library to access the API. So instead of prematurely committing to a public API (pkg) I would rather use it as an internal util.

PeterSchafer (commented):
Suggestion: since we are missing tests for app.go, one possibility is to either add them or create a branch of the CLI and link this branch there and test manually inside the CLI.

pkg/app/app.go
asaf92: Thanks! I'll fix this

PeterSchafer (approved):

PR #20, opened 2023-01-31
enrich documentation [HEAD-35]

j-luong:
**What does this PR do?**
This PR should enrich the documentation for all packages under the public API`pkg/`

### DIFF
--- a/pkg/workflow/invocationcontextimpl.go
+++ b/pkg/workflow/invocationcontextimpl.go
@@ -8,6 +8,7 @@ import (
 	"github.com/snyk/go-application-framework/pkg/networking"
 )
 
+// InvocationContextImpl is the default implementation of the InvocationContext interface.
 type InvocationContextImpl struct {
 	WorkflowID     Identifier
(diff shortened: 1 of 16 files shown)


pkg/workflow/invocationcontextimpl.go
```diff
@@ -17,26 +18,32 @@ type InvocationContextImpl struct {
 	logger         *log.Logger
 }
 
+// GetWorkflowIdentifier returns the identifier of the workflow that is being invoked.
 func (ici *InvocationContextImpl) GetWorkflowIdentifier() Identifier {
 	return ici.WorkflowID
 }
 
+// GetConfiguration returns the configuration that is being used by the workflow engine.
```
PeterSchafer, line 26: nitpick: here and in the following docs instead of
> ... that is being used by the workflow engine.

using 

> ... that can be used inside the workflow.

would be more accurate.

PeterSchafer (approved):
Great work on documenting the API, it all looks really good! I just left one optional comment.

PR #21, opened 2023-02-03
add whoami workflow [HEAD-35]

j-luong:
**What does this PR do?**
This PR should introduce a new localworkflow hidden behind the experimental flag, it should allow the CLI to call the user/me endpoint.


pkg/local_workflows/whoami_workflow.go
```diff
@@ -0,0 +1,142 @@
...
+	// return full payload if json flag is set
+	if config.GetBool(jsonFlag) {
+		// parse response
+		userMeData := createWorkflowData(userMe)
+
+		// return userme data
+		return []workflow.Data{userMeData}, err
+	}
+
+	userData := createWorkflowData(user)
```
PeterSchafer, line 83: In this case the mimetype should be `plain/text` rather than `application/json`.

j-luong: good catch, fixed :)

pkg/local_workflows/whoami_workflow_test.go
```diff
@@ -0,0 +1,265 @@
...
+	t.Run("returns user name", func(t *testing.T) {
+		// setup
+		expectedResponse := "user.name@snyk.io"
+
+		// execute
+		output, err := whoAmIWorkflowEntryPoint(invocationContextMock, nil)
+
+		// assert
+		assert.Nil(t, err)
+		assert.Equal(t, expectedResponse, output[0].GetPayload())
```
PeterSchafer, line 115: Suggestion: Maybe test the mime-type here and below.

pkg/local_workflows/whoami_workflow_test.go
```diff
@@ -0,0 +1,265 @@
...
+	assert.Equal(t, expectedError.Error(), err.Error())
+}
+
+func Test_WhoAmI_whoAmIWorkflowEntryPoint_happyPath(t *testing.T) {
+	// setup
+	logger := log.New(os.Stderr, "test", 0)
+	config := configuration.New()
+	config.Set("experimental", true)
+
+	payload := `{
```
PeterSchafer, line 65: nitpick: from a maintainability aspect it is good not to share knowledge if not necessary, this means you could consider using the api contract data structure to create the json payload.

PeterSchafer (approved):
Looks great!
Same as before, please squash before merging :)

PR #25, opened 2023-02-10
feat: add OAuth authentication [HEAD-110][HEAD-111][HEAD-112]

PeterSchafer:


internal/api/urls.go
```diff
@@ -0,0 +1,49 @@
...
+
+func GetCanonicalApiUrl(userDefinedUrl string) (string, error) {
+	result := ""
+	url, err := url.Parse(userDefinedUrl)
+	if err != nil {
+		return result, err
+	}
+
+	// for localhost we don't change the host, since there are no subdomains
+	if !strings.Contains(url.Host, "localhost") {
```
bastiandoetsch, line 17: Maybe you could invert the if statement to make it a guard and easier to follow the flow?

internal/api/urls.go
```diff
@@ -0,0 +1,49 @@
...
+	url.Path = ""
+	url.Fragment = ""
+	url.RawQuery = ""
+
+	result = url.String()
+
+	return result, nil
+}
+
+func DeriveAppUrl(canocialUrl string) (string, error) {
```
bastiandoetsch, line 37: `canonicalUrl` ;)

internal/api/urls.go
```diff
@@ -0,0 +1,49 @@
...
+	result := ""
+	url, err := url.Parse(userDefinedUrl)
+	if err != nil {
+		return result, err
+	}
+
+	// for localhost we don't change the host, since there are no subdomains
+	if !strings.Contains(url.Host, "localhost") {
+		appRegexp, _ := regexp.Compile("^app\\.")
+		url.Host = appRegexp.ReplaceAllString(url.Host, "api.")
```
bastiandoetsch, line 19: Looks like a candidate for a global constant

internal/api/urls.go
```diff
@@ -0,0 +1,49 @@
...
+func GetCanonicalApiUrl(userDefinedUrl string) (string, error) {
+	result := ""
+	url, err := url.Parse(userDefinedUrl)
+	if err != nil {
+		return result, err
+	}
+
+	// for localhost we don't change the host, since there are no subdomains
+	if !strings.Contains(url.Host, "localhost") {
+		appRegexp, _ := regexp.Compile("^app\\.")
```
bastiandoetsch, line 18: Looks like a candidate for a global constant

internal/api/urls.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func DeriveAppUrl(canocialUrl string) (string, error) {
+	result := ""
+	url, err := url.Parse(canocialUrl)
+	if err != nil {
+		return result, err
+	}
+
+	apiRegexp, _ := regexp.Compile("^api\\.")
```
bastiandoetsch, line 44: Looks like a candidate for a global constant

internal/constants/constants.go
```diff
@@ -1,4 +1,4 @@
 package constants
 
-const SNYK_DEFAULT_API_URL = "https://app.snyk.io/api"
+const SNYK_DEFAULT_API_URL = "https://api.snyk.io"
```
bastiandoetsch, line 3: authentication from the IDEs always failed when not using `/api`. Are we sure, this is the correct default api?

pkg/app/app.go
```diff
@@ -27,18 +25,26 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
...
-				if apiUrl, err := url.Parse(temp); err == nil {
-					apiUrl.Path = strings.Replace(apiUrl.Path, "/v1", "", 1)
-					apiString = apiUrl.String()
-				}
+				urlString = temp
 			}
-			return apiString
 		}
+
+		apiString, _ := api.GetCanonicalApiUrl(urlString)
```
bastiandoetsch, line 36: why is the error ignored here?

pkg/app/app.go
```diff
@@ -27,18 +25,26 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
...
+		return apiString
+	})
+
+	config.AddDefaultValue(configuration.WEB_APP_URL, func(existingValue any) any {
+		canonicalApiUrl := config.GetString(configuration.API_URL)
+		appUrl, _ := api.DeriveAppUrl(canonicalApiUrl)
```
bastiandoetsch, line 42: same here: why is the error ignored?

pkg/app/app.go
```diff
@@ -27,18 +25,26 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
...
+		return appUrl
+	})
+
+	config.AddDefaultValue(configuration.OAuthTokenUrl, func(existingValue any) any {
+		return "https://app.fedramp-alpha.snykgov.io"
```
bastiandoetsch, line 47: Most likely, this should be a constant

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+	"time"
+
+	"github.com/pkg/browser"
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"golang.org/x/oauth2"
+)
+
+const (
+	CONFIG_KEY_OAUTH_TOKEN string = "OAUTH_TOKEN"
+	OAUTH_CLIENT_ID        string = "0oa37b7oa3zOoDWCe4h7"
```
bastiandoetsch, line 24: Is this something that maybe should be generated at runtime, or injected at compile time?

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+
+func getRedirectUri(port int) string {
+	callback := fmt.Sprintf("http://%s:%d%s", CALLBACK_HOSTNAME, port, CALLBACK_PATH)
+	return callback
+}
+
+func getOAuthConfiguration(config configuration.Configuration) *oauth2.Config {
+
+	appUrl := config.GetString(configuration.WEB_APP_URL)
+	// tokenUrl := strings.Replace(appUrl, "app.", "id.", 1) + "/oauth2/default/v1/token"
```
bastiandoetsch, line 70: Can this comment be removed? source control will get it back for us, if needed.

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	return conf
+}
+
+func getCodeChallenge(verifier []byte) string {
+	shasum := sha256.Sum256(verifier)
+	return base64.RawURLEncoding.EncodeToString(shasum[:])
+}
+
+func createVerifier(count int) []byte {
```
bastiandoetsch, line 89: this function is a mystery to me...

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+		index := rand.Int() % len(lut)
+		verifier[i] = lut[index]
+	}
+
+	return verifier
+}
+
+func getToken(config configuration.Configuration) (*oauth2.Token, error) {
+	oauthTokenString := config.GetString(CONFIG_KEY_OAUTH_TOKEN)
+	if len(oauthTokenString) > 0 {
```
bastiandoetsch, line 109: I would invert this if and just return if len <= 0 and then do stuff if it's not the case.

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+	if o.headless {
+		// TODO:
+	} else {
+		srv := &http.Server{}
+
+		http.HandleFunc(CALLBACK_PATH, func(w http.ResponseWriter, r *http.Request) {
+			responseError = html.EscapeString(r.URL.Query().Get("error"))
+			if len(responseError) > 0 {
+				details := html.EscapeString(r.URL.Query().Get("error_description"))
+				fmt.Fprintf(w, "Error during authentication! (%s)\n%s", responseError, details)
```
bastiandoetsch, line 158: should this maybe use a logging framework?

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+	if state != responseState {
+		return fmt.Errorf("incorrect response state: %s != %s", responseState, state)
+	}
+
+	// Use the custom HTTP client when requesting a token.
+	if o.httpClient != nil {
+		ctx = context.WithValue(ctx, oauth2.HTTPClient, o.httpClient)
+	}
+
+	token, err := o.oauthConfig.Exchange(ctx, responseCode, oauth2.SetAuthURLParam("code_verifier", string(verifier)))
```
bastiandoetsch, line 209: Still not sure what the verifier does - is it a nonce/challenge? Would a UUID suffice here?

pkg/app/app.go
PeterSchafer: I think the configuration value "OAuthTokenUrl" is obsolete, since it can be derived from the web_app_url. It used to be temporarily hardcoded in the oauthauthenticator since the api doesn't support the correct endpoints yet. but there is no need to have an extra config value for it.

pkg/auth/oauth2authenticator.go
PeterSchafer: This is writing to the local web page that is showing either authentication successful or an error message. We actually need to check with Product if we need a more/designed webpage here.
But yes, in addition, we could log this into the debug logs.

pkg/auth/oauth2authenticator.go
PeterSchafer: The verifier and how it has to be generated is defined in the standard. It is based on a secret (called code) and it is used at the end of the flow by both parties to verify that the final request comes from the originating party.

pkg/networking/middleware/auth_header.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func (n *AuthHeaderMiddleware) RoundTrip(request *http.Request) (*http.Response, error) {
+	if request.URL != nil {
+		// determine configured api url
+		apiUrlString := n.config.GetString(configuration.API_URL)
+		apiUrl, err := url.Parse(apiUrlString)
+		if err != nil {
+			apiUrl, _ = url.Parse(constants.SNYK_DEFAULT_API_URL)
```
bastiandoetsch, line 37: Not sure, we should do this. if somebody configures an endpoint e.g. MT-EU, and they are accidentally re-routed to MT-US, that's not expected and regarding GDPR may lead to questions.

pkg/networking/networking.go
```diff
@@ -23,55 +23,43 @@ const (
...
-
-	return request
 }
 
 // RoundTrip is an implementation of the http.RoundTripper interface.
-func (crt *customRoundtripper) RoundTrip(request *http.Request) (*http.Response, error) {
-	request = crt.decorateRequest(request)
-	return crt.encapsulatedRoundtripper.RoundTrip(request)
+func (crt *customRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	_ = crt.networkAccess.AddDefaultHeader(request)
```
bastiandoetsch, line 61: Why do we ignore errors here? Shouldn't an error adding a header be handled?

pkg/networking/networking.go
```diff
@@ -183,3 +174,21 @@ func (n *NetworkImpl) AddRootCAs(pemFileLocation string) error {
 
 	return err
 }
+
+func (n *NetworkImpl) GetAuthenticator() auth.Authenticator {
+	if n.authenticator == nil {
+		// rt := n.configureRoundTripper(http.DefaultTransport.(*http.Transport))
```
bastiandoetsch, line 180: I guess, this can go away

pkg/auth/oauth2authenticator.go
bastiandoetsch: Then there should be comment with todo/fixme explaining that this is the code to be used in the future :)

pkg/auth/oauth2authenticator.go
bastiandoetsch: most likely 😂

internal/constants/constants.go
PeterSchafer: interesting, there was some confusion around the api url, which is why we consulted with the API teams. but worth testing,

pkg/auth/authenticator.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/snyk/go-application-framework/pkg/configuration"
+)
+
+type Authenticator interface {
+	Authenticate() error
+	AddAuthenticationHeader(request *http.Request) error
+}
+
+func CreateAuthenticator(config configuration.Configuration, httpClient *http.Client) Authenticator {
+	useOAuth := config.GetBool(configuration.OAuthEnabled)
```
PeterSchafer, line 15: Question: I'm not sure this will do what is required, the reason why I'm not sure is, is that I don't understand where OAuthEnabled is being set.

pkg/auth/oauth2authenticator.go
PeterSchafer: It's something that is configured/defined by the auth provider, we can inject it during compile time if we like. hardcoding for now was just the most pragmatic way.

pkg/auth/oauth2authenticator.go
PeterSchafer: I just checked, we return this error if it occurs and it can be logged by the caller

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,252 @@
...
+
+		http.HandleFunc(CALLBACK_PATH, func(w http.ResponseWriter, r *http.Request) {
+			responseError = html.EscapeString(r.URL.Query().Get("error"))
+			if len(responseError) > 0 {
+				details := html.EscapeString(r.URL.Query().Get("error_description"))
+				fmt.Fprintf(w, "Error during authentication! (%s)\n%s", responseError, details)
+			} else {
+				responseCode = html.EscapeString(r.URL.Query().Get("code"))
+				responseState = html.EscapeString(r.URL.Query().Get("state"))
+				fmt.Fprintf(w, "Succesfully Authenticated!")
```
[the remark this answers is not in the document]
PeterSchafer: this message is logged already

internal/api/urls.go
asaf92: +1

pkg/app/app.go
asaf92: It was hard coded, but I wanted a way to control it in the tests so I added a configuration

pkg/auth/auth_test.go
asaf92, line 1: Remove this

pkg/workflow/engineimpl.go
```diff
@@ -88,11 +88,6 @@ func (e *EngineImpl) Init() error {
 		e.analytics.SetIntegration(e.config.GetString(configuration.INTEGRATION_NAME), e.config.GetString(configuration.INTEGRATION_VERSION))
 		e.analytics.SetApiUrl(e.config.GetString(configuration.API_URL))
 		e.analytics.SetOrg(e.config.GetString(configuration.ORGANIZATION))
-		e.analytics.AddHeader(func() http.Header {
-			url := e.config.GetUrl(configuration.API_URL)
-			header := e.networkAccess.GetDefaultHeader(url)
-			return header
-		})
```
asaf92, line 95: Is this ok?

pkg/networking/middleware/auth_header.go
PeterSchafer: there won't be re-routing based on this code but you are right, this is not the place to do the check, it already happens in the configuration.

pkg/app/app.go
PeterSchafer: Unfortunately we currently don't have error reporting or logging capabilities here. I would like to add this in a new PR if this is okay with you.

pkg/app/app.go
PeterSchafer: see above

pkg/networking/networking.go
asaf92: I looked at the code and we never return an error, so I refactored the error return value out of it

pkg/configuration/storage.go
```diff
@@ -0,0 +1,46 @@
...
+func (s *JsonStorage) Set(key string, value any) error {
+	// Check if path to file exists
+	err := os.MkdirAll(filepath.Dir(s.path), utils.FILEPERM_755)
+	if err != nil {
+		return err
+	}
+
+	fileBytes, _ := os.ReadFile(s.path)
+	config := make(map[string]any)
+	_ = json.Unmarshal(fileBytes, &config)
```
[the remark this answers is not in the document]
asaf92: yeah good point, I'll add error handling

pkg/app/app.go
asaf92: +1
Also I don't know if there's a way to recover from this error, so it's just a matter of failing/logging an error earlier

pkg/networking/networking.go
```diff
@@ -1,77 +1,63 @@
...
 // NetworkAccess is the interface for network access.
 type NetworkAccess interface {
-	// GetDefaultHeader returns the default header for a given URL.
-	GetDefaultHeader(url *url.URL) http.Header
-	// GetRoundtripper returns the http.Roundtripper.
-	GetRoundtripper() http.RoundTripper
+	// AddHeaders adds all the custom and authentication headers to the request.
+	AddHeaders(request *http.Request) error
+	// AddDefaultHeader adds the default headers request.
+	AddDefaultHeader(request *http.Request)
```
PeterSchafer, line 26: Question: It appears we only use this method internally. How about removing it from the public API to keep the API slim?

PeterSchafer: this would have the benefit we wouldn't need to explain to anyone the difference between AddHeader() and AddDefaultHeader()

internal/api/urls_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var instanceList = []string{
+	"snyk.io",
+	"fedramp-alpha.snykgov",
```
bastiandoetsch, line 11: `.io` would be charming

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,287 @@
...
+	httpClient         *http.Client
+	config             configuration.Configuration
+	oauthConfig        *oauth2.Config
+	token              *oauth2.Token
+	headless           bool
+	openBrowserFunc    func(authUrl string)
+	shutdownServerFunc func(server *http.Server)
+}
+
+func (o *oAuth2Authenticator) AddEnvironmentVariables(env []string) ([]string, error) {
```
bastiandoetsch, line 47: We may want to refactor this in a later PR - I don't think this should be here. Let's add a todo and move on for now.

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,287 @@
...
+func createVerifier(count int) []byte {
+	/*
+	  unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
+	   ALPHA = %x41-5A / %x61-7A
+	   DIGIT = %x30-39
+	*/
+	lut := []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~")
+	verifier := make([]byte, count)
+
+	// TODO is this good enough?
```
bastiandoetsch, line 122: of course!

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+		return token, nil
+	}
+	return nil, nil
+}
+
+func NewOAuth2Authenticator(config configuration.Configuration, httpClient *http.Client) Authenticator {
+	token, _ := GetOAuthToken(config)
+	oauthConfig := getOAuthConfiguration(config)
+	config.PersistInConfigFile(CONFIG_KEY_OAUTH_TOKEN)
```
bastiandoetsch, line 148: this is a side-effect - would it be possible to do that before / after the instantiation of the authenticator?

pkg/auth/oauth2authenticator.go
```diff
@@ -0,0 +1,287 @@
...
+		httpClient:         httpClient,
+		config:             config,
+		oauthConfig:        oauthConfig,
+		token:              token,
+		openBrowserFunc:    openBrowser,
+		shutdownServerFunc: shutdownServer,
+	}
+}
+
+func (o *oAuth2Authenticator) IsSupported() bool {
```
bastiandoetsch, line 160: let's rename to `isActive`

pkg/auth/oauth2authenticator_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	config := configuration.New()
+	config.Set(configuration.WEB_APP_URL, webapp)
+
+	oauthConfig := getOAuthConfiguration(config)
+
+	assert.Equal(t, "", oauthConfig.RedirectURL)
+	assert.Equal(t, OAUTH_CLIENT_ID, oauthConfig.ClientID)
+	assert.Equal(t, webapp+"/oauth/authorize", oauthConfig.Endpoint.AuthURL)
+	// assert.Equal(t, "https://id.fedramp-alpha.snykgov.io/oauth2/default/v1/token", oauthConfig.Endpoint.TokenURL)
```
bastiandoetsch, line 76: do we need this commented code?

pkg/configuration/configuration.go
```diff
@@ -40,6 +43,11 @@ type extendedViper struct {
 	viper           *viper.Viper
 	alternativeKeys map[string][]string
 	defaultValues   map[string]DefaultValueFunction
+
+	// persistedKeys stores the keys that need to be persisted to storage when Set is called.
+	// Only specific keys are persisted, so viper's native functionality is not used.
+	persistedKeys map[string]bool
```
bastiandoetsch, line 49: `toBePersistedKeys`

pkg/local_workflows/local_workflows.go
```diff
@@ -5,14 +5,15 @@ import (
...
-// localworkflows are initialised when create a new workflow engine via app.CreateAppEngine()
+// localworkflows are initialized when create a new workflow engine via app.CreateAppEngine()
 func Init(engine workflow.Engine) error {
 	var err error
 
 	initMethods := []func(workflow.Engine) error{
 		InitDepGraphWorkflow,
 		InitOutputWorkflow,
 		InitWhoAmIWorkflow,
+		InitAuth, //Use legacy CLI for authentication for now, until OAuth is ready
```
bastiandoetsch, line 16: maybe add `TODO` to get a marker in the IDE

pkg/mocks/networking.go
```diff
@@ -35,6 +35,20 @@ func (m *MockNetworkAccess) EXPECT() *MockNetworkAccessMockRecorder {
 	return m.recorder
```
bastiandoetsch, line 35: May be good to have the mock in its own file, to not mix test code with prod code

pkg/local_workflows/local_workflows.go
asaf92: This is a stale comment. The entire line was commented out

pkg/networking/networking.go
```diff
@@ -1,77 +1,63 @@
...
 
 // NetworkImpl is the default implementation of the NetworkAccess interface.
 type NetworkImpl struct {
-	config       configuration.Configuration
-	userAgent    string
-	staticHeader http.Header
-	logger       *log.Logger
-	proxy        func(req *http.Request) (*url.URL, error)
-	caPool       *x509.CertPool
+	config        configuration.Configuration
```
bastiandoetsch, line 43: struct could be private

bastiandoetsch (approved):

PR #26, opened 2023-02-16
feat: Handle additional mime types in output module

PeterSchafer:
This PR adds support for unknown mime types which are convertible to string by default, so that they are displayed on the console as if they would be text/plain.

### DIFF
--- a/pkg/local_workflows/output_workflow.go
+++ b/pkg/local_workflows/output_workflow.go
@@ -64,12 +64,21 @@ func outputWorkflowEntryPoint(invocation workflow.InvocationContext, input []wor
 				outputDestination.Remove(writeJsonToFile)
 				outputDestination.WriteFile(writeJsonToFile, singleData, iUtils.FILEPERM_666)
 			}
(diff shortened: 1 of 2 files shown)


pkg/local_workflows/output_workflow.go
```diff
@@ -64,12 +64,22 @@ func outputWorkflowEntryPoint(invocation workflow.InvocationContext, input []wor
...
+			// try to convert payload to a string
+			singleDataAsString := ""
+			singleData, typeCastSuccessful := input[i].GetPayload().([]byte)
+			if !typeCastSuccessful {
+				singleDataAsString, typeCastSuccessful = input[i].GetPayload().(string)
+			} else {
+				singleDataAsString = string(singleData)
+			}
+
+			if typeCastSuccessful {
```
bastiandoetsch, line 77: It would be slightly easier to understand if that handling were inside the first if/else. But no blocker.

PeterSchafer: Done :)

PR #34, opened 2023-03-27
fix: appropriately set default headers if undefined

jlourenc:
Append default headers to requests if not set.

### DIFF
--- a/pkg/networking/networking.go
+++ b/pkg/networking/networking.go
@@ -28,7 +28,7 @@ type NetworkAccess interface {
 	// GetUnauthorizedHttpClient returns an HTTP client that does not use authentication headers.
 	GetUnauthorizedHttpClient() *http.Client
 	// AddHeaderField adds a header field to the default header.
-	AddHeaderField(key string, value string)
+	AddHeaderField(key, value string)
 	// AddRootCAs adds the root CAs from the given PEM file.
 	AddRootCAs(pemFileLocation string) error
 	// GetAuthenticator returns the authenticator.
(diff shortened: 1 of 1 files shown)


pkg/networking/networking.go
```diff
@@ -61,45 +61,35 @@ func (crt *customRoundTripper) RoundTrip(request *http.Request) (*http.Response,
...
 		proxy:        http.ProxyFromEnvironment,
 	}
-
-	return &c
 }
 
-func (n *networkImpl) AddHeaderField(key string, value string) {
-	n.staticHeader[key] = append(n.staticHeader[key], value)
+func (n *networkImpl) AddHeaderField(key, value string) {
+	n.staticHeader.Add(key, value)
```
jlourenc, line 77: Safer to use the `http.Header` API as it takes care of converting keys to their canonical form.

pkg/networking/networking.go
```diff
@@ -61,45 +61,35 @@ func (crt *customRoundTripper) RoundTrip(request *http.Request) (*http.Response,
...
-
 	// iterate over default headers and add them if there is no existing entry yet
-	for k, v := range defaultHeader {
-		for i := range v {
-			request.Header.Set(k, v[i])
+	for k, v := range n.staticHeader.Clone() {
+		if request.Header.Values(k) == nil {
+			for i := range v {
+				request.Header.Add(k, v[i])
+			}
```
jlourenc, line 92: I believe this logic better aligns with the comment above. Though, the comment may be out-of date so please take a look.

PeterSchafer: Yes, the comment is actually out of date. We changed it here to actually use set, so that the framework overrides header that might already be in the request. One example are headers that are set by the legacy CLI and need to be replaced by the Extensible CLI.

jlourenc: Cool, then I think the key must be deleted if the static headers specify it, and then add header values to be correct.

PeterSchafer: Question: Is there a reason to use del() and add() instead of set()?

jlourenc: Yes, you can have multiple header values for a same header, for instance: `Cookies` or `Set-Cookies` is an example.
If you happen to have multiple `Cookies` headers in your `staticHeader` object, only the last `Cookie` header would remain using the `Set` method.

PeterSchafer (approved):

PR #35, opened 2023-03-27
chore: handle token specified via env var [HEAD-156]

j-luong:
This PR should ensure that api token environment variables override what is stored in a configuration file of the CLI.

This should mean that the OAuth flow should be skipped if an OAuth token env var is provided, even if the OAuth flow feature flag is enabled in the config.

### DIFF
--- a/pkg/configuration/constants.go
+++ b/pkg/configuration/constants.go
@@ -16,5 +16,6 @@ const (
 	WORKFLOW_USE_STDIO             string = "wflstdio"
 	RAW_CMD_ARGS                   string = "raw_cmd_args"
 	WEB_APP_URL                    string = "internal_snyk_app"
-	OAUTH_AUTH_ENABLED             string = "internal_snyk_oauth_enabled"
+	// feature flag to enable the new oAuth user flow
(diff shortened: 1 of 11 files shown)


pkg/configuration/constants.go
```diff
@@ -16,5 +16,6 @@ const (
 	WORKFLOW_USE_STDIO             string = "wflstdio"
 	RAW_CMD_ARGS                   string = "raw_cmd_args"
 	WEB_APP_URL                    string = "internal_snyk_app"
-	OAUTH_AUTH_ENABLED             string = "internal_snyk_oauth_enabled"
+	// feature flag to enable the new oAuth user flow
+	OAUTH_AUTH_FLOW_ENABLED string = "internal_snyk_oauth_enabled"
```
PeterSchafer, line 20: Suggestion: Since we are already renaming this, should we add `FF_` as a prefix?

j-luong: done

bastiandoetsch: I guess I need to change the feature flag name then in LS, too?

PR #39, opened 2023-03-31
feat: Improve analytics filtering

PeterSchafer:
Before sending Analytics data, we redact sensitive data. With a new additional command line argument format we support, we need to adapt the filtering of this format as well `--key value`. In addition this PR adds some more sensitive field values to ensure that everything is captured.

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -113,11 +113,12 @@ var (
 	// sensitiveFieldNames is a list of field names that should be sanitized.
 	// data sanitization is used to prevent sensitive data from being sent to the analytics server.
 	sensitiveFieldNames []string = []string{
-		"tfc-token",
-		"azurerm-account-key",
-		"fetch-tfstate-headers",
-		"username",
-		"password",
+		"headers",
+		"user",
+		"passw",
+		"token",
+		"key",
+		"secret",
 	}
 )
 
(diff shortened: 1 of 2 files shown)


pkg/analytics/analytics.go
```diff
@@ -290,10 +291,12 @@ func (a *AnalyticsImpl) Send() (*http.Response, error) {
 // * key : "value"
 // * key = "value"
 // * key = value
+// * key "value"
+
 func SanitizeValuesByKey(keysToFilter []string, replacementValue string, content []byte) ([]byte, error) {
 	for i := range keysToFilter {
 		filter := keysToFilter[i]
-		r, err := regexp.Compile("(?i)([\"']?\\w*" + filter + "\\w*\"?)(((\\s?[:]\\s?[\"'])[^\n\"']*([\"']))|((\\s?[=]\\s?[\"']?)[^\n\"']*([\"']?)))")
+		r, err := regexp.Compile("(?i)([\"']?\\w*" + filter + "\\w*\"?)(((\\s?[:]\\s?[\"'])[^\n\"']*([\"']))|((\\s?[= ,]\\s?[\"']?)[^\n\"']*([\"']?)))")
```
bastiandoetsch, line 299: what does the regexp change do?

PeterSchafer: It supports different separators between key and value, before the only separator was `=` now we support the following pattern:
* key = value
* key , value
* key value

pkg/analytics/analytics.go
```diff
@@ -113,11 +113,12 @@ var (
 	// sensitiveFieldNames is a list of field names that should be sanitized.
 	// data sanitization is used to prevent sensitive data from being sent to the analytics server.
 	sensitiveFieldNames []string = []string{
-		"tfc-token",
-		"azurerm-account-key",
-		"fetch-tfstate-headers",
-		"username",
-		"password",
+		"headers",
```
michelkaporin, line 116: Is this meant to be an exhaustive list of all field names that could be potentially sensitive? If yes, it's practically infeasible to predict what kind of arguments could be sensitive. 

Why not to sanitise all arguments that are not registered with the engine instead, as a potential solution?

PeterSchafer: Good point, instead of blacklist, use a whitelist.

In general for CMD arguments that might be possible, even if we currently don't have a list of all known/supported args for the legacy CLI. CMD arguments are just one example, the sanitization function works on a more abstract level, the json data structure that is send. Analytics have additional fields that can be filled by Extensions or commands that are product line specific. These will currently also be sanitized, as well as the usernames in file system paths.

bastiandoetsch: I think that this would be out-of-scope of the PR. We should re-think logging & analytics for our code-bases, though.

bastiandoetsch (approved):

PR #45, opened 2023-04-06
fix: always read additional CA certs

PeterSchafer:
Before not every request made use of additionally specified CA certs. Now every request checks them.


pkg/app/app.go
```diff
@@ -19,7 +19,10 @@ import (
 
 // initConfiguration initializes the configuration with initial values.
 func initConfiguration(config configuration.Configuration, apiClient api.ApiClient, logger *log.Logger) {
-	dir, _ := utils.SnykCacheDir()
+	dir, err := utils.SnykCacheDir()
+	if err != nil {
```
bastiandoetsch, line 23: nice :)

pkg/app/app.go
```diff
@@ -44,7 +47,11 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
 
 	config.AddDefaultValue(configuration.WEB_APP_URL, func(existingValue any) any {
 		canonicalApiUrl := config.GetString(configuration.API_URL)
-		appUrl, _ := api.DeriveAppUrl(canonicalApiUrl)
+		appUrl, err := api.DeriveAppUrl(canonicalApiUrl)
+		if err != nil {
+			logger.Println("Failed to determine default value for \"WEB_APP_URL\":", err)
```
bastiandoetsch, line 52: Same here - should we exit with RC=2 as this must not happen?

pkg/app/app.go
```diff
@@ -58,15 +65,20 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
 			isSlugName := err != nil
 			if isSlugName {
 				orgId, err = apiClient.GetOrgIdFromSlug(existingValue.(string))
-				if err == nil {
+				if err != nil {
```
bastiandoetsch, line 68: The returns could probably be simplified (no blocker though)

pkg/configuration/constants.go
```diff
@@ -17,7 +17,9 @@ const (
 	WORKFLOW_USE_STDIO             string = "wflstdio"
 	RAW_CMD_ARGS                   string = "raw_cmd_args"
 	WEB_APP_URL                    string = "internal_snyk_app"
+	INPUT_DIRECTORY                string = "targetDirectory"
+	ADD_TRUSTED_CA_FILE            string = "internal_additional_trusted_ca_file" // pem file location containing additional CAs to trust
```
bastiandoetsch, line 21: If this description is correct, the initialization with the env var name is not correct - it should be resolved before being added.

pkg/networking/networking.go
```diff
@@ -65,12 +65,24 @@ func NewNetworkAccess(config configuration.Configuration) NetworkAccess {
...
 
-	return &networkImpl{
+	n := &networkImpl{
 		config:       config,
 		staticHeader: http.Header{},
 		logger:       logger,
 		proxy:        http.ProxyFromEnvironment,
 	}
+
+	extraCaCertFile := config.GetString(configuration.ADD_TRUSTED_CA_FILE)
```
bastiandoetsch, line 75: This was initialized with a `[]string`, so a getString may not be correct?

pkg/networking/networking.go
```diff
@@ -65,12 +65,24 @@ func NewNetworkAccess(config configuration.Configuration) NetworkAccess {
...
+	if len(extraCaCertFile) > 0 {
+		err := n.AddRootCAs(extraCaCertFile)
+		if err != nil {
+			logger.Printf("Failed to AddRootCAs from '%s' (%v)\n", extraCaCertFile, err)
```
bastiandoetsch, line 79: Should this be a fatal error? If something is configured, but cannot be added, it feels like it should fail.

bastiandoetsch (approved):

pkg/app/app.go
PeterSchafer: We are already trying to create it inside SnykCacheDir(). I'm careful with exiting since it might be a too aggressive approach and might break things unexpectedly. I think step one is logging to increase transparency.

pkg/app/app.go
PeterSchafer: I'm careful with exiting since it might be a too aggressive approach and might break things unexpectedly. I think step one is logging to increase transparency.

pkg/networking/networking.go
PeterSchafer: I'm careful with exiting since it might be a too aggressive approach and might break things unexpectedly. I think step one is logging to increase transparency.

PR #46, opened 2023-04-09
feat: add `auth refresh` command for oauth 2 token refresh

asaf92:


pkg/local_workflows/authworkflow/refresh.go
```diff
@@ -0,0 +1,46 @@
...
+func refreshEntryPoint(invocationCtx workflow.InvocationContext, _ []workflow.Data) (_ []workflow.Data, err error) {
+	config := invocationCtx.GetConfiguration()
+	oauthEnabled := config.GetString(authTypeParameter) == "oauth"
+	if !oauthEnabled { // Only used in OAuth flow
+		const oauthDisabled = "OAuth disabled - cannot refresh token"
+		fmt.Println(oauthDisabled)
+		return nil, errors.New(oauthDisabled)
+	}
+
+	authenticator, ok := invocationCtx.GetNetworkAccess().GetAuthenticator().(*auth.OAuth2Authenticator)
```
bastiandoetsch, line 33: This should not use the real type. It should just get an authenticator. If you need to use an interface you could e.g. Define a `RefreshableAuth` interface to be used, exported that has a `Refresh` method, but I'd think twice about exporting the concrete type and asserting the type here.

pkg/local_workflows/authworkflow/workflow.go
```diff
@@ -25,26 +26,43 @@ If you can't wait use this url:
...
+func NewAuthWorkflow(writer io.Writer) *AuthWorkflow {
+	return &AuthWorkflow{
+		writer: writer,
+	}
 }
 
-func OpenBrowser(authUrl string) {
-	fmt.Println(fmt.Sprintf(templateConsoleMessage, authUrl))
+func (aw *AuthWorkflow) OpenBrowser(authUrl string) {
+	_, _ = fmt.Fprintf(aw.writer, templateConsoleMessage, authUrl)
```
bastiandoetsch, line 58: This breaks language server... don't change API, please :)

PR #51, opened 2023-04-26
feat: synchronize oauth token refresh

PeterSchafer:
avoid issues when concurrently trying to refresh the oauth token

### DIFF
--- a/pkg/auth/oauth2authenticator.go
+++ b/pkg/auth/oauth2authenticator.go
@@ -132,34 +132,51 @@ func GetOAuthToken(config configuration.Configuration) (*oauth2.Token, error) {
 	return nil, nil
 }
 
-func refreshToken(ctx context.Context, oauthConfig *oauth2.Config, token *oauth2.Token) (*oauth2.Token, error) {
+func RefreshToken(ctx context.Context, oauthConfig *oauth2.Config, token *oauth2.Token) (*oauth2.Token, error) {
 	tokenSource := oauthConfig.TokenSource(ctx, token)
 	return tokenSource.Token()
 }
 
+//goland:noinspection GoUnusedExportedFunction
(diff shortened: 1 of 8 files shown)


pkg/auth/oauth2authenticator.go
```diff
@@ -266,17 +276,29 @@ func (o *oAuth2Authenticator) AddAuthenticationHeader(request *http.Request) err
...
+			// use TokenSource to refresh the token
+			validToken, err := o.tokenRefresherFunc(ctx, o.oauthConfig, o.token)
+			if err != nil {
+				return err
+			}
 
-	if validToken != o.token {
-		o.persistToken(validToken)
+			if validToken != o.token {
+				o.persistToken(validToken)
```
bastiandoetsch, line 294: It would be good to have be able to register observers here (e.g. via callback functions), that are notified on change. That way language server could be informed of a token refresh and a new token and forward it to its Language Client.

pkg/auth/oauth2authenticator.go
```diff
@@ -150,6 +158,7 @@ func NewOAuth2AuthenticatorWithCustomFuncs(
 		token:              token,
 		openBrowserFunc:    openBrowserFunc,
 		shutdownServerFunc: shutdownServerFunc,
+		tokenRefresherFunc: refreshToken,
```
bastiandoetsch, line 161: as described below, it would be good to be able to also register observer functions that are executed on tokenRefresh/tokenChange

bastiandoetsch: I'll add a PR as a proposal.

bastiandoetsch: see https://github.com/snyk/go-application-framework/pull/54

bastiandoetsch (approved):

PR #53, opened 2023-04-26
feat: add devtools.Cmd helper function

jason-snyk:
This PR adds a `devtools` library package that contains a function, `Cmd`, that is intended to help CLI extension authors write lightweight CLIs to test their extensions outside of the main Snyk CLI. The included example test demonstrates the usage of this function.


pkg/cmd/dev.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/spf13/cobra"
+)
+
+// Dev is a helper utility for extension authors to define a lightweight CLI to
+// test their extensions outside of the main Snyk CLI.
+func Dev(initializers ...workflow.ExtensionInit) (*cobra.Command, error) {
+	// Initialize the engine with the given workflows
+	logger := log.New(os.Stderr, "", 0)
+	engine := app.CreateAppEngineWithLogger(logger)
```
bastiandoetsch, line 19: Better use the `CreateAppEngineWithOpts` func, as this one is just delegating and will most likely be deprecated in the future.

pkg/cmd/dev.go
```diff
@@ -0,0 +1,41 @@
...
+	engine.SetConfiguration(configuration.New())
```
bastiandoetsch, line 20: This would also be possible with the `WithConfiguration` option in above mentioned method.

pkg/cmd/dev.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, i := range initializers {
+		engine.AddExtensionInitializer(i)
```
bastiandoetsch, line 22: Could this be also an option that can be passed to `CreateAppEngineWithOpts`? E.g. `WithExtensionInitializers`?

pkg/cmd/tree.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"fmt"
+
+	localworkflows "github.com/snyk/go-application-framework/pkg/local_workflows"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/spf13/cobra"
+)
+
+type cmdTree struct {
```
bastiandoetsch, line 11: I may be blind... but is this used anywhere? 

Also, I don't think the command tree would belong into the go application framework - commands are only used in CLI applications, and go application framework is used as a library, not a CLI.

jason-snyk: Yes - this struct used here in the `Dev()` function: https://github.com/snyk/go-application-framework/pull/53/files#diff-f9ae7a73af8f4be90839b2b2fd48a2659dd25f5a4d157cfc86973247cb5b6069R28-R35

> Also, I don't think the command tree would belong into the go application framework - commands are only used in CLI applications, and go application framework is used as a library, not a CLI.

Just in case it's not clear - this package that I've added is a library for building a lightweight CLI, not a CLI itself.

This is something that I added to the extension that my team is developing in order to have a way to run our extension outside of the main Snyk CLI. It seems to me that the only alternative to get that kind of fast turnaround is to use a local `replace` directive in the cliv2 `go.mod`. Personally, I don't like to keep those around for very long, because you have to be careful not to commit them and, because `go.mod` and `go.sum` change pretty regularly during normal development, you need to stash it every time you pull in upstream changes. There's also lots of other stuff that the Snyk CLI does, for example analytics, that I don't really want when I just want to try out our extension.

I found this tool useful, so if I were going to write another extension, I would copy this code into it. That indicates that it would make sense as a library. But, if you don't think other extension authors would find it useful then maybe it doesn't belong in this library.

pkg/cmd/dev.go
jason-snyk: Thanks! Addressed in d569237

pkg/cmd/dev.go
jason-snyk: Thanks! Addressed in d569237

pkg/cmd/dev.go
jason-snyk: Thanks! Addressed in d569237

pkg/cmd/dev.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/spf13/cobra"
+)
+
+// Dev is a helper utility for extension authors to define a lightweight CLI to
+// test their extensions outside of the main Snyk CLI.
+func Dev(initializers ...workflow.ExtensionInit) (*cobra.Command, error) {
+	// Initialize the engine with the given workflows
+	logger := log.New(os.Stderr, "", 0)
+	engine := app.CreateAppEngineWithOptions(
+		app.WithLogger(logger),
```
PeterSchafer, line 20: Suggestion: In the latest changes on main, we introduced zerolog, so you could make use of zerolog if you like.

jason-snyk: Thanks! Added in 5e8360f

pkg/cmd/tree.go
PeterSchafer: Personally I think we are okay with this for now, for now it solves the purpose and since nothing is exported it should be relatively safe to change later on if necessary. Originally I was struggling a bit if this should be in `pkg` as this is part of the public API of the library. 

The  only thing in my mind is if we should call the package `dev` oder `devtools`, to be super even double explicit about its intention :)

pkg/cmd/dev_test.go
```diff
@@ -0,0 +1,48 @@
+package cmd_test
```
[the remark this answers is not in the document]
jason-snyk: I'd prefer to keep it as an example test - both because it [runs as part of `go test`](https://pkg.go.dev/testing#hdr-Examples) because godoc renders it as a runnable example in the page, like https://pkg.go.dev/github.com/snyk/policy-engine@v0.21.1/pkg/input#example-Loader.Load

Typically I've kept these in the same place as the code that they're testing, but it could make sense to link to it from further up.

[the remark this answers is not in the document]
PeterSchafer: TIL :)

pkg/cmd/tree.go
jason-snyk: > The only thing in my mind is if we should call the package dev oder devtools, to be super even double explicit about its intention

`devtools.Cmd()` sounds pretty good to me. I'll push this change.

jason-snyk: Renamed in 7472c8e

PeterSchafer (approved):
LGTM

PR #57, opened 2023-04-27
feat: introduce zerolog to better integrate with LS

PeterSchafer:
* using zerolog as a more feature rich logger makes generally sense, in addition to the fact that snyk-ls uses it as well
* At some places we are keeping both log.Logger and zerolog.Logger for backwards compatibility

### DIFF
--- a/internal/utils/zerologConverter.go
+++ b/internal/utils/zerologConverter.go
@@ -3,6 +3,7 @@
 package utils
 
 import (
+	"log"
 	"strings"
 
 	"github.com/rs/zerolog"
@@ -14,5 +15,14 @@ type ToZeroLogDebug struct {
 
 func (w *ToZeroLogDebug) Write(p []byte) (n int, err error) {
 	w.Logger.Debug().Msg(strings.TrimSpace(string(p)))
+	return len(p), nil
+}
+
+type ToLog struct {
+	Logger *log.Logger
+}
+
+func (w *ToLog) Write(p []byte) (n int, err error) {
+	w.Logger.Println(strings.TrimSpace(string(p)))
 	return len(p), nil
 }
--- a/internal/utils/zerologConverter.go
+++ b/internal/utils/zerologConverter.go
@@ -1,0 +1,18 @@
+// OS interface wrappers
+
+package utils
+
+import (
+	"strings"
+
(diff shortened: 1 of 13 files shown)


internal/utils/zerologConverter.go
```diff
@@ -0,0 +1,33 @@
...
+package utils
+
+import (
+	"log"
+	"strings"
+
+	"github.com/rs/zerolog"
+)
+
+// since the are a couple of places that take different types of loggers outside of our scope, we introduce these simple wrapper to adapt between the worlds.
```
bastiandoetsch, line 12: typos in the comment

internal/utils/zerologConverter.go
```diff
@@ -0,0 +1,33 @@
...
+
+// ToZeroLogDebug is an io.Writer that can be used for example in log.Logger to write to an existing zerolog writer
+type ToZeroLogDebug struct {
```
bastiandoetsch, line 15: any change to use `LevelWriter` (that's zerolog) and something like that in the log framework to not loose level information?

internal/utils/zerologConverter.go
```diff
@@ -0,0 +1,33 @@
...
+	w.Logger.Debug().Msg(strings.TrimSpace(string(p)))
+	return len(p), nil
+}
+
+// ToLog is an io.Writer that can be used to write into a log.Logger, for example from a zerolog.Logger
+type ToLog struct {
+	Logger *log.Logger
+}
+
+func (w *ToLog) Write(p []byte) (n int, err error) {
```
bastiandoetsch, line 30: If this implements the `LevelWriter` interface, zerolog uses it, and the log levels are preserved :)

bastiandoetsch (approved):

internal/utils/zerologConverter.go
PeterSchafer: That would be possible. Since we have this only to support legacy usage, it might not be worth though.

internal/utils/zerologConverter.go
PeterSchafer: Since we need this to be an io.Writer to fit into log.Logger, there won't be any level information available.

PR #58, opened 2023-05-02
fix: clone configuration

bastiandoetsch:
fixes clone of configurations with custom storage

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -151,7 +151,15 @@ func (ev *extendedViper) Clone() Configuration {
 	defer ev.mutex.Unlock()
 
 	// manually clone the Configuration instance
-	clone := NewFromFiles(ev.viper.ConfigFileUsed())
+	configFileUsed := ev.viper.ConfigFileUsed()
+	var clone Configuration
+	if configFileUsed != "" {
+		clone = NewFromFiles(configFileUsed)
+	} else {
+		clone = NewInMemory()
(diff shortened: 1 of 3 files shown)


pkg/configuration/configuration.go
```diff
@@ -151,7 +151,15 @@ func (ev *extendedViper) Clone() Configuration {
 	defer ev.mutex.Unlock()
 
 	// manually clone the Configuration instance
-	clone := NewFromFiles(ev.viper.ConfigFileUsed())
+	configFileUsed := ev.viper.ConfigFileUsed()
```
PeterSchafer, line 154: suggestion: should we derive the type of the cloned configuration from this configuration?

pkg/configuration/configuration.go
```diff
@@ -151,7 +151,15 @@ func (ev *extendedViper) Clone() Configuration {
...
+	var clone Configuration
+	if configFileUsed != "" {
+		clone = NewFromFiles(configFileUsed)
+	} else {
+		clone = NewInMemory()
+		clone.SetStorage(ev.storage)
```
PeterSchafer, line 160: suggestion: SetStorage() should be always used.

pkg/configuration/configuration.go
bastiandoetsch: Dunno what you mean - let's chat :)
=

PeterSchafer (approved):

PR #68, opened 2023-05-25
fix: forward --debug flag in depgraph workflow

mcombuechen:
This forwards the `--debug` flag to the `legacycli` workflow when being called in the `depgraph` workflow.

### DIFF
--- a/pkg/local_workflows/depgraph_workflow.go
+++ b/pkg/local_workflows/depgraph_workflow.go
@@ -111,6 +111,10 @@ func depgraphWorkflowEntryPoint(invocation workflow.InvocationContext, input []w
 		debugLogger.Println("File:", file)
 	}
 
+	if config.GetBool(configuration.DEBUG) {
+		snykCmdArguments = append(snykCmdArguments, "--debug")
+	}
+
 	config.Set(configuration.RAW_CMD_ARGS, snykCmdArguments)
 	legacyData, legacyCLIError := engine.InvokeWithConfig(workflow.NewWorkflowIdentifier("legacycli"), config)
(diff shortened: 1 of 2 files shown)


pkg/local_workflows/depgraph_workflow_test.go
```diff
@@ -170,6 +170,27 @@ func Test_Depgraph_depgraphWorkflowEntryPoint(t *testing.T) {
 		assert.Equal(t, expected, actual)
 	})
 
+	t.Run("should support 'debug' flag", func(t *testing.T) {
+		// setup
+		config.Set("debug", true)
```
PeterSchafer, line 175: nitpick: please replace "debug" with configuration.DEBUG

PeterSchafer (approved):

pkg/local_workflows/depgraph_workflow.go
```diff
@@ -111,6 +111,10 @@ func depgraphWorkflowEntryPoint(invocation workflow.InvocationContext, input []w
 		debugLogger.Println("File:", file)
 	}
 
+	if debug := config.GetBool(configuration.DEBUG); debug {
```
bastiandoetsch, line 114: Why so complicated with assigning debug?

mcombuechen: made it shorter!

pkg/local_workflows/depgraph_workflow_test.go
mcombuechen: done!

bastiandoetsch (approved):

PR #72, opened 2023-06-12
Workflows

tommyknows:
This commit introduces a `Workflow` type that can be leveraged to build workflows more easily & in a more standardised fashion. to go with that, this commit also introduces a `Flag` type to easily define & re-use (and potentially access) defined flags of commands.

To demonstrate this new functionality, this commit also changes the `local_workflows` package to use it.

While doing so, I've also refactored the implementations a bit.

Some notes:
- I wasn't sure whether I should make public variables for the Workflows, or if they'd be better off "hidden" behind functions (e.g. `func Auth() *authWorkflow` instead of a `var Auth = &authWorkflow{}`). Let me know if you'd prefer the functions.
- I haven't (yet) updated the contributors-guide, let me know what you think about these changes and if that works out, I'll be happy to update those. Should get a bit shorter now with this new framework.
- Let me know if introducing breaking changes would be okay. for now, I've kept all existing global variables (& constants), although some are definitely redundant now (like the `WORKFLOWID_AUTH`). Personally, I think it should be okay to remove them, but wanted to get some guidance on this first.
- The DepGraph workflow is already built for being extended with a `ContainerDepGraph` workflow - you can view those changes / commits on the [workflows branch](https://github.com/snyk/go-application-framework/compare/main...workflows).
- Sorry if this is a bit much, I got carried away 😅 
- While there's a lot of changes, I barely touched the tests (only made sure they compile). They still work, so I'd say the changes themselves can be considered refactors :)

### DIFF
--- a/pkg/local_workflows/auth_workflow.go
+++ b/pkg/local_workflows/auth_workflow.go
@@ -7,7 +7,6 @@ import (
 	"github.com/snyk/go-application-framework/pkg/auth"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/workflow"
-	"github.com/spf13/pflag"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


tommyknows:
Closing this as we're attempting to move this somewhere else.

PR #73, opened 2023-06-13
chore: Add User-Agent headers [HEAD-344]

asaf92:


pkg/networking/middleware/auth_header.go
```diff
@@ -69,10 +91,14 @@ func ShouldRequireAuthentication(apiUrl string, url *url.URL, additionalSubdomai
...
 
-func AddAuthenticationHeader(authenticator auth.Authenticator, config configuration.Configuration, request *http.Request) error {
+func AddAuthenticationHeader(
+	authenticator auth.Authenticator,
+	config configuration.Configuration,
+	request *http.Request,
+) error {
 	apiUrl := config.GetString(configuration.API_URL)
 	additionalSubdomains := config.GetStringSlice(configuration.AUTHENTICATION_SUBDOMAINS)
-	isSnykApi, err := ShouldRequireAuthentication(apiUrl, request.URL, additionalSubdomains)
```
bastiandoetsch, line 75: As this is an exported function, I would not rename it, as project using the lib could be affected.

pkg/networking/middleware/capture.go
```diff
@@ -0,0 +1,16 @@
+package middleware
+
+import "net/http"
+
+// HeaderCapture A middleware that captures the headers of the request and doesn't send it
+type HeaderCapture struct {
```
bastiandoetsch, line 6: Probably used for testing?

pkg/networking/middleware/useragent.go
```diff
@@ -0,0 +1,49 @@
...
+func (n *UserAgentMiddleware) RoundTrip(request *http.Request) (*http.Response, error) {
+	// Only add headers if the request is going to a Snyk API URL.
+	apiUrl := n.config.GetString(configuration.API_URL)
+	parsedUrl, err := url.Parse(apiUrl)
+	if err != nil {
+		return n.next.RoundTrip(request)
+	}
+
+	isSnykApi, err := IsSnykApi(apiUrl, request.URL, nil)
+	if err != nil || n.snykAppEnvironment == nil || !isSnykUrl(parsedUrl.Hostname()) || !isSnykApi {
```
bastiandoetsch, line 38: Doesn't `isSnykAPI` (or the previous name) encompass `isSnykUrl`?

pkg/networking/networking.go
```diff
@@ -37,23 +43,25 @@ type NetworkAccess interface {
...
+	proxy              func(req *http.Request) (*url.URL, error)
+	caPool             *x509.CertPool
+	logger             *zerolog.Logger
+	snykAppEnvironment *middleware.UserAgentInfo
 }
 
-// customRoundTripper is a custom http.RoundTripper which decorates the request with default headers.
-type customRoundTripper struct {
+// DefaultHeadersRoundTripper is a custom http.RoundTripper which decorates the request with default headers.
+type DefaultHeadersRoundTripper struct {
```
bastiandoetsch, line 55: why is this exported now? why the name change?

pkg/networking/middleware/auth_header.go
```diff
@@ -33,15 +33,37 @@ func (n *AuthHeaderMiddleware) RoundTrip(request *http.Request) (*http.Response,
...
+// url: "https://snyk.io/api/v1/projects"
+// additionalSubdomains: []string{}
+// Result: true, nil // The URL matches the base API URL
+//
+// Example 2:
+// apiUrl: "https://snyk.io/api/"
+// url: "https://test.snyk.io/api/v1/projects"
+// additionalSubdomains: []string{"test"}
+// Result: true, nil // The URL matches the API URL with the "test" subdomain
+func IsSnykApi(apiUrl string, url *url.URL, additionalSubdomains []string) (matchesPattern bool, err error) {
```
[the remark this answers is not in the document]
asaf92: Good point, I'll rename it back and create a new one

pkg/networking/middleware/useragent.go
asaf92: IsSnykUrl is whether it's "snyk.io/snykgov.io", IsSnykApi can be true for any URL given the right parameters.

But the naming is confusing so I'll change it

pkg/networking/middleware/capture.go
asaf92: I use it in "AddHeaders". 

Previously we were calling functions defined in the middlewares that add headers to a request one by one. This is more generic, I run the request through all the MW and collect all the headers that were added without needing both a "RoundTrip" and an "AddXHeaders" function for every MW.

But yeah it's also useful for testing :)

pkg/networking/networking.go
```diff
@@ -87,7 +95,20 @@ func (n *networkImpl) AddHeaderField(key, value string) {
 
 func (n *networkImpl) AddHeaders(request *http.Request) error {
 	n.addDefaultHeader(request)
-	return middleware.AddAuthenticationHeader(n.GetAuthenticator(), n.config, request)
+	hc := &middleware.HeaderCapture{}
+	var rt http.RoundTripper = hc
+	rt = n.addMiddlewaresToRoundTripper(rt)
+	_, err := rt.RoundTrip(request)
+	for k, v := range hc.CapturedHeaders {
+		request.Header.Set(k, v)
```
bastiandoetsch, line 103: how does this work? `hc` is just initiated in the same method empty. I think it would be more explicit to name the `HeaderCapture` like `HeaderCapturingRoundtripper`, as I found the whole method implementation a bit confusing how about something like:

```
	rt, ok := n.addMiddlewaresToRoundTripper(&middleware.HeaderCapture{}).(CapturedHeaders)
	if !ok { return errors.New...) }
	_, err := rt.RoundTrip(request)
	for k, v := range rt.CapturedHeaders {
		request.Header.Set(k, v)
```

pkg/networking/networking.go
```diff
@@ -87,7 +95,20 @@ func (n *networkImpl) AddHeaderField(key, value string) {
...
+	}
+
+	return err
+}
+
+func (n *networkImpl) SetUserAgent(userAgent middleware.UserAgentInfo) error {
+	n.snykAppEnvironment = &userAgent
```
bastiandoetsch, line 110: why is this called snykAppEnvironment, if it's a `UserAgentInfo`?

pkg/networking/networking.go
```diff
@@ -101,19 +122,21 @@ func (n *networkImpl) addDefaultHeader(request *http.Request) {
...
 }
 
-func (n *networkImpl) GetRoundTripper() http.RoundTripper {
-	transport := n.configureRoundTripper(http.DefaultTransport.(*http.Transport))
-
-	rt := middleware.NewAuthHeaderMiddleware(n.config, n.GetAuthenticator(), transport)
-
-	// encapsulate everything
-	roundTrip := customRoundTripper{
+func (n *networkImpl) addMiddlewaresToRoundTripper(rt http.RoundTripper) http.RoundTripper {
```
bastiandoetsch, line 125: This changes the API we should not change exported functions of packages that are not in `internal`

pkg/networking/networking_test.go
```diff
@@ -54,8 +55,9 @@ func Test_HttpClient_CallingApiUrl_UsesAuthHeaders(t *testing.T) {
 
 func Test_HttpClient_CallingApiUrl_UsesAuthHeaders_OAuth(t *testing.T) {
 	config := getConfig()
-	userAgent := "James Bond"
 	accessToken := "access me"
+	integrationName := "my-integration"
```
bastiandoetsch, line 59: why would this test check the user agent? isn't its purpose to check the auth header?

pkg/networking/networking_test.go
```diff
@@ -231,3 +252,125 @@ func Test_AddHeaders_AddsDefaultAndAuthHeaders(t *testing.T) {
...
+	osName := "DARWIN"
+	arch := "ARM64"
+	integrationName := "VS_CODE"
+	integrationVersion := "1.20.1"
+	integrationEnvironment := "language-server"
+	integrationEnvironmentVersion := "1.2.3.4"
+	processName := "snyk-ls"
+	expectedHeader := fmt.Sprint(
+		app, "/", appVersion,
+		" (", osName, ";", arch, ";", processName, ") ",
```
bastiandoetsch, line 268: Is `processName` an addition to what we discussed when discussing the ticket?

bastiandoetsch:
I love the elaborate testing of the functionality and this is good work - but it feels a bit "too much" for just adding a header. If you look at https://github.com/snyk/snyk-ls/pull/319/files, I wonder if we really need a new middleware, a user-agent-info struct and all that. Isn't that complexity that's harder to maintain than necessary?

pkg/networking/middleware/useragent.go
bastiandoetsch: How can something be a Snyk API without being part of `snyk.io` or `snykgov.io`? I think that would not be an allowed pattern.

pkg/networking/networking_test.go
asaf92: It previously checked for the user agent, I changed it to only check for authentication.

These are just leftovers

pkg/networking/networking_test.go
asaf92: No, it's part of the pattern (binary name)

pkg/networking/networking.go
asaf92: Un-exporting it 👍 

I believe it was called "customRoundTripper" when it was the only RT, but now we have several and they are all "custom". 

Essentially it's an RT that adds the default headers, so I named it "defaultHeadersRoundTripper"

pkg/networking/networking.go
asaf92: Renamed 👍 

addMiddlewares... returns a different RoundTriper, so it can't be cast to CaptureHeaders (A -> B -> C -> CaptureHeadersMiddleware).

pkg/networking/networking.go
asaf92: Changed to naming to be consistent 👍

PR #76, opened 2023-06-19
fix: config organization value is unexpectedly the org ID and not org slug 

j-luong:
This PR should add a new configuration field, `ORGANIZATION_SLUG` to differentiate the org ID from the org slug name.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1,6 +1,9 @@
 package app
 
 import (
+	"io"
+	"log"
+
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	zlog "github.com/rs/zerolog/log"
@@ -13,8 +16,6 @@ import (
 	"github.com/snyk/go-application-framework/pkg/networking"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 	"github.com/snyk/go-httpauth/pkg/httpauth"
-	"io"
-	"log"
 )
 
 // initConfiguration initializes the configuration with initial values.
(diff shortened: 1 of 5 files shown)


pkg/app/app.go
```diff
@@ -85,6 +86,34 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
...
+					logger.Print("Failed to determine default value for \"ORGANIZATION_SLUG\":", err)
+				} else {
+					return orgName
+				}
+			} else {
+				return orgName
+			}
+		}
+
+		orgId, err := apiClient.GetDefaultOrgId()
```
PeterSchafer, line 109: Issue: here the default org id is returned, which should actually be the default org slug name.

pkg/app/app.go
```diff
@@ -85,6 +86,34 @@ func initConfiguration(config configuration.Configuration, apiClient api.ApiClie
 		return orgId
 	})
 
+	config.AddDefaultValue(configuration.ORGANIZATION_SLUG, func(existingValue any) any {
+		client := networking.NewNetworkAccess(config).GetHttpClient()
+		url := config.GetString(configuration.API_URL)
+		apiClient.Init(url, client)
+		if existingValue != nil && len(existingValue.(string)) > 0 {
```
PeterSchafer, line 93: Since this value will itself never be set by a user, I think we need to change the logic here a bit. The value specified by the user will be org, holding either the slugname or id of an org.

j-luong:
Will not implement, the orgID should be sufficient for all use cases so far. The additional slug name is not needed yet.

PR #78, opened 2023-06-23
fix: remove now obsolete process name from user-agent

PeterSchafer:


pkg/networking/networking_test.go
```diff
@@ -316,7 +313,6 @@ func Test_AddUserAgent_MissingIntegrationEnvironment_FormattedCorrectly(t *testi
 		IntegrationVersion: integrationVersion,
 		OS:                 osName,
 		Arch:               arch,
-		ProcessName:        processName,
```
bastiandoetsch, line 319: This could be a problem, as this is public.

bastiandoetsch (approved):

pkg/networking/networking_test.go
PeterSchafer: So far it was only used in the CLI and I will adapt it there.

PR #80, opened 2023-07-20
fix: generic bearer_token name to avoid collisions

j-luong:
Currently we do not identify all Snyk/GAF specific bearer token environment variable, this can lead to issues where a bearer_token env var is used by the client for non-snyk purposes. This PR uses the Snyk specific bearer_token env var to avoid this.

### DIFF
--- a/pkg/configuration/constants.go
+++ b/pkg/configuration/constants.go
@@ -4,7 +4,7 @@ const (
 	API_URL                         string = "snyk_api"                // AKA "endpoint" in the config file
 	AUTHENTICATION_SUBDOMAINS       string = "internal_auth_subdomain" // array of additional subdomains to add authentication for
 	AUTHENTICATION_TOKEN            string = "snyk_token"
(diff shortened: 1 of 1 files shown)


pkg/configuration/constants.go
```diff
@@ -4,7 +4,7 @@ const (
 	API_URL                         string = "snyk_api"                // AKA "endpoint" in the config file
 	AUTHENTICATION_SUBDOMAINS       string = "internal_auth_subdomain" // array of additional subdomains to add authentication for
 	AUTHENTICATION_TOKEN            string = "snyk_token"
-	AUTHENTICATION_BEARER_TOKEN     string = "bearer_token"
+	AUTHENTICATION_BEARER_TOKEN     string = "snyk_bearer_token"
```
PeterSchafer, line 7: Suggestion: To reduce lookups, better use the actual variable name 'snyk_oauth_token', which can then be removed from cli main

j-luong: updated 👍

PR #88, opened 2023-08-21
refactor: add UserInterface module [HEAD-58]

asaf92:


pkg/ui/userinterface.go
```diff
@@ -0,0 +1,55 @@
...
+func (ui *consoleUi) Output(output string) error {
+	return utils.ErrorOf(fmt.Fprintln(ui.writer, output))
+}
+
+func (ui *consoleUi) OutputError(err error) error {
+	return utils.ErrorOf(fmt.Fprintln(ui.errorWriter, "Error: "+err.Error()))
+}
+
+func (ui *consoleUi) NewProgressBar() ProgressBar {
+	return newProgressBar(ui.writer)
```
PeterSchafer, line 39: Question: Are you sure that the writer is where the progress should go? For a console progress, I think it might need to go to stderror, since stdout contains the payload data. Maybe double check with the legacy CLI.

pkg/ui/progressbar.go
```diff
@@ -0,0 +1,49 @@
...
+
+func newProgressBar(writer io.Writer) *consoleProgressBar {
+	return &consoleProgressBar{writer: writer}
+}
+
+type consoleProgressBar struct {
+	writer io.Writer
+}
+
+func (p consoleProgressBar) SetProgress(progress float64) error {
```
PeterSchafer, line 32: Question: Could we add a Description in addition to the value? So that use cases like
> Analyzing file 50%
can be enabled.

pkg/ui/progressbar.go
```diff
@@ -0,0 +1,49 @@
...
+	barWidth        = 50
+	clearLine       = "\r\033[K"
+)
+
+type ProgressBar interface {
+	SetProgress(progress float64) error
+	Clear() error
+}
+
+func newProgressBar(writer io.Writer) *consoleProgressBar {
```
PeterSchafer, line 24: Suggestion: This should return ProgressBar, I guess.

pkg/ui/consoleui_test.go
```diff
@@ -0,0 +1,13 @@
+package ui
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDefaultBuilder_NoError(t *testing.T) {
+	ui, err := NewConsoleUiBuilder().Build()
```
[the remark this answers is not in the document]
asaf92: Done

pkg/ui/progressbar.go
asaf92: Done. I add the title at the end though (so it won't move the bar horizontally)

pkg/ui/builder.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/fatih/color"
+)
+
+var red = color.New(color.FgRed)
+
+type ConsoleUiBuilder struct {
```
PeterSchafer, line 14: Question: Do we really need a Builder? It seems that we can support everything with the UserInterface interface. I see the possibility to reduce the amount of code here.

PeterSchafer (approved):

PR #94, opened 2023-09-15
feat: prevent stella:8000 redirecting to api.stella:8000

robh-snyk:
This change prevents `stella:8000` redirecting to use `api.stella:8000`. This is needed to allow developers to successfully configure their snyk cli instance to communicate with their stella environment.

### DIFF
--- a/internal/api/urls.go
+++ b/internal/api/urls.go
@@ -14,6 +14,7 @@ const (
 	api_prefix_dot string = API_PREFIX + "."
 	app_prefix     string = "app"
 	port_suffix    string = ":[0-9]*$"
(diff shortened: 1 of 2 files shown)


PeterSchafer (approved):

internal/api/urls.go
```diff
@@ -32,6 +33,10 @@ func isLocalhost(host string) bool {
 	return addr.IsLoopback()
 }
 
+func isUsingStella (host string) bool {
```
bastiandoetsch, line 36: I would have preferred having a `map[string]bool` with key = hostname for whitelisting, instead one func per host.

PR #97, opened 2023-10-10
feat: temporarily add legacy Session-Token [HEAD-884]

PeterSchafer:
As a temporary workaround introduce the Session-Token Header

### DIFF
--- a/pkg/networking/networking_test.go
+++ b/pkg/networking/networking_test.go
@@ -12,13 +12,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/snyk/go-httpauth/pkg/httpauth"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/oauth2"
+
 	"github.com/snyk/go-application-framework/internal/constants"
 	"github.com/snyk/go-application-framework/pkg/auth"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/networking/certs"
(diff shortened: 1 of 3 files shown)


pkg/networking/networking_test.go
```diff
@@ -41,6 +42,7 @@ func Test_HttpClient_CallingApiUrl_UsesAuthHeaders(t *testing.T) {
 	expectedHeader := http.Header{
 		"User-Agent":    {userAgent},
 		"Authorization": {"token " + token},
+		"Session-Token": {"token " + token},
```
bastiandoetsch, line 45: In Snyk-LS we set it differently:
```
if c.AuthenticationMethod() == lsp.TokenAuthentication {
		if len(token) > 0 {
			req.Header.Set("Session-Token", token) // FIXME: this should be set by GAF, is in the works
		} else {
			err := errors.New("no token found, auth header not added")
			s.errorReporter.CaptureError(err)
			return nil, err
		}
	} else {
		oauthToken, err := c.TokenAsOAuthToken()
		if err == nil && len(oauthToken.AccessToken) > 0 {
			req.Header.Set("Session-Token", "Bearer "+oauthToken.AccessToken) // FIXME: this should be set by GAF, is in the works
		} else {
			err = errors.New("token could not be converted to OAuth token, auth header not added")
			s.errorReporter.CaptureError(err)
			return nil, err
		}
	}
```

bastiandoetsch (approved):

PR #99, opened 2023-10-20
feat: add trace level logging for network requests

PeterSchafer:


pkg/networking/networking.go
```diff
@@ -52,11 +53,34 @@ type defaultHeadersRoundTripper struct {
...
 
+func (rt *defaultHeadersRoundTripper) logRoundTrip(request *http.Request, response *http.Response) {
+	logHeader := http.Header{}
+	loglevel := zerolog.TraceLevel
+
+	if rt.networkAccess.logger.GetLevel() == loglevel {
+		for i, v := range request.Header {
+			for _, value := range v {
+				if i == "Authorization" || i == "Session-Token" {
+					value = strings.Split(v[0], " ")[0] + " ***"
```
bastiandoetsch, line 64: why not display the last 4 characters?

pkg/networking/networking.go
```diff
@@ -52,11 +53,34 @@ type defaultHeadersRoundTripper struct {
...
+			for _, value := range v {
+				if strings.ToLower(i) == "authorization" || strings.ToLower(i) == "session-token" {
+					value = strings.Split(v[0], " ")[0] + " ***"
+				}
+				logHeader.Add(i, value)
+			}
+		}
+	}
+
+	rt.networkAccess.logger.WithLevel(loglevel).Msgf("> request: %s, %s, %v", request.Method, request.URL.String(), logHeader)
```
bastiandoetsch, line 71: Question: Why not: `... logger.Trace()`? Why make the log-level configurable? I think it may not be needed here :)

bastiandoetsch (approved):

pkg/networking/networking.go
PeterSchafer: Done!

pkg/networking/networking.go
PeterSchafer: this is a performance improvement. loglevel is also used to decide wether or not to prepare a special version of the http header for logging.

pkg/networking/networking.go
```diff
@@ -52,11 +53,50 @@ type defaultHeadersRoundTripper struct {
 	networkAccess            *networkImpl
 }
 
+func (rt *defaultHeadersRoundTripper) logRoundTrip(request *http.Request, response *http.Response, err error) {
+	if rt.networkAccess == nil && rt.networkAccess.logger == nil {
+		return
+	}
+
```
cat2608, line 60: Could it be that we need an `||` instead of and `&&`?
https://go.dev/play/p/8WiLadj0lnv

pkg/networking/networking.go
```diff
@@ -52,11 +53,50 @@ type defaultHeadersRoundTripper struct {
...
+					authHeader := strings.Split(value, " ")
+					if len(authHeader) == 2 && len(authHeader[1]) > 4 {
+						value = authHeader[0] + " " + authHeader[1][0:4] + "***"
+					} else {
+						value = "***"
+					}
+				}
+				logHeader.Add(i, value)
+			}
+		}
```
cat2608, line 77: `nitpick`: I wonder if it could be possible to put away the loop logic to a function and assign the result to `logHeader`, something like:

```go
logHeader := sanitiseLogHearder(request.Header)
```

PR #101, opened 2023-10-25
feat: add report analytics extension [HEAD- 879]

cat2608:
We added a new extension that can report analytics data for the "Scan done" event type to the Snyk REST API at `"/rest/api/orgs/"+orgId+"/analytics"`.

Input is validated against the committed JSON Schema which is derived from the Open API spec agreed on between implementing teams.

### DIFF
--- a/pkg/local_workflows/report_analytics_workflow.go
+++ b/pkg/local_workflows/report_analytics_workflow.go
@@ -2,7 +2,6 @@ package localworkflows
 
 import (
 	"bytes"
-	"encoding/json"
 	"fmt"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/workflow"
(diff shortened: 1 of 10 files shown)


bastiandoetsch (approved):
💜

pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -0,0 +1,90 @@
...
+	// register workflow with engine
+	_, err := engine.Register(WORKFLOWID_REPORT_ANALYTICS, workflow.ConfigurationOptionsFromFlagset(config), reportAnalyticsEntrypoint)
+	return err
+}
+
+// reportAnalyticsEntrypoint is the entry point for the reportAnalytics workflow.
+func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputData []workflow.Data) (output []workflow.Data, err error) {
+	// get necessary objects from invocation context
+	config := invocationCtx.GetConfiguration()
+	logger := invocationCtx.GetLogger()
```
PeterSchafer, line 38: FYI: there is also GetEnhancedLogger() if you want to use leveled logging

PR #105, opened 2023-11-14
feat: introduce GetAndIsSet() to determine if [HEAD-1031]

PeterSchafer:
Enable to get configuration values and determine if the retrieved value was explicitly set or not


bastiandoetsch (approved):
passt scho

PR #106, opened 2023-11-15
feat: add RuntimeInfo to workflow engine

mcombuechen:
This adds `RuntimeInfo` to the `workflow` package.


pkg/app/options.go
```diff
@@ -42,3 +42,12 @@ func WithInitializers(initializers ...workflow.ExtensionInit) Opts {
 		}
 	}
 }
+
+func WithRuntimeInfo(name, version string) Opts {
```
mcombuechen, line 46: I'm not sure about the signature; if the `RuntimeInfo` struct will be extended (which is very likely) this will mean a breaking change to this function. Might be better to pass in `workflow.RuntimeInfo` instead? @PeterSchafer

pkg/workflow/invocationcontextimpl.go
```diff
@@ -80,3 +82,7 @@ func (ici *invocationContextImpl) GetEnhancedLogger() *zerolog.Logger {
 func (ici *invocationContextImpl) GetUserInterface() ui.UserInterface {
 	return ici.ui
 }
+
+func (ici *invocationContextImpl) GetRuntimeInfo() RuntimeInfo {
```
mcombuechen, line 86: Obviously this basically proxies `engine.GetRuntimeInfo()`. I was not sure if this is desired or not, would be nice to get clarification.

pkg/app/options.go
PeterSchafer: Question/Suggestion: Why not use RuntimeInfo as a parameter instead of the individual values.

pkg/workflow/invocationcontextimpl.go
PeterSchafer: That is totally fine for now. it enables us to add additional logic later if required.

pkg/workflow/types.go
```diff
@@ -18,6 +18,10 @@ import (
 type Identifier = *url.URL
 type Callback func(invocation InvocationContext, input []Data) ([]Data, error)
 type ExtensionInit func(engine Engine) error
+type RuntimeInfo struct {
```
PeterSchafer, line 21: Suggestion: For the API interface it would be best to have RuntimeInfo as an interface and instead of public members, use getter and setter function. This way we are more open for future evolution of it.

pkg/app/options.go
mcombuechen: I think that's what my comment was suggesting, too. Will change to that then.

pkg/workflow/types.go
mcombuechen: @PeterSchafer please check the recent changes. There is now a package `runtimeinfo` with an `interface`, implementation and some convenience functions.

PeterSchafer (commented):
Looks great! 

One little ask, could you add `networking.UaWithRutime` to use the name and version from a runtime info instance?

pkg/runtimeinfo/runtimeinfo.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (ri *defaultRuntimeInfo) GetVersion() string {
+	return ri.version
+}
+
+func (ri *defaultRuntimeInfo) SetVersion(v string) {
+	ri.version = v
+}
+
+func New(opts ...opt) *defaultRuntimeInfo {
```
PeterSchafer, line 38: Suggestion: please return the interface here

PR #115, opened 2023-12-15
feat: support OAuth client credentials grant

PeterSchafer:
This PR adds OAuth [Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) to the OAuth2Authenticator, in addition to the existing Authorization Code Grant. It therefore switches between these types, per default it'll continue to use Authorization Code Grant and if Client ID and Client Secret are configured, it'll use the Client Credentials Grant.

### DIFF
--- a/pkg/auth/oauth2authenticator.go
+++ b/pkg/auth/oauth2authenticator.go
@@ -24,12 +24,21 @@ import (
 )
 
 const (
-	CONFIG_KEY_OAUTH_TOKEN string        = "INTERNAL_OAUTH_TOKEN_STORAGE"
-	OAUTH_CLIENT_ID        string        = "b56d4c2e-b9e1-4d27-8773-ad47eafb0956"
-	CALLBACK_HOSTNAME      string        = "127.0.0.1"
-	CALLBACK_PATH          string        = "/authorization-code/callback"
-	TIMEOUT_SECONDS        time.Duration = 120 * time.Second
-	AUTHENTICATED_MESSAGE                = "Your account has been authenticated."
+	CONFIG_KEY_OAUTH_TOKEN  string        = "INTERNAL_OAUTH_TOKEN_STORAGE"
+	OAUTH_CLIENT_ID         string        = "b56d4c2e-b9e1-4d27-8773-ad47eafb0956"
+	CALLBACK_HOSTNAME       string        = "127.0.0.1"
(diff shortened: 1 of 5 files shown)


pkg/auth/oauth2authenticator.go
```diff
@@ -193,6 +240,36 @@ func (o *oAuth2Authenticator) persistToken(token *oauth2.Token) {
...
+	config := getOAuthConfigurationClientCredentials(o.oauthConfig)
+
+	// Use the custom HTTP client when requesting a token.
+	if o.httpClient != nil {
+		ctx = context.WithValue(ctx, oauth2.HTTPClient, o.httpClient)
+	}
+
+	// get token
+	token, err := config.Token(ctx)
+	if err == nil {
```
bastiandoetsch, line 265: suggestion: I would invert that logic. It confuses me to have the default flow in the if, and the error flow outside.

pkg/auth/oauth2authenticator_test.go
```diff
@@ -205,3 +206,62 @@ func Test_AddAuthenticationHeader_expiredToken_somebodyUpdated(t *testing.T) {
...
+	config.Set(PARAMETER_CLIENT_SECRET, "secret")
+	expected = AuthorizationCodeGrant
+	actual = determineGrantType(config)
+	assert.Equal(t, expected, actual)
+
+	config.Set(PARAMETER_CLIENT_ID, "id")
+	expected = ClientCredentialsGrant
+	actual = determineGrantType(config)
+	assert.Equal(t, expected, actual)
+}
```
cat2608, line 225: As far as I understand, this test is communicating that the order of the assertions are linked given that for the assertion in line **224** to be true, line **216** must have happened, right? 

Is it correct to say that this test is asserting the case when one of the required params for grant type to be `ClientCredentialsGrant` is missing, the test should fail?

Would it be worth creating units that stress:
- **neither** *client ID* nor *client secret* is set and the expectation is `AuthorizationCodeGrant`.
- **only** the *client secret* is set and the expectation is `AuthorizationCodeGrant`.
- **only** the *client ID* is set and the expectation is `AuthorizationCodeGrant`
- **both** are set and the and the expectation is `ClientCredentialsGrant`

via parametrise test like https://github.com/snyk/issue-policies/blob/main/internal/core/enums/class_type_test.go?

cat2608 (approved):

PR #116, opened 2024-01-03
feat: adds log scrubbing tool

thisislawatts:
Moves the log scrubbing behaviour available in snyk-ls to the shared library. This is preparatory work for adopting this behaviour in the CLI. 
Original implementation: https://github.com/snyk/snyk-ls/blob/main/internal/logging/scrubbingLogWriter.go

Once this has been merged we can update https://github.com/snyk/snyk-ls/pull/424


PeterSchafer (approved):
LGTM

PR #118, opened 2024-01-04
chore: improve reproducibility of username sanitize tests

cmars:
Using the actual user.Current() during tests makes them more difficult to reproduce.

My username is `c`, which causes a lot of false positives; because each "c" substring gets replaced with "REDACTED", I can't run the tests locally as a result.

This makes `user.Current()` patchable, so that test scenarios can use a normalized, consistent "username".

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -55,6 +55,8 @@ type AnalyticsImpl struct {
 	integrationVersion string
 	os                 string
 	command            string
+
+	userCurrent func() (*user.User, error)
 }
 
 // metadataOutput defines the metadataOutput payload.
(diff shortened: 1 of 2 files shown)


pkg/analytics/analytics.go
```diff
@@ -247,6 +247,8 @@ func (a *AnalyticsImpl) GetOutputData() *analyticsOutput {
 	return output
 }
 
+var userCurrent = user.Current
+
```
jbowes, line 251: I'd feel better about this if it were an argument to `New`, or a setter like `SetClient` (depending on the actual use case for those functions; don't wanna mix them)

cmars: Good suggestion, updated!

jbowes (approved):

PR #120, opened 2024-01-04
chore: golangci-lint fixes, gosec

cmars:
Several gosec linter fixes:

- Use secure random source for creating oauth2 verifer.
  https://www.rfc-editor.org/rfc/rfc7636#section-4.1 prescribes a "high-entropy
  cryptographic random [string]".
- nolint comments for sha1 usage, which is not ideal, but should not be a
  security issue. Could be a future FedRAMP issue though...
- TLS listener test uses randomly assigned port and listens only on localhost.
  This should avoid an annoying security pop-up on macOS. Also makes the test
  more reproducible, in the event the dev machine has something listening on
  8443.
- Satisfy slowloris warnings on http.Server setup without read timeouts.

Stacked on #119. PTAL at 39faa9faf90bf3bdf948952a47a1868ab6055258 if you want to review just the new changes introduced here.

CLI-49

### DIFF
--- a/pkg/auth/oauth2authenticator.go
+++ b/pkg/auth/oauth2authenticator.go
@@ -2,15 +2,14 @@ package auth
 
 import (
 	"context"
-	crypto_rand "crypto/rand"
+	"crypto/rand"
 	"crypto/sha256"
 	"encoding/base64"
-	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"html"
-	"math/rand"
+	"math/big"
 	"net"
 	"net/http"
 	"sync"
@@ -24,6 +23,7 @@ import (
 )
 
 const (
+	//nolint:gosec // not a token value, but a configuration key
 	CONFIG_KEY_OAUTH_TOKEN  string        = "INTERNAL_OAUTH_TOKEN_STORAGE"
 	OAUTH_CLIENT_ID         string        = "b56d4c2e-b9e1-4d27-8773-ad47eafb0956"
 	CALLBACK_HOSTNAME       string        = "127.0.0.1"
@@ -56,21 +56,6 @@ type oAuth2Authenticator struct {
 	openBrowserFunc    func(authUrl string)
 	shutdownServerFunc func(server *http.Server)
 	tokenRefresherFunc func(ctx context.Context, oauthConfig *oauth2.Config, token *oauth2.Token) (*oauth2.Token, error)
-}
-
-func init() {
-	var seed int64
-	var b [8]byte
-
-	// try to use a real random value to seed the pseudo random number generator and
-	// only fall back to time based seed if this didn't work.
-	_, err := crypto_rand.Read(b[:])
-	if err != nil {
-		seed = time.Now().UnixNano() // fallback to time only if necessary
-	} else {
-		seed = int64(binary.LittleEndian.Uint64(b[:])) // based on https://stackoverflow.com/a/54491783
-	}
-	rand.Seed(seed)
 }
 
 func OpenBrowser(authUrl string) {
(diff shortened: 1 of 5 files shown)


cmars:
There's a technique for mitigating modulo bias we can apply here to get a uniform distribution out of crypto/rand. Will update shortly...

pkg/auth/oauth2authenticator.go
```diff
@@ -136,15 +120,45 @@ func createVerifier(count int) []byte {
...
+func randIndex(limit int) int {
+	if limit > maxRandIndexLimit {
+		panic(fmt.Sprintf("unsupported randIndex limit %d", limit))
+	}
+	rejectLimit := 256 - (256 % limit)
+	for {
+		var buf [1]byte
+		_, err := rand.Reader.Read(buf[:])
+		if err != nil {
+			panic(fmt.Sprintf("failed to read secure random bytes: %v", err))
```
cmars, line 153: We could also return the error and wind that through `authenticateWithAuthorizationCode`, returning it from there. OTOH, if this fails, [your system is probably FUBAR](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/crypto/rand/rand.go;l=11).

I can go either way on this one.

PeterSchafer: In general I would prefer to return errors also above in order to gracefully fail. sure we can try to gracefully fail on panics as well, but in my opinion, it is much clearer and easier with errors.

pkg/auth/oauth2authenticator.go
```diff
@@ -136,15 +120,45 @@ func createVerifier(count int) []byte {
...
+// random source cannot be read.
+//
+// Further reading:
+// https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/
+// https://stackoverflow.com/a/10984975
+func randIndex(limit int) int {
+	if limit > maxRandIndexLimit {
+		panic(fmt.Sprintf("unsupported randIndex limit %d", limit))
+	}
+	rejectLimit := 256 - (256 % limit)
```
PeterSchafer, line 148: Question: I'm not sure yet, why it wouldn't be enough to do only rejection sampling, using the given limit.

pkg/auth/oauth2authenticator.go
```diff
@@ -136,15 +120,45 @@ func createVerifier(count int) []byte {
...
+	for {
+		var buf [1]byte
+		_, err := rand.Reader.Read(buf[:])
+		if err != nil {
+			panic(fmt.Sprintf("failed to read secure random bytes: %v", err))
+		}
+		r := int(buf[0])
```
PeterSchafer, line 155: Question: Is there a reason we use only byte 0? If we would make use of the full integer siye, I assume we wouldn't need to limit this to 256, or am I missing something?

pkg/auth/oauth2authenticator.go
cmars: I'm convinced, will update!

pkg/auth/oauth2authenticator.go
cmars: We could, but we'd end up rejecting a lot of potentially useful values for a small `limit`.

Rejecting any sample greater than the largest multiple of `limit` makes more use of the raw random value. The selected sample `% limit` will still have a uniform distribution.

cmars: Actually.. I'm feeling like it's probably safer to use `crypto/rand.Int` here. For the ranges we're working with... it's not too different from what I was attempting here. Best to keep it simple.

cmars:
@PeterSchafer PTAL. Updated to use `crypto/rand.Int` instead, which feels much safer.

PeterSchafer (approved):

PR #121, opened 2024-01-08
fix: Authentication when using an invalid pre-existing token [HEAD-1197]

PeterSchafer:
This PR ensures to unset any available token when executing token authentication in auth_workflow.go
Therefore it adds cloning of the global NetworkAccess object before invoking an extension to avoid side effects between Extensions. This means that an Extension is even more sandboxed and will not affect global states or other Extension states.

To enable the CI/CD pipeline, I updated the runner image used to have a more recent golang version available.

### DIFF
--- a/pkg/workflow/engine_test.go
+++ b/pkg/workflow/engine_test.go
@@ -178,17 +178,25 @@ func Test_Engine_ClonedNetworkAccess(t *testing.T) {
 
 	workflowId := NewWorkflowIdentifier("cmd")
 	_, err := engine.Register(workflowId, ConfigurationOptionsFromFlagset(pflag.NewFlagSet("1", pflag.ExitOnError)), func(invocation InvocationContext, input []Data) ([]Data, error) {
+		assert.Equal(t, expected, invocation.GetNetworkAccess().GetConfiguration().GetInt(valueName))
+		assert.Equal(t, expected, invocation.GetConfiguration().GetInt(valueName))
+
+		newValue := 1
+
 		// changing the network configuration inside a callback
-		invocation.GetNetworkAccess().GetConfiguration().Set(valueName, 1)
+		invocation.GetNetworkAccess().GetConfiguration().Set(valueName, newValue)
+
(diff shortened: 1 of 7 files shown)


pkg/workflow/engine_test.go
```diff
@@ -167,3 +167,30 @@ func Test_Engine_SetterRuntimeInfo(t *testing.T) {
...
+
+	engine := NewWorkFlowEngine(config)
+
+	workflowId := NewWorkflowIdentifier("cmd")
+	_, err := engine.Register(workflowId, ConfigurationOptionsFromFlagset(pflag.NewFlagSet("1", pflag.ExitOnError)), func(invocation InvocationContext, input []Data) ([]Data, error) {
+		// changing the network configuration inside a callback
+		invocation.GetNetworkAccess().GetConfiguration().Set(valueName, 1)
+		return []Data{}, nil
+	})
+	assert.Nil(t, err)
```
cmars, line 185: Consider `assert.NoError` over `assert.Nil`. The output is a little nicer. https://go.dev/play/p/9ulULYKqeqr for example.

pkg/workflow/engine_test.go
```diff
@@ -167,3 +167,30 @@ func Test_Engine_SetterRuntimeInfo(t *testing.T) {
...
+	expected := 815
+	config := configuration.NewInMemory()
+	config.Set(valueName, expected)
+
+	engine := NewWorkFlowEngine(config)
+
+	workflowId := NewWorkflowIdentifier("cmd")
+	_, err := engine.Register(workflowId, ConfigurationOptionsFromFlagset(pflag.NewFlagSet("1", pflag.ExitOnError)), func(invocation InvocationContext, input []Data) ([]Data, error) {
+		// changing the network configuration inside a callback
+		invocation.GetNetworkAccess().GetConfiguration().Set(valueName, 1)
```
cmars, line 182: Can we assert the configuration override from within the registered extension?

cmars (approved):
LGTM, couple of comments to consider.

PR #122, opened 2024-01-11
fix: path to the experimental analytics endpoint

chdorner-snyk:
I'm assuming when we were writing this the endpoint wasn't functional yet and we couldn't test it against a real endpoint. The path to it has slightly changed, which this PR fixes.


bastiandoetsch:
LGTM, please integrate this GAF commit into CLI & LS after merging.

PR #124, opened 2024-02-05
feat: implement persistent message outbox for report analytics workflow [IDE-53]

bastiandoetsch:
This PR adds 

- the concrete implementation to use a database as outbox for the report analytics workflow

It uses SQLite using a [PureGo library](https://pkg.go.dev/modernc.org/sqlite?utm_source=godoc). Also it uses [goose](https://github.com/pressly/goose) for table creation / migrations.

The internal RFC is linked in the JIRA ticket.


pkg/persistence/persistence_impl.go
```diff
@@ -0,0 +1,66 @@
+package persistence
+
+import "fmt"
+
+// PersistableID is the type of the UUID of a persistable object.
+type PersistableID string
```
PeterSchafer, line 6: Nitpick: The package name defines a scope and therefore I would recommend to remove the prefix `Persistable` in all the Type names. Eventually it'll be `persistence.ID` ...

PeterSchafer:
Suggestion: In general I would recommend to separate the general persistable interface in a file and have the sqlite related parts in their own file as well. I know this sounds like a matter of taste and it probably is, but limiting the files impacted by sqlite mentioning will improve maintainability in the long run.

internal/localworkflows/reportanalytics/migrations/0001_initial.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
+CREATE TABLE IF NOT EXISTS outbox (
+    id TEXT PRIMARY KEY,
+    timestamp INTEGER NOT NULL,
+    retries INTEGER DEFAULT 0,
+    payload BLOB NOT NULL
```
cmars, line 6: If the payload is an opaque blob... how would a consumer of this outbox know whether it can process the contents? Is there a schema associated with it?

cmars (changes requested):
Several concerns with this:
- Payload is opaque... does it need versioning? Some kind of schema defined?
- Is it only for analytics, or other use cases?
- How do we avoid filling up a disk with unsent / unsendable analytics?
- How do we expire / rotate out unsent analytics?

Given that Snyk doesn't really work offline today... do we need a durable file store of these analytics? Or would backoff retry attempts from a circular buffer in memory suffice?

bastiandoetsch:
Closing for now, as we'll only introduce this additional resilience when necessary.

PR #126, opened 2024-02-27
fix: normalize durations in report analytics workflow

bastiandoetsch:
Previously, we have had values that were way to high, especially in Jetbrains. This PR sets the duration to 0, if values are reported that cannot be true:

- duration > 12h
- duration < 0ms

A second part of the fix would be to investigate, why Jetbrains sometimes reports arbitrarily high numbers. One possibility is that the start time is not set correctly, so the substraction from end time - start time is end time - 0, yielding the duration from 1970 to now in milliseconds.

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -2,8 +2,9 @@ package analytics
 
 import (
 	"bytes"
+	"crypto/sha256"
+
 	//nolint:gosec // insecure sha1 used for legacy identifier
-	"crypto/sha1"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -36,7 +37,7 @@ type Analytics interface {
 	AddHeader(headerFunc func() http.Header)
 	SetClient(clientFunc func() *http.Client)
 	IsCiEnvironment() bool
-	GetOutputData() *analyticsOutput
+	GetOutputData() *AnalyticsOutput
 	GetRequest() (*http.Request, error)
 	Send() (*http.Response, error)
 }
@@ -70,8 +71,8 @@ type metadataOutput struct {
 type metricsOutput struct {
 }
 
-// analyticsOutput defines the analyticsOutput payload.
-type analyticsOutput struct {
+// AnalyticsOutput defines the AnalyticsOutput payload.
+type AnalyticsOutput struct {
 	Command                       string         `json:"command"`
 	Args                          []string       `json:"args"`
 	OsPlatform                    string         `json:"osPlatform"`
@@ -94,7 +95,7 @@ type analyticsOutput struct {
 
 // dataOutput defines the dataOutput payload.
 type dataOutput struct {
-	Data analyticsOutput `json:"data"`
+	Data AnalyticsOutput `json:"data"`
 }
 
 var (
@@ -214,8 +215,8 @@ func (a *AnalyticsImpl) IsCiEnvironment() bool {
 }
 
 // GetOutputData returns the analyticsOutput data.
-func (a *AnalyticsImpl) GetOutputData() *analyticsOutput {
-	output := &analyticsOutput{}
+func (a *AnalyticsImpl) GetOutputData() *AnalyticsOutput {
+	output := &AnalyticsOutput{}
 
 	errorCount := len(a.errorList)
 	if errorCount > 0 {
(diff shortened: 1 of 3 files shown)


pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -81,27 +84,53 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
...
+		return input
+	}
+
+	durationMs, err := strconv.ParseInt(scanDoneEvent.Data.Attributes.DurationMs, 10, 64)
+	if err != nil {
+		return input
+	}
+
+	if durationMs > 1000*3600*12 || durationMs < 0 {
+		scanDoneEvent.Data.Attributes.DurationMs = "0"
```
PeterSchafer, line 114: Question: wouldn't it make sense to indicate an error case with a distinguishable value instead of "0" like "-1"

bastiandoetsch: Isn't that the same? 0 ms is impossible anyways.

pkg/analytics/analytics.go
```diff
@@ -225,13 +226,17 @@ func (a *AnalyticsImpl) GetOutputData() *analyticsOutput {
...
-	// deepcode ignore InsecureHash: It is just being used to generate an id, without any security concerns
-	//nolint:gosec // sha1 only used to generate an id
-	shasum := sha1.New()
-	//nolint:errcheck // breaking api change needed to fix this
-	uuid, _ := uuid.GenerateUUID()
-	//nolint:errcheck // breaking api change needed to fix this
-	io.WriteString(shasum, uuid)
+	shasum := sha256.New() // generating a sha256 is not necessary, but gets rid of linter and sec warnings
+	generatedUUID, err := uuid.GenerateUUID()
+	if err != nil {
```
PeterSchafer, line 231: Issue: this is a change of behaviour, in the previous version errors have been ignored, now they lead to returning nil. Similar changes have already created bugs, which is why I wouldn't just do it on the side.

cmars: If you can't generate a UUID this needs to fail. Either change the return parameters to add the error, or panic. I'd prefer the first, but I'd accept the latter (a broken secure rng is probably a good reason to crash).

cmars (changes requested):
I think errors need to be surfaced

pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -94,16 +99,39 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
 	return nil, nil
 }
 
-func callEndpoint(invocationCtx workflow.InvocationContext, input workflow.Data, url string) error {
+func normalizeInput(input []byte) []byte {
+	var scanDoneEvent json_schemas.ScanDoneEvent
+	err := json.Unmarshal(input, &scanDoneEvent)
+	if err != nil {
+		return input
```
cmars, line 106: Why would we get an error here? Is this an HTTP 400 kind of situation?

pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -94,16 +99,39 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
...
+	}
+
+	durationMs, err := strconv.ParseInt(scanDoneEvent.Data.Attributes.DurationMs, 10, 64)
+	if err != nil {
+		return input
+	}
+
+	if durationMs > 1000*3600*12 || durationMs < 0 {
```
cmars, line 114: If the duration is longer than 12 hours, we make it look like 0? That seems wrong

pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -94,16 +99,39 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
...
 
-func callEndpoint(invocationCtx workflow.InvocationContext, input workflow.Data, url string) error {
+func normalizeInput(input []byte) []byte {
+	var scanDoneEvent json_schemas.ScanDoneEvent
+	err := json.Unmarshal(input, &scanDoneEvent)
+	if err != nil {
+		return input
+	}
+
+	durationMs, err := strconv.ParseInt(scanDoneEvent.Data.Attributes.DurationMs, 10, 64)
```
PeterSchafer, line 109: Suggestion: You might want to use something like `durationMs float64 `json:",string"`` from https://pkg.go.dev/encoding/json#Marshal. This makes the code much smaller.

pkg/local_workflows/report_analytics_workflow.go
cmars: I think it's better to capture what the client actually sees. If you hide it, you can't troubleshoot it. If it messes up reports, that can be filtered on the reporting side

PeterSchafer: or some other way to indicate the error

pkg/local_workflows/report_analytics_workflow.go
cmars: `int64` would be conventional here but yes, we shouldn't be re-parsing if we can help it.

pkg/local_workflows/report_analytics_workflow.go
cmars: How and where is this used? If we surface the error, can we handle it explicitly up the stack?

If it goes into an analytics request, maybe the better choice is to set the duration to null, or reject the request.

If we're getting durations > 12h, isn't that a bug in the caller? I don't know that hiding that is helping us address the root cause.

bastiandoetsch:
Closing this, as - as Casey pointed out - we can filter wrong values at the backend, so I'm saving the effort needed for the good suggestions from you guys :).

PR #127, opened 2024-02-28
fix: write error output to stderr

mcombuechen:
Since this repo does not have Issues enabled, I'm using a PR to raise the topic. I noticed that in the UI package, `Stdout` is being used for both output as well as error messages. This changes the default error writer to `os.Stderr`. Happy to hear your thoughts on this.


mcombuechen:
Deemed unsafe, closing this PR. See here: https://snyk.slack.com/archives/C06B0PKC0HL/p1709115477388389.

PR #128, opened 2024-03-01
chore: enable golangci-lint

cmars:
Adding golangci-lint to the lint Makefile target, which means it will gate CI.

This PR is a little long, but I've tried to scope it down to a handful of rulesets whose changes are mechanical and generally low-risk (do not introduce significant logic changes, breaking API changes or refactoring).

I've also commented out the rules that I'd like to revisit in followups, especially the ones that will require careful review and possibly some discussion to address.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -61,12 +61,13 @@ func initConfiguration(engine workflow.Engine, config configuration.Configuratio
 		client := engine.GetNetworkAccess().GetHttpClient()
 		url := config.GetString(configuration.API_URL)
 		apiClient.Init(url, client)
-		if existingValue != nil && len(existingValue.(string)) > 0 {
-			orgId := existingValue.(string)
+		existingString, ok := existingValue.(string)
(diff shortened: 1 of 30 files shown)


thisislawatts (approved):

pkg/app/app.go
```diff
@@ -61,12 +61,13 @@ func initConfiguration(engine workflow.Engine, config configuration.Configuratio
 		client := engine.GetNetworkAccess().GetHttpClient()
 		url := config.GetString(configuration.API_URL)
 		apiClient.Init(url, client)
-		if existingValue != nil && len(existingValue.(string)) > 0 {
-			orgId := existingValue.(string)
+		existingString, ok := existingValue.(string)
```
PeterSchafer, line 64: Question: is it safe in golang to do a type conversion without checking if the value is nil?

cmars: Yes. https://go.dev/play/p/Lm_pTMxJLRl

PeterSchafer (approved):

PR #129, opened 2024-03-04
chore: satisfy lint errcheck rule

cmars:
Handle error return values to the greatest extent possible.

Ignore errcheck in areas where a breaking API change would be required, to be addressed in followups.

### DIFF
--- a/pkg/auth/oauth2authenticator.go
+++ b/pkg/auth/oauth2authenticator.go
@@ -59,11 +59,13 @@ type oAuth2Authenticator struct {
 }
 
 func OpenBrowser(authUrl string) {
+	//nolint:errcheck // breaking api change needed to fix this
 	_ = browser.OpenURL(authUrl)
 }
 
 func ShutdownServer(server *http.Server) {
 	time.Sleep(500)
+	//nolint:errcheck // breaking api change needed to fix this
 	_ = server.Shutdown(context.Background())
 }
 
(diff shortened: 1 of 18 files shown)


pkg/auth/oauth2authenticator.go
```diff
@@ -365,7 +372,9 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
 		return err
 	}
 
-	o.persistToken(token)
+	if err = o.persistToken(token); err != nil {
```
PeterSchafer, line 375: Suggestion: How about returning err always. I think the code can be made easier here
```suggestion
	err = o.persistToken(token)
	return err
```

pkg/auth/oauth2authenticator_test.go
```diff
@@ -272,39 +279,27 @@ func Test_determineGrantType_both(t *testing.T) {
...
 
 func Test_Authenticate_CredentialsGrant(t *testing.T) {
-	go func() {
-		mux := http.NewServeMux()
-		srv := &http.Server{
-			Handler:           mux,
-			Addr:              "localhost:3221",
-			ReadHeaderTimeout: time.Second * 30,
+	mux := http.NewServeMux()
+	srv := httptest.NewServer(mux)
```
PeterSchafer, line 283: Nice 💯

PeterSchafer (approved):

PR #130, opened 2024-03-13
feat: progress tracker [IDE-175]

teodora-sandu:
This PR is extracting the progress tracker to GAF so that it can be used by the CLI too. The LS uses it to update the display for our IDEs, the CLI could use it to log or do whatever sort of updates they want (for example, update the progress bar).

The changes in `snyk-ls` are in https://github.com/snyk/snyk-ls/pull/465.


go.mod
```diff
@@ -1,56 +1,56 @@
 module github.com/snyk/go-application-framework
 
-go 1.18
+go 1.21
```
bastiandoetsch, line 3: @PeterSchafer this should be ok, now that CLI is using a higher go version, right?

pkg/progress/progress.go
```diff
@@ -0,0 +1,164 @@
...
+	t.ReportWithMessage(percentage, "")
+}
+
+func (t *Tracker) End() {
+	t.EndWithMessage("")
+}
+
+func (t *Tracker) EndWithMessage(message string) {
+	if t.finished {
+		panic("Called end progress twice. This breaks LSP in Eclipse fix me now and avoid headaches later")
```
bastiandoetsch, line 111: *cough*

teodora-sandu: oops

pkg/progress/types.go
```diff
@@ -0,0 +1,104 @@
+/*
+ * © 2022 Snyk Limited All rights reserved.
```
bastiandoetsch, line 2: 2024

pkg/progress/progress_test.go
```diff
@@ -0,0 +1,112 @@
+/*
+ * © 2022 Snyk Limited All rights reserved.
```
bastiandoetsch, line 2: 2024

bastiandoetsch (approved):

pkg/progress/progress.go
teodora-sandu: Is this something we should fix now? It's a bit annoying to mention LSP and Eclipse, plus not ideal to cause a panic in the CLI although if the function is used correctly we shouldn't cause it. It might be better to log an error instead and return early maybe?

go.mod
thisislawatts: We are on go 1.21 now

PR #131, opened 2024-03-14
feat: adds code test workflow [CLI-187]

thisislawatts:


pkg/local_workflows/code_workflow.go
```diff
@@ -0,0 +1,46 @@
...
+	codeWorkflowName = "code.test"
+)
+
+// WORKFLOWID_CODE defines a new workflow identifier
+var WORKFLOWID_CODE workflow.Identifier = workflow.NewWorkflowIdentifier(codeWorkflowName)
+
+// InitCodeWorkflow initializes the code workflow before registering it with the engine.
+func InitCodeWorkflow(engine workflow.Engine) error {
+	// register workflow with engine
+	flags := pflag.FlagSet{}
```
PeterSchafer, line 21: Issue: I think we will need to list the flags properly since when this is integrated in the CLI, cobra will not accept any argument that is not listed.

pkg/local_workflows/code_workflow.go
```diff
@@ -0,0 +1,46 @@
...
+	_, err := engine.Register(WORKFLOWID_CODE, workflow.ConfigurationOptionsFromFlagset(&flags), codeWorkflowEntryPoint)
+	return err
+}
+
+// codeWorkflowEntryPoint is the entry point for the code workflow.
+// it provides a wrapper for the legacycli workflow
+func codeWorkflowEntryPoint(invocationCtx workflow.InvocationContext, _ []workflow.Data) (output []workflow.Data, err error) {
+	// get necessary objects from invocation context
+	config := invocationCtx.GetConfiguration()
+	logger := invocationCtx.GetLogger()
```
PeterSchafer, line 31: Suggestion: There is an [enhanced logger](https://github.com/snyk/go-application-framework/blob/main/pkg/workflow/invocationcontextimpl.go#L79), that allows level logging. Not necessary now but generally a good approach to use this logger.

pkg/local_workflows/code_workflow_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/snyk/go-application-framework/pkg/mocks"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Code_codeWorkflowEntryPoint_happyPath(t *testing.T) {
+	// set
+	logger := log.New(os.Stderr, "test", 0)
+	config := configuration.New()
+	engine := mocks.NewMockEngine(gomock.NewController(t))
```
PeterSchafer, line 19: note: another nice way to mock is to use the normal engine, register a test workflow under the ID legacycli and in this workflow do your validations. See [here](https://github.com/snyk/go-application-framework/blob/main/pkg/workflow/engine_test.go) as an example.

pkg/local_workflows/code_workflow.go
thisislawatts: https://github.com/snyk/go-application-framework/pull/131/commits/14996bd1b700a47ad756a0cfabece8968b5efbb4

pkg/local_workflows/code_workflow_test.go
thisislawatts: What would be the advantages of that approach?

.golangci.yaml
```diff
@@ -170,6 +170,10 @@ issues:
     - path: test/
       linters:
         - testpackage
+    - path: pkg/local_workflows/auth|report_analytics|whoami|output_workflow.go
```
cmars, line 173: Why are we ignoring static checks on new changes?

thisislawatts: In https://github.com/snyk/go-application-framework/pull/131/commits/edff805cf117356939f7b901844f3f84ed54425e, I introduced an annotation which is now caught by the linter. However I don't want to address the existing instances of `GetLogger` as part of this PR. I've captured the work in a ticket for follow up either in the next sprint(s) or as part of cool down.

pkg/local_workflows/code_workflow_test.go
PeterSchafer: In general I think both are valid approaches to mock. 

The current solution needs to specify  mock behaviour for calls that are not actually of interest for the actual function under test, like the `invocationContextMock`, it exposes some implementation detail to the test, that the test doesn't need. Which can make maintainability more complex. 

Using a real engine and registering an artificial callback uses the real engine implementation and just mocks the `legacycli` implementation. Which is probably would your main interest is in this test and keeps the test more focussed.

.golangci.yaml
PeterSchafer: Should we maybe not use the `deprecated` label yet? It'll surface in every extension as well. So maybe we just mention to use the newer `GetEnhancedLogger()` in the docs for now and remove the linter exception.

PeterSchafer (approved):

PR #133, opened 2024-03-22
feat: introduce feature flag for code test workflow

PeterSchafer:
This PR adds a generic feature flag mechanism to the existing api client and uses this implementation to determine the value of a feature flag to enable a new code test workflow.
Currently the. feature flag only changes the log message but the test logic is the same.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -97,7 +97,7 @@ func (a *snykApiClient) GetFeatureFlag(flagname string) (bool, error) {
 		return defaultResult, fmt.Errorf("unable to retrieve feature flag (status: %d): %w", res.StatusCode, err)
 	}
 
-	if res.StatusCode != http.StatusOK || flag.Code == 401 || flag.Code == 403 {
+	if res.StatusCode != http.StatusOK || flag.Code == "401" || flag.Code == "403" {
 		return defaultResult, nil
 	}
 
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -16,6 +16,7 @@ type ApiClient interface {
 	GetDefaultOrgId() (orgID string, err error)
 	GetOrgIdFromSlug(slugName string) (string, error)
 	Init(url string, client *http.Client)
+	GetFeatureFlag(flagname string) (bool, error)
 }
 
 type snykApiClient struct {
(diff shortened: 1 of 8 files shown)


internal/api/api.go
```diff
@@ -76,6 +77,33 @@ func (a *snykApiClient) GetDefaultOrgId() (string, error) {
...
+		return defaultResult, fmt.Errorf("unable to retrieve feature flag: %w", err)
+	}
+
+	var flag contract.OrgFeatureFlagResponse
+	if err = json.Unmarshal(body, &flag); err != nil {
+		return defaultResult, fmt.Errorf("unable to retrieve feature flag (status: %d): %w", res.StatusCode, err)
+	}
+
+	if res.StatusCode != http.StatusOK || flag.Code == "401" || flag.Code == "403" {
+		return defaultResult, nil
```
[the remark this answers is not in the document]
PeterSchafer: 1. There is a bit of optionality in the API response, which makes it unclear under which circumstances which result is being returned. I used [this code](https://github.com/snyk/cli/blob/main/src/lib/feature-flags/index.ts#L27) as a start and tested the API response. This condition looks like this to catch all cases.
2. I don't think it matters for the consumer of the function call. If we reach this line, there is no error happening.

pkg/app/app.go
```diff
@@ -17,6 +17,25 @@ import (
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
+func defaultFunc_FF_CODE_CONSISTENT_IGNORES(engine workflow.Engine, config configuration.Configuration, apiClient api.ApiClient, logger *zerolog.Logger) configuration.DefaultValueFunction {
+	callback := func(existingValue interface{}) interface{} {
```
j-luong, line 21: question: are we ok not defining an interface and using `interface{}` here? It feels analogous to TypeScript's `any` which is a bit of a bad habit one can fall into

internal/api/contract/OrgFeatureFlagResponse.go
```diff
@@ -0,0 +1,8 @@
+package contract
+
+type OrgFeatureFlagResponse struct {
+	Ok          bool   `json:"ok,omitempty"`
+	UserMessage string `json:"userMessage,omitempty"`
+	Code        string `json:"code,omitempty"`
```
thisislawatts, line 6: **Question:** Did you consider unmarshaling Code into an `int`? This would mean we could switch our checks to use the `http.StatusUnauthorized` and `http.StatusForbidden` constants. 

```suggestion
	Code        int `json:"code,omitempty,string"`
```

PeterSchafer: The API provides a String, using an Integer makes the unmarshalling fail.

PeterSchafer: `{"code":"401","status":"Unauthorized"}`

pkg/app/app.go
PeterSchafer: We are just aligning with [Vipers Get() interface](https://github.com/spf13/viper/blob/master/viper.go#L900C22-L900C25)

pkg/app/app.go
```diff
@@ -17,6 +17,25 @@ import (
...
+		if existingValue == nil {
+			flagname := "snykCodeConsistentIgnores"
+			client := engine.GetNetworkAccess().GetHttpClient()
+			url := config.GetString(configuration.API_URL)
+			apiClient.Init(url, client)
+			result, err := apiClient.GetFeatureFlag(flagname)
```
thisislawatts, line 27: **Question:** Is there any retry logic implemented here, or are we introducing a point of failure here where an intermittent network failure could give a false negative?

internal/api/api.go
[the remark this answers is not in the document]
j-luong: > I don't think it matters for the consumer of the function call. If we reach this line, there is no error happening.

Would it be useful for logging/debugging purposes? e.g. if we pass in a miss-spelled flagname, how easy would this be to spot by looking at the logs?

internal/api/api.go
```diff
@@ -76,6 +77,33 @@ func (a *snykApiClient) GetDefaultOrgId() (string, error) {
...
+	}
+
+	return true, nil
+}
+
```
thisislawatts, line 106: **praise:** 🥳 Great to see this added as a general helper, I look forward to all of the features we can see roll out across the CLI the now

internal/api/api.go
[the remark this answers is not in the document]
PeterSchafer: done

pkg/app/app.go
PeterSchafer: Currently there is no retry logic at all. The proper place for it would be in the http client.

j-luong (approved):

PR #135, opened 2024-03-25
fix: missing check if file exists before attempting to remove it

PeterSchafer:
After a recent refactoring we are not ignoring return values anymore. This causes a bug upstream when trying to remove a file that doesn't exist


internal/utils/outputs.go
```diff
@@ -20,6 +21,9 @@ func (odi *OutputDestinationImpl) Println(a ...any) (n int, err error) {
 }
 
 func (odi *OutputDestinationImpl) Remove(name string) error {
+	if _, err := os.Stat(name); errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
```
cmars, line 26: Can we add a test for this behavior?

cmars (approved):
LGTM but consider adding a test that the remove implementation is now idempotent.

I can't tell if the output workflow test is actually exercising that code path or not; it might be good to unit test `Remove` in isolation?

internal/utils/outputs.go
PeterSchafer: Done! Thanks for the feedback!

PR #137, opened 2024-04-09
feat: presenter for Sarif test results

thisislawatts:
Introduce a handler for generating string representation of Test Results based on sarif input and test metadata properties, org name and test path. 

[
Screenshot 2024-04-16 at 15 27 23
](url)


go.mod
```diff
@@ -1,28 +1,32 @@
 module github.com/snyk/go-application-framework
 
-go 1.18
+go 1.21
+
+toolchain go1.22.0
```
PeterSchafer, line 5: Issue: Watch out for this toolchain entry, this could/will cause issues in downstream projects.

internal/presenters/test_results.gotmpl
```diff
@@ -0,0 +1,22 @@
+[1m[37m[39m[22m
```
PeterSchafer, line 1: Question: How do we consider/support different TTY types that have different capabilities?

thisislawatts: There are two options we could take here:
* strip them out at a later date in the process
* Adopt a library like https://github.com/charmbracelet/lipgloss which handles this automagically for us 💜

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -0,0 +1,156 @@
...
+	"slices"
+	"text/template"
+
+	"github.com/snyk/code-client-go/sarif"
+)
+
+//go:embed test_results.gotmpl
+var templateString string
+
+type SarifResults struct {
```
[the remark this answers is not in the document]
thisislawatts: That is what I wanted to do but I wasn't sure how to address it within the struct, will take another looks, thanks for flagging!

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -0,0 +1,208 @@
...
+			severityLevel := 0
+			var rule sarif.Rule
+			for _, ruleItem := range run.Tool.Driver.Rules {
+				if ruleItem.ID == result.RuleID {
+					rule = ruleItem
+					break
+				}
+			}
+
+			if result.Level == "note" {
```
PeterSchafer, line 47: Suggestion: this is a bit of redundant logic, the conversion from sarif level to severity is already available and the ordering is also available through the test summary. Let's remove this code and keep a single source of thruth.

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -0,0 +1,208 @@
...
+	return response
+}
+
+func getTip() string {
+	return `💡 Tip
+
+To view ignored issues, use the --include-ignores option. To view ignored issues only, use the --only-ignores option.`
+}
+
+func presenterSummary(summary *json_schemas.TestSummary, meta TestMeta) string {
```
PeterSchafer, line 131: nitpick: for consistency, should it be `renderSummary()`?

pkg/local_workflows/output_workflow.go
```diff
@@ -95,39 +98,76 @@ func handleContentTypeOthers(input []workflow.Data, i int, mimeType string, outp
...
 
+func humanReadanbleSarifOutput(config configuration.Configuration, input []workflow.Data, i int, outputDestination iUtils.OutputDestination, debugLogger *zerolog.Logger, singleData []byte) {
+	var sarif sarif.SarifDocument
+	err := json.Unmarshal(singleData, &sarif)
+	if err != nil {
+		debugLogger.Println(err)
+	}
+
+	meta := presenters.TestMeta{
+		OrgName:  config.GetString(configuration.ORGANIZATION),
```
PeterSchafer, line 160: Question: Do we need to provide the slug name here instead of the id?

pkg/local_workflows/code_workflow/native_workflow.go
```diff
@@ -135,7 +134,7 @@ func sarifLevelToSeverity(level string) string {
 }
 
 // Iterate through the sarif data and create a summary out of it.
-func createCodeSummary(input *sarif.SarifDocument) *json_schemas.TestSummary {
+func CreateCodeSummary(input *sarif.SarifDocument) *json_schemas.TestSummary {
```
PeterSchafer, line 137: Suggestion: let's move this to a more common place, now that it is being re-used. similar to the content_types

internal/presenters/presenter_sarif_results_pretty.go
thisislawatts: Fix: b5f851690684c708eb69a42d84d6d1e323a4e6d4

pkg/local_workflows/output_workflow.go
thisislawatts: We will pick this item up as a separate bug fix, there seems to be some issues when converting an org slug into a uuid. Further investigation needed to understand if this issue is tied to the pre-prod env or the client.

internal/presenters/presenter_sarif_results_pretty_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestPresenterSarifResultsPretty_MediumHighIssues(t *testing.T) {
+	fd, err := os.Open("testdata/4-high-5-medium.json")
+	require.Nil(t, err)
+
+	var input sarif.SarifDocument
+
+	err = json.NewDecoder(fd).Decode(&input)
+	require.Nil(t, err)
+
+	result, err := presenters.PresenterSarifResultsPretty(input, testMeta)
```
PeterSchafer, line 60: Suggestion: let's be explicit about the colorprofile in all tests, these are all without color
`lipgloss.SetColorProfile(termenv.TrueColor)`

thisislawatts: Addressed in https://github.com/snyk/go-application-framework/pull/137/commits/51a84f0ac9009970ab2ce75c4f9d25e2d2cb238e

PeterSchafer (approved):

PR #138, opened 2024-04-10
feat: introduce test summary schema

thisislawatts:
Document the schema and content type for extensions to use when outputting test results.

### DIFF
--- a/pkg/local_workflows/json_schemas/data_test_summary.go
+++ b/pkg/local_workflows/json_schemas/data_test_summary.go
@@ -1,0 +1,51 @@
+package json_schemas
+
+const TestSummarySchema = `{
+	"$schema": "http://json-schema.org/draft-07/schema#",
+	"title": "Test Results",
+	"description": "Schema for security test results with severity and counts",
+	"type": "object",
+	"required": ["results", "type"],
+	"properties": {
+	  "results": {
+		"type": "array",
+		"minItems": 1,
+		"items": {
+		  "type": "object",
+		  "required": ["severity", "total", "open", "ignored"],
+		  "properties": {
+			"severity": {
+			  "type": "string",
+			  "enum": ["high", "medium", "low", "critical"]
+			},
+			"total": {
+			  "type": "integer",
+			  "minimum": 0
+			},
+			"open": {
+			  "type": "integer",
+			  "minimum": 0
+			},
+			"ignored": {
+			  "type": "integer",
+			  "minimum": 0
+			}
+		  }
+		}
+	  },
+	  "type": {
+		"type": "string",
+		"enum": ["sast", "sbom"]
+	  }
+	}
+  }`
+
+type TestSummary struct {
+	Results []struct {
+		Severity string `json:"severity"`
+		Total    int    `json:"total"`
+		Open     int    `json:"open"`
+		Ignored  int    `json:"ignored"`
+	} `json:"results"`
+	Type string `json:"type"`
+}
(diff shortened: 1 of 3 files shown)


pkg/local_workflows/json_schemas/data_test_summary.go
```diff
@@ -0,0 +1,51 @@
+package json_schemas
+
+const TestSummarySchema = `{
```
cmars, line 3: What options might be available to us to further integrate this schema with the CLI codebase? Wondering specifically about:

- Generating Go "wire-format" structure types from the schema (so it becomes SSoT)
- Validating responses against it, at runtime if it's performant enough, in tests or in "strict mode" otherwise

pkg/workflow/constants.go
```diff
@@ -0,0 +1,6 @@
+package workflow
+
+const (
+	// ScanDoneEventSchema is the JSON schema for the scan done event.
```
cmars, line 4: `ScanDoneEventSchema`? This doesn't match the symbol... is there something missing here?

thisislawatts: Hm… weird ghost in the machine. Removed :broom:

pkg/local_workflows/json_schemas/data_test_summary.go
thisislawatts: I think this introduces the opportunities to set those up but I will be honest I need to do some more research on the tooling options here. Both of those sound interesting but having some types is a step in the right direction. 

A quick search surfaces these:
* https://github.com/omissis/go-jsonschema
* https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema/v5

One opportunity for runtime validation could be here:
https://github.com/snyk/cli/pull/5157/files#diff-9c2017eb57391e3dc8da9d6c7af7df0ee0fdb23aa039485a3876179d37070a08R195-R197

j-luong: question: instead of validating responses at runtime, would it be better/sufficient to couple the schema to a workflow test? i.e. a test to validate a workflow output against its schema.

this way we can be sure that the schema matches the output and so the schema can be used as the SSOT/contract to hand over to dependencies.

PeterSchafer: > question: instead of validating responses at runtime, would it be better/sufficient to couple the schema to a workflow test? i.e. a test to validate a workflow output against its schema.

I think we will not be able to not validate the schema, just due to the plug and play nature with extensions. When consuming the data for example in the output module we want to ensure that the data has the format as expected.

j-luong (approved):

pkg/workflow/json_schemas/data_test_summary.go
```diff
@@ -0,0 +1,51 @@
...
+	  "results": {
+		"type": "array",
+		"minItems": 1,
+		"items": {
+		  "type": "object",
+		  "required": ["severity", "total", "open", "ignored"],
+		  "properties": {
+			"severity": {
+			  "type": "string",
+			  "enum": ["high", "medium", "low", "critical"]
```
PeterSchafer, line 19: Question: Does it have to be an enum?

PeterSchafer: I mean do we actually care about the valiue?

thisislawatts: I think you're right in this instance it doesn't actually matter as we have nothing looking at the severity value yet. My preference is to use this schema to document what we _expect_ to get here. We can always relax to a string if needed, but it would be harder to move in the other direction.

PR #140, opened 2024-04-10
chore: support .gitignores when sending files to code-client-go

j-luong:
Utilises the glob pattern matching logic in [snyk-ls](https://github.com/snyk/snyk-ls/tree/main/infrastructure/filefilter) to filter files based on the contents of a `.gitignore` file. The filtered list of files is then sent to `code-client-go` for analysis.


thisislawatts (approved):
**praise:** Looks great, thanks for getting this added.

thisislawatts:
**nitpick:** As we are introducing capabilities to GAF, should the commit prefix for this for a `fix|feat`?

pkg/utils/file_filter.go
```diff
@@ -0,0 +1,167 @@
...
+				filteredFilesCh <- file
+			}
+		}
+	}()
+
+	return filteredFilesCh
+}
+
+// buildGlobs iterates a list of ignore filesToFilter and returns a list of glob patterns that can be used to test for ignored filesToFilter
+func (fw *FileFilter) buildGlobs(ignoreFiles []string) ([]string, error) {
```
PeterSchafer, line 89: Suggestion: all the functions here and below don't require a receiver as they don't persist a state. Just be clear about this and remove the receivers.

PR #142, opened 2024-04-10
chore: move schemas to common place

PeterSchafer:


thisislawatts:
Sharing for visibility. 
> The `workflow` package is a generic package for workflow orchestration and shouldn’t contain domain specific things.

PR #144, opened 2024-04-11
feat: integrate experimental go native code client [CLI-224]

PeterSchafer:


pkg/app/app.go
```diff
@@ -37,6 +38,72 @@ func defaultFunc_FF_CODE_CONSISTENT_IGNORES(engine workflow.Engine, config confi
...
+		if err != nil {
+			logger.Warn().Err(err).Str(configuration.API_URL, urlString).Msg("failed to get api url")
+		}
+		return apiString
+	}
+	return callback
+}
+
+func defaultInputDirectory() configuration.DefaultValueFunction {
+	callback := func(existingValue interface{}) interface{} {
```
j-luong, line 93: question: is this a go pattern? do we need to declare/assign `callback` in these methods if we just return it? Could we not just return the callback: `return func(existingValue interface{}) interface{} { ...`

thisislawatts (approved):

pkg/local_workflows/code_workflow/native_workflow.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	codeScanner := codeclient.NewCodeScanner(httpClient, codeScannerConfig, codeInstrumentor, codeErrorReporter, logger)
+	result, _, err = codeScanner.UploadAndAnalyze(ctx, interactionId, path, files, changedFiles)
+	return result, err
+}
+
+// todo: recursively iterate and filter
+func getFilesForPath(path string) (<-chan string, error) {
+	filter := utils.NewFileFilter(path)
+	rules, err := filter.GetRules([]string{})
```
j-luong, line 93: question: we're not supplying any rule files here so it will only ignore `.git` files by default.

j-luong (approved):

pkg/app/app.go
PeterSchafer: would be possible as well.

PR #146, opened 2024-04-18
feat: support --include-ignores flag

thisislawatts:
Introduce support for hiding ignored items by default, this aligns with the current behaviour of the Typescript implementation. 

The new functionality here is to introduce 2 new flags `--include-ignores` and `--only-ignores` which can be used to toggle which items are displayed in the output. 

These screengrabs show the different flags and resulting output. 

By default we only show the open issues, there are four items visible:
Demo001

Using `--include-ignores` will show also the ignored issue, there are now five items visible:
Demo002

Using `--only-ignores` will only show the ignored issue(s), there is now one item visible:
Demo003

### DIFF
--- a/internal/presenters/components.go
+++ b/internal/presenters/components.go
@@ -50,24 +50,21 @@ func RenderFindings(findings []Finding, showIgnored bool, showOpen bool) string 
 }
 
 func RenderFinding(finding Finding) string {
-	ignoredProperties := "\n"
+	properties := "\n"
 	titlePrefix := "✗ "
 
 	if finding.Ignored {
 		titlePrefix = "! [ IGNORED ] "
-		ignoredProperties = getIgnoredProperties(finding, ignoredProperties)
-	} else {
-
 	}
+
+	properties = getFormattedProperties(finding.Properties)
 
 	return strings.Join([]string{
 		fmt.Sprintf(" %s %s",
 			renderInSeverityColor(finding.Severity, fmt.Sprintf("%s[%s]", titlePrefix, strings.ToUpper(finding.Severity))),
 			renderBold(finding.Title),
 		),
-		fmt.Sprintf("   Path: %s, line %d", finding.Path, finding.Line),
-		fmt.Sprintf("   Info: %s", finding.Message),
-		ignoredProperties,
+		properties,
 	}, "\n")
 }
 
@@ -77,6 +74,29 @@ func RenderDivider() string {
 
 func RenderTitle(str string) string {
 	return fmt.Sprintf("\n%s\n\n", renderBold(str))
+}
+
+func getFormattedProperties(properties []FindingProperty) string {
+	formattedProperties := ""
+	labelLength := 0
+
+	for _, property := range properties {
+		if len(property.Label) > labelLength {
+			labelLength = len(property.Label) + 1
+		}
+	}
+
+	labelAndPropertyFormat := "   %-" + fmt.Sprintf("%d", labelLength) + "s %s\n"
+
+	for _, property := range properties {
+		if property.Label == "" {
+			formattedProperties += "\n"
+			continue
+		}
+		formattedProperties += fmt.Sprintf(labelAndPropertyFormat, property.Label+":", property.Value)
+	}
+
+	return formattedProperties
 }
 
 func RenderTip(str string) string {
--- a/internal/presenters/components.go
+++ b/internal/presenters/components.go
@@ -102,7 +102,7 @@ func RenderTip(str string) string {
 	return fmt.Sprintf("\n💡 Tip\n\n%s", str)
 }
 
-func RenderSummary(summary *json_schemas.TestSummary, orgName string, testPath string) string {
+func RenderSummary(summary *json_schemas.TestSummary, orgName string, testPath string) (string, error) {
 	var buff bytes.Buffer
 	var summaryTemplate = template.Must(template.New("summary").Parse(`Test Summary
 
@@ -159,8 +159,8 @@ func RenderSummary(summary *json_schemas.TestSummary, orgName string, testPath s
 		IgnoredIssueCountWithSeverities: ignoredIssueCountWithSeverities,
 	})
 	if err != nil {
-		return fmt.Sprintf("failed to execute summary template: %v", err)
+		return "", fmt.Errorf("failed to generete test summary from template: %w", err)
 	}
 
-	return boxStyle.Render(buff.String())
+	return boxStyle.Render(buff.String()), nil
 }
--- a/internal/presenters/components.go
+++ b/internal/presenters/components.go
@@ -1,0 +1,145 @@
+package presenters
+
+import (
+	"bytes"
+	"fmt"
+	"slices"
+	"strconv"
+	"strings"
+	"text/template"
+
+	"github.com/snyk/go-application-framework/pkg/local_workflows/json_schemas"
+)
+
+func RenderFindings(findings []Finding, showIgnored bool, showOpen bool) string {
+	if len(findings) == 0 {
+		return ""
+	}
+
+	response := ""
+
+	if showOpen {
+		response += RenderTitle("Open Issues")
+
+		for _, finding := range findings {
+			if finding.Ignored {
+				continue
+			}
+			response += RenderFinding(finding)
+		}
+	}
+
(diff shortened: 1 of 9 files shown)


internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -73,7 +77,7 @@ type TestMeta struct {
 	TestPath string
 }
 
-func PresenterSarifResultsPretty(input sarif.SarifDocument, meta TestMeta) (string, error) {
+func PresenterSarifResultsPretty(input sarif.SarifDocument, meta TestMeta, showIgnored bool, showOpen bool) (string, error) {
```
PeterSchafer, line 80: Suggestion: possible use case for the functional options pattern?! :)

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -84,22 +88,30 @@ Testing %s ...
...
+		renderFindings(SortFindings(findings, summary.SeverityOrderAsc), showIgnored, showOpen),
 		renderSummary(summary, meta),
 		getTip(),
 	)
 
 	return str, nil
 }
 
-func renderFindings(findings []Finding) string {
+func renderFindings(findings []Finding, showIgnored bool, showOpen bool) string {
```
PeterSchafer, line 99: Issue: Based on the design, if ignored issues are shown, they should get an additional `[ IGNORED ]` indicator.

internal/presenters/presenter_sarif_results_pretty_test.go
```diff
@@ -76,8 +76,40 @@ func TestPresenterSarifResultsPretty_MediumHighIssuesWithColor(t *testing.T) {
...
 
 	lipgloss.SetColorProfile(termenv.TrueColor)
-	result, err := presenters.PresenterSarifResultsPretty(input, testMeta)
+	result, err := presenters.PresenterSarifResultsPretty(input, testMeta, false, true)
 
 	require.Nil(t, err)
 	snaps.MatchSnapshot(t, result)
 }
+
+func TestPresenterSarifResultsPretty_DefaultHideIgnored(t *testing.T) {
```
PeterSchafer, line 85: Suggestion: there is no test case for `true, false`

internal/presenters/presenter_sarif_results_pretty.go
PeterSchafer: Also it seems that the additional data describing the ignore reason etc is not shown.

PeterSchafer (changes requested):
There seem to be differences to the design and it is not clear they are on purpose or accidental.

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -84,22 +88,30 @@ Testing %s ...
 %s
 `,
 		meta.TestPath,
-		renderFindings(SortFindings(findings, summary.SeverityOrderAsc)),
+		renderFindings(SortFindings(findings, summary.SeverityOrderAsc), showIgnored, showOpen),
 		renderSummary(summary, meta),
 		getTip(),
```
PeterSchafer, line 93: Issue: the output of `getTip()` needs to adapt based on the setting, see design.

internal/presenters/presenter_sarif_results_pretty_test.go
thisislawatts: Yes this case is currently covered by a workflow level test so wasn't convinced we wanted duplicate test coverage.

pkg/local_workflows/output_workflow.go
```diff
@@ -19,10 +19,12 @@ import (
...
 const (
-	OUTPUT_CONFIG_KEY_JSON       = "json"
-	OUTPUT_CONFIG_KEY_JSON_FILE  = "json-file-output"
-	OUTPUT_CONFIG_KEY_SARIF      = "sarif"
-	OUTPUT_CONFIG_KEY_SARIF_FILE = "sarif-file-output"
+	OUTPUT_CONFIG_KEY_JSON            = "json"
+	OUTPUT_CONFIG_KEY_JSON_FILE       = "json-file-output"
+	OUTPUT_CONFIG_KEY_SARIF           = "sarif"
+	OUTPUT_CONFIG_KEY_SARIF_FILE      = "sarif-file-output"
+	OUTPUT_CONFIG_KEY_INCLUDE_IGNORES = configuration.FLAG_INCLUDE_IGNORES
```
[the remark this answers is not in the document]
thisislawatts: Removed

internal/presenters/components.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return formattedProperties
+}
+
+func RenderTip(str string) string {
+	return fmt.Sprintf("\n💡 Tip\n\n%s", str)
+}
+
+func RenderSummary(summary *json_schemas.TestSummary, orgName string, testPath string) string {
```
thisislawatts, line 105: **suggestion:** 
```suggestion
func RenderTestSummary(summary *json_schemas.TestSummary, orgName string, testPath string) string {
```

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -55,129 +115,88 @@ func convertSarifToFindingsList(input sarif.SarifDocument) []Finding {
...
+				ignoreProperties = append(ignoreProperties, FindingProperty{
+					Label: "Expiration",
+					Value: expiration,
+				})
 
-`, renderInSeverityColor(finding.Severity, fmt.Sprintf("✗ [%s]", strings.ToUpper(finding.Severity))), renderBold(finding.Title), finding.Path, finding.Line, finding.Message)
-	}
+				ignoreProperties = append(ignoreProperties, FindingProperty{
+					Label: "Category",
+					Value: strings.Replace(string(suppression.Properties.Category), "wont-fix", "Won't fix", 1),
```
PeterSchafer, line 153: Suggestion: use something like this https://pkg.go.dev/golang.org/x/text/cases instead of a known replacements

internal/presenters/presenter_sarif_results_pretty.go
```diff
@@ -55,129 +115,88 @@ func convertSarifToFindingsList(input sarif.SarifDocument) []Finding {
...
 
-	return response
-}
+				// TODO: Verify date formatting
+				s, err := time.Parse(time.RFC3339, suppression.Properties.IgnoredOn)
 
-func getTip() string {
-	return `
-💡 Tip
+				if err == nil {
```
PeterSchafer, line 159: Suggestion: How about printing the original timestamp when we experience a parser error

internal/presenters/presenter_sarif_results_pretty.go
thisislawatts: Will pick this up in a future iteration, we need to understand all of the variations available here 👍🏻

internal/presenters/components.go
```diff
@@ -0,0 +1,166 @@
...
+		response += RenderTitle("Ignored Issues")
+
+		for _, finding := range findings {
+			if !finding.Ignored {
+				continue
+			}
+			response += RenderFinding(finding)
+		}
+
+		response += RenderTip("Ignores are currently managed in the Snyk Web UI.\nTo edit or remove the ignore please go to: https://app.snyk.io/") + "\n"
```
PeterSchafer, line 46: Suggestion: Replace the hard-coded URL by the configuration value `WEB_APP_URL` or even better a direct link to edit ignores if it is available.

internal/presenters/components.go
```diff
@@ -0,0 +1,166 @@
...
+		Org:                             orgName,
+		TestPath:                        testPath,
+		Type:                            testType,
+		TotalIssueCount:                 totalIssueCount,
+		IgnoreIssueCount:                ignoredIssueCount,
+		OpenIssueCountWithSeverities:    openIssueCountWithSeverities,
+		IgnoredIssueCountWithSeverities: ignoredIssueCountWithSeverities,
+	})
+	if err != nil {
+		return fmt.Sprintf("failed to execute summary template: %v", err)
```
PeterSchafer, line 162: Issue: We shouldn't just print errors to stdout, let's return an error object in addition to the string.

PeterSchafer (approved):

PR #148, opened 2024-04-19
feat: use new API endpoint for org retrieval [IDE-221]

bastiandoetsch:
This changes the API call to retrieve an org by slugname to the new endpoint at /rest/orgs

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -5,11 +5,12 @@ package api
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/snyk/go-application-framework/internal/api/contract"
-	"github.com/snyk/go-application-framework/internal/constants"
 	"io"
 	"net/http"
 	url2 "net/url"
+
+	"github.com/snyk/go-application-framework/internal/api/contract"
+	"github.com/snyk/go-application-framework/internal/constants"
 )
 
 type ApiClient interface {
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -5,11 +5,11 @@ package api
 import (
 	"encoding/json"
 	"fmt"
-	"io"
-	"net/http"
-
 	"github.com/snyk/go-application-framework/internal/api/contract"
 	"github.com/snyk/go-application-framework/internal/constants"
+	"io"
+	"net/http"
+	url2 "net/url"
 )
 
 type ApiClient interface {
@@ -25,7 +25,10 @@ type snykApiClient struct {
 }
 
 func (a *snykApiClient) GetOrgIdFromSlug(slugName string) (string, error) {
-	url := a.url + "/v1/orgs"
+	apiVersion := "version=2024-03-12"
+	escapedSlug := url2.PathEscape(slugName)
+	url := fmt.Sprintf("%s/rest/orgs?%s&slug=%s", a.url, apiVersion, escapedSlug)
+
 	res, err := a.client.Get(url)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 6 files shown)


internal/api/api.go
```diff
@@ -25,40 +26,48 @@ type snykApiClient struct {
...
+}
+
 func (a *snykApiClient) GetDefaultOrgId() (string, error) {
 	url := a.url + "/rest/self?version=" + constants.SNYK_API_VERSION
 	res, err := a.client.Get(url)
 	if err != nil {
 		return "", fmt.Errorf("unable to retrieve org ID: %w", err)
 	}
-	defer res.Body.Close()
+	defer func() { _ = res.Body.Close() }()
```
cmars, line 70: I don't understand the value this adds. The original defer was fine (ignoring a close error is reasonable).

Do we need to adjust our linter?

internal/api/api.go
```diff
@@ -25,40 +26,48 @@ type snykApiClient struct {
...
-
 	return "", fmt.Errorf("org ID not found for slug %v", slugName)
 }
 
+func GetOrgsApiURL(baseURL, slugName string) string {
+	apiVersion := "version=2024-03-12"
+	escapedSlug := url2.PathEscape(slugName)
+	url := fmt.Sprintf("%s/rest/orgs?%s&slug=%s", baseURL, apiVersion, escapedSlug)
+	return url
+}
```
cmars, line 62: Suggest constructing a `*url.URL` instead, and using the API to set the query parameters, letting the stdlib deal with escaping. Something like this:

```suggestion
func OrgsApiURL(baseURL, slugName string) (*url.URL, error) {
	u, err := url2.Parse(baseURL+"/rest/orgs")
	if err != nil {
		return nil, err
	}
	q := u.Query()
	q.Set("version", orgsAPIVersion)  // use a constant for the orgs API
	q.Set("slug", slugName)
	return u, nil
}
```

Other variations, you could accept `baseURL` as a *url.URL, copy the struct and append the path (avoiding the need for a parse).

I'd also drop the `Get` prefix (https://go.dev/doc/effective_go#Getters). `New` might be ok if it's creating a *url.URL.

internal/api/api_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/pkg/errors"
```
cmars, line 9: Error wrapping is great, but this package may not be well-maintained. Not something to address here (we're already using it), but I'd like us to follow up with a better alternative across LDX...

cmars (commented):
Some suggestions below.

internal/api/api.go
bastiandoetsch: Or deactivate the IntelliJ linting rule.

internal/api/api.go
bastiandoetsch: Done. Only added the Get prefix, because it's rather common in go-application-framework, even though it's not golang-style. But yeah, no worries, dropped it.

internal/api/api.go
```diff
@@ -25,39 +26,58 @@ type snykApiClient struct {
...
 	if err != nil {
 		return "", err
 	}
+
+	res, err := a.client.Get(u.String())
+	if err != nil {
+		return "", err
+	}
+
+	//goland:noinspection GoUnhandledErrorResult
```
cmars, line 39: Not everyone uses GoLand so I have mixed feelings about this. Could we standardize on golangci-lint as the single-source of truth for code quality & standards?

cmars (approved):
LGTM

internal/api/api.go
bastiandoetsch: Sure, wondering how to handle it without restricting our capability to leverage Golands quickfixes, though.

PR #149, opened 2024-04-19
fix: replace legacy endpoint used by whoami

mgyorke:
Replaced the legacy endpoint  /v1/user/me with the rest equivalent /rest/self in the whoami workflow

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -18,6 +18,8 @@ type ApiClient interface {
 	GetOrgIdFromSlug(slugName string) (string, error)
 	Init(url string, client *http.Client)
 	GetFeatureFlag(flagname string, origId string) (bool, error)
+	GetUserMe() (string, error)
+	GetSelf() (contract.SelfResponse, error)
 }
 
 type snykApiClient struct {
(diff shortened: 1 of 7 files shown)


pkg/local_workflows/whoami_workflow_test.go
```diff
@@ -71,7 +80,7 @@ func Test_WhoAmI_whoAmIWorkflowEntryPoint_happyPath(t *testing.T) {
 
 	mockClient := newTestClient(func(req *http.Request) *http.Response {
 		// Test request parameters
-		assert.Equal(t, "/v1/user/me", req.URL.String())
+		assert.Equal(t, "/rest/self?version=2024-02-26%7Eexperimental", req.URL.String())
```
thisislawatts, line 83: **suggestion:** (non-blocking) Maybe we could introduce a shared constant here to store this value and reduce the number of changes when updating this in the future.

pkg/local_workflows/whoami_workflow.go
```diff
@@ -56,7 +56,7 @@ func whoAmIWorkflowEntryPoint(invocationCtx workflow.InvocationContext, _ []work
 
 	// define userme API userMeEndpoint
 	baseUrl := config.GetString(configuration.API_URL)
-	url := baseUrl + apiVersion + userMeEndpoint
+	url := fmt.Sprintf("%s%s?version=%s", baseUrl, restSelfEndpoint, apiVersion)
 
 	// call userme API userMeEndpoint
 	userMe, err := fetchUserMe(httpClient, url, logger)
```
thisislawatts, line 62: **suggestion:** We discussed during a call, move fetchUserMe across to `api.go` so that is collocated with our other Snyk API transactions. 

Also look at bumping the endpoint version of /rest/self used by the CLI to a more recent version.

internal/api/api.go
```diff
@@ -72,64 +74,90 @@ func OrgsApiURL(baseURL, slugName string) (*url.URL, error) {
...
+	}
+
+	if err = json.Unmarshal(body, &selfData); err != nil {
+		return selfData, fmt.Errorf("unable to retrieve self data: %w", err)
+	}
+
+	return selfData, nil
+}
+
+func clientGet(a *snykApiClient, endpoint string) ([]byte, error) {
```
thisislawatts, line 140: **suggestion:** Extend interface to include version as attribute and move URL building into clientGet. eg endpoint, version => `${endpoint}?version=${version}`

thisislawatts (approved):

PR #150, opened 2024-04-22
refactor: use function options pattern

teodora-sandu:
Updates the code-client-go library and picks up the logging improvements and function options pattern changes. We have also introduced a fallback instrumentor and error reporter so that the CLI doesn't need to create dummy ones. If you want to override them you can still do, but for now the fallbacks do exactly what the ones implemented here do.


PeterSchafer (approved):
Nice! Thank you!

PR #152, opened 2024-04-25
feat: CLI renders real findings for consistent ignores

teodora-sandu:
The LS can now upgrade to the latest code-client-go version, so that we return the real findings for repos from the Analysis API.

I will also update the CLI once https://github.com/snyk/snyk-ls/pull/497 is also merged.

### DIFF
--- a/pkg/local_workflows/output_workflow_test.go
+++ b/pkg/local_workflows/output_workflow_test.go
@@ -338,14 +338,7 @@ func getSarifInput() sarif.SarifDocument {
 									Level: "error",
 								},
 								Properties: sarif.RuleProperties{
-									Tags: []string{"tag1", "tag2"},
-									ShortDescription: sarif.ShortDescription{
-										Text: "Rule 1 description",
-									},
-									Help: struct {
-										Markdown string `json:"markdown"`
-										Text     string `json:"text"`
-									}{Markdown: "", Text: ""},
+									Tags:               []string{"tag1", "tag2"},
 									Categories:         []string{},
 									ExampleCommitFixes: []sarif.ExampleCommitFix{},
 									Precision:          "",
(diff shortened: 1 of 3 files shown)


pkg/local_workflows/output_workflow_test.go
```diff
@@ -338,14 +338,7 @@ func getSarifInput() sarif.SarifDocument {
...
 								Properties: sarif.RuleProperties{
-									Tags: []string{"tag1", "tag2"},
-									ShortDescription: sarif.ShortDescription{
-										Text: "Rule 1 description",
-									},
-									Help: struct {
-										Markdown string `json:"markdown"`
-										Text     string `json:"text"`
-									}{Markdown: "", Text: ""},
+									Tags:               []string{"tag1", "tag2"},
```
teodora-sandu, line 341: These fields are already on the rule itself, so I don't think they're required in the property bag

PeterSchafer: Question: did this break something?

teodora-sandu: Do you mean by manually testing GAF? Here are screenshots.

- Without FF
Screenshot 2024-04-25 at 17 05 53

- With FF
Screenshot 2024-04-25 at 17 06 12

PeterSchafer: I mean it is just some test data for a unit test, so I was wondering if it hurts that it is there.

PeterSchafer: I just think that the test case has nothing to do with your changes, so maybe we don't need to remove the data, but actually if the tests still pass it is probably obsolete.

teodora-sandu: Ah I get you now. I changed the test case because the fields have been removed from the type. So the tests were failing as a result

teodora-sandu: Here are the changes in `code-client-go`, for reference: https://github.com/snyk/code-client-go/pull/37

teodora-sandu:
The println has been removed and I also updated a log so it doesn't include the `findings_url`, as per [this](https://snyk.slack.com/archives/C06MZKWG0NB/p1714064382149879) thread.
Screenshot 2024-04-26 at 10 48 10
Screenshot 2024-04-26 at 10 49 15

PeterSchafer (approved):
Nice!

PR #154, opened 2024-05-02
feat(code_workflow): support .snyk style excludes when applying filtering rules

j-luong:
This PR should add support for `.snyk` style excludes when applying filtering rules for the `code_workflow`

To test locally:
- use local version of GAF in CLI `replace github.com/snyk/go-application-framework => ../../go-application-framework` in `go.mod`
- build a local cli binary `make clean build`
- add a `.snyk` file to repository that has a valid `excludes` section
- run local cli binary `<path/to/local/cli/binary> code test`
- validate that `.snyk` `excludes` are filtered

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"fmt"
+	"gopkg.in/yaml.v3"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


pkg/utils/file_filter.go
```diff
@@ -117,6 +123,33 @@ func parseIgnoreFile(content []byte, filePath string) (ignores []string) {
...
+		Exclude struct {
+			Code   []string `yaml:"code"`
+			Global []string `yaml:"global"`
+		} `yaml:"exclude"`
+	}
+
+	var rules DotSnykRules
+	err := yaml.Unmarshal(content, &rules)
+	if err != nil {
+		fmt.Printf("parse .snyk failed: %v", err)
```
thisislawatts, line 138: **issue(blocking):** Rather than outputting to fmt.Printf, can we pass a logger through to this method.

thisislawatts (approved):

PR #156, opened 2024-05-08
fix: handle different get self api responses between user and service account CLI-21

mgyorke:


internal/api/api.go
```diff
@@ -85,11 +85,22 @@ func (a *snykApiClient) GetUserMe() (string, error) {
 		return "", fmt.Errorf("error while fetching self data: %w", err) // Prioritize error
 	}
 
-	if len(selfData.Data.Attributes.Username) == 0 {
-		return "", fmt.Errorf("error while extracting user: missing property 'username'")
+	// according to API spec for get /self
+	// username is not a mandatory field, only name and email
+	// service accounts contain only name (name of the service account token) and org_context uuid
```
thisislawatts, line 90: **nitpick (non-blocking):** Maybe we could include a link to the public facing documentation for this endpoint

pkg/local_workflows/whoami_workflow_test.go
```diff
@@ -249,7 +249,7 @@ func Test_WhoAmI_whoAmIWorkflowEntryPoint_extractUserFailures(t *testing.T) {
 	networkAccessMock.EXPECT().GetHttpClient().Return(mockClient).AnyTimes()
 
 	// setup
-	expectedError := errors.New("error fetching user data: error while extracting user: missing property 'username'")
+	expectedError := errors.New("error fetching user data: error while extracting user: missing properties username/name/email")
```
thisislawatts, line 253: **issue (blocking):** Introduce a test that verifies we correctly handle the Service account response payload from the /self endpoint. This was the issue that is currently blocking upgrading the go-application-framework.

thisislawatts (changes requested):

internal/api/api.go
mgyorke: good idea, done.

thisislawatts (approved):

PR #158, opened 2024-05-09
test(whoami): refactor for rest api specific cases - CLI-21

mgyorke:
This PR contains:
- refactoring of the tests for the whoami workflow and coverage for new rest API cases
- a fix for an edge case spotted during testing, when specifying --json output should also return name

### DIFF
--- a/pkg/local_workflows/whoami_workflow.go
+++ b/pkg/local_workflows/whoami_workflow.go
@@ -68,6 +68,7 @@ func whoAmIWorkflowEntryPoint(invocationCtx workflow.InvocationContext, _ []work
 			Id:       &selfRes.Data.Id,
 			UserName: &selfRes.Data.Attributes.Username,
 			Email:    &selfRes.Data.Attributes.Email,
+			Name:     &selfRes.Data.Attributes.Name,
 		}
 
 		// parse response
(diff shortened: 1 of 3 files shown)


thisislawatts (approved):

pkg/local_workflows/whoami_workflow.go
```diff
@@ -68,6 +68,7 @@ func whoAmIWorkflowEntryPoint(invocationCtx workflow.InvocationContext, _ []work
 			Id:       &selfRes.Data.Id,
 			UserName: &selfRes.Data.Attributes.Username,
 			Email:    &selfRes.Data.Attributes.Email,
+			Name:     &selfRes.Data.Attributes.Name,
```
PeterSchafer, line 71: Issue: This field was not in the UserMe data structure. For Service Accounts the Service Account Name was in User Name. Since the intention here is to map the new data to the old data structure, I would propose to map Name to UserName instead of adding a new field.

PeterSchafer: @thisislawatts you already merged these changes. was this something that was considered?

thisislawatts: A good question, I suspect we are missing some clear user requirements/expectations here around the response schema for the whoami command. Are we are looking to align ourselves to the upstream REST payload schema or introduce a subtly different interpretation. If so, what are the user needs here, aligning on Service or User account as the desired target state? 

I think it would be worth having this conversation more broadly, perhaps as part of an exercise to promote the `snyk whoami` from an --experimental state to generally available command.

PR #161, opened 2024-05-13
feat: introduce empty state for ignores [CLI-311]

thisislawatts:
When there are no ignores returned we should include a clear
label to state this rather than create ambiguity as to whether
something was broken.

Screenshot 2024-05-13 at 17 12 38

### DIFF
--- a/internal/presenters/components.go
+++ b/internal/presenters/components.go
@@ -31,11 +31,20 @@ func RenderFindings(findings []Finding, showIgnored bool) string {
 		response += RenderDivider()
 		response += RenderTitle("Ignored Issues")
 
+		ignoredFindings := ""
+
 		for _, finding := range findings {
 			if !finding.Ignored {
 				continue
 			}
-			response += RenderFinding(finding)
+			ignoredFindings += RenderFinding(finding)
+		}
+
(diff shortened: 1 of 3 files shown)


internal/presenters/components.go
```diff
@@ -31,11 +31,20 @@ func RenderFindings(findings []Finding, showIgnored bool) string {
...
+		}
+
+		fmt.Printf("ignored findings: %s", ignoredFindings)
+		if ignoredFindings == "" {
+			response += renderBold("  There are no ignored issues\n")
+		} else {
+			response += ignoredFindings
 		}
 
 		response += RenderTip("Ignores are currently managed in the Snyk Web UI.\nTo edit or remove the ignore please go to: https://app.snyk.io/") + "\n"
```
j-luong, line 50: question: is it better to link to a doc page on editing/removing ignores rather than `https://app.snyk.io/`?

j-luong (approved):

internal/presenters/components.go
thisislawatts: A good spot, either documentation or a link directly to an Application page where the user can apply ignores.

PR #165, opened 2024-05-13
fix: add bold weight to summary output [CLI-311]

thisislawatts:
Small nit, ensure Test Summary title is rendered in bold text. 

Screenshot 2024-05-13 at 17 01 54

### DIFF
--- a/internal/presenters/__snapshots__/presenter_sarif_results_pretty_test.snap
+++ b/internal/presenters/__snapshots__/presenter_sarif_results_pretty_test.snap
@@ -161,7 +161,7 @@ To view ignored issues, use the --include-ignores option.
 
 
 ╭──────────────────────────────────────────────╮
-│ Test Summary                                 │
+│ [1mTest Summary[0m                                 │
 │                                              │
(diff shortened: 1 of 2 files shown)


internal/presenters/__snapshots__/presenter_sarif_results_pretty_test.snap
```diff
@@ -161,7 +161,7 @@ To view ignored issues, use the --include-ignores option.
 
 
 ╭──────────────────────────────────────────────╮
-│ Test Summary                                 │
+│ [1mTest Summary[0m                                 │
```
j-luong, line 164: nitpick: why is that pipe `|` out of line

j-luong (approved):

internal/presenters/__snapshots__/presenter_sarif_results_pretty_test.snap
thisislawatts: I believe it is tied to the control characters which are rendered in the snapshot but will be hidden when viewed in the terminal.

PR #166, opened 2024-05-13
feat(instrumentation): derive category vector CLI-307

mgyorke:
This PR adds the `ParseCliArgs ` function to parse command-line arguments into product/command/flag categories as per the RFC.  The function uses whitelists and known commands to determine the correct categories, defaulting to "oss" if no product is explicitly provided.

### DIFF
--- a/pkg/analytics/categories.go
+++ b/pkg/analytics/categories.go
@@ -1,0 +1,163 @@
+package analytics
+
+import (
+	"slices"
+	"strings"
+)
+
+// ParseCliArgs categorizes command-line arguments into a structured slice.
+//
+// This function receives the entire os.Args slice (including the program name at index 0)
+// and processes the arguments to determine the product, commands, and flags.
+// It is designed to handle arguments in any order.
+//
+// The function performs the following actions:
+//  1. It identifies the product based on the following rules:
+//     - If the command is "test" and no other commands are present, the product is "oss" (open-source software).
+//     - Otherwise, no product is explicitly identified.
+//  2. It extracts known commands and whitelisted flags from the arguments. Flags after a double dash ("--") are ignored.
+//  3. It constructs a slice containing the categorized arguments in the following order:
+//     - [product] (if applicable)
+//     - [command(s)]
+//     - [flags] (without leading dashes)
+//
+// If no known commands or flags are found, the returned slice will be empty.
+//
+// Example:
+//
+//	args := []string{"cli", "scan", "--debug", "--"}
+//	knownCommands := []string{"scan", "build"}
+//	flagsAllowList := []string{"debug", "verbose"}
+//	result := ParseCliArgs(args, knownCommands, flagsAllowList)
+//	// result: ["scan", "debug"]
+//
+// Parameters:
+//   - args: A slice of strings representing the entire os.Args.
+//   - knownCommands: A slice of valid command names.
+//   - flagsAllowList: A slice of valid flag names (without leading dashes).
+//
+// Returns:
+//   - A slice of strings containing the categorized arguments.
+func ParseCliArgs(args []string, knownCommands []string, flagsAllowList []string, knownProducts []string) []string {
+	commands := []string{}
+	flags := []string{}
+	result := []string{}
+
+	productFallback := "oss"
+
+	// Separate parsing of commands and flags to ensure correct ordering in the category vector
+	// regardless of how they are provided in the command line.
+	for _, arg := range args[1:] {
+		if strings.HasPrefix(arg, "-") {
+			if arg == "--" {
+				break
+			}
+
+			// Split the flag into key and potential value
+			flagParts := strings.SplitN(arg, "=", 2)
+			flagName := strings.TrimLeft(flagParts[0], "-")
+
+			if slices.Contains(flagsAllowList, flagName) {
+				flags = append(flags, flagName)
+			}
+		} else if slices.Contains(knownCommands, arg) {
+			if len(commands) == 0 && arg == "test" {
+				result = append(commands, productFallback)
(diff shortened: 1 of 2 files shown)


pkg/analytics/categories_test.go
```diff
@@ -0,0 +1,47 @@
...
+	testCases := []struct {
+		name           string
+		args           []string
+		expectedOutput []string
+		expectedError  error
+	}{
+		// Happy Paths
+		{"OSS test", []string{"snyk", "test"}, []string{"oss", "test"}, nil},
+		{"OSS test with unmanaged flag", []string{"snyk", "test", "--unmanaged"}, []string{"oss", "test", "unmanaged"}, nil},
+		{"Code test", []string{"snyk", "code", "test"}, []string{"oss", "code", "test"}, nil},
```
PeterSchafer, line 22: Issue: it is either "oss" or "code", not both at the same time.

pkg/analytics/categories_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{"IAC describe", []string{"snyk", "iac", "describe"}, []string{"oss", "iac", "describe"}, nil},
```
PeterSchafer, line 23: Issue: same as above, it is either "iac" or "oss"

pkg/analytics/categories_test.go
```diff
@@ -0,0 +1,47 @@
...
+	knownCommands := []string{"test", "code", "monitor", "iac", "rules", "describe", "sbom", "container"}
+	whitelistedFlags := []string{"unmanaged", "all-projects", "debug"}
+
+	testCases := []struct {
+		name           string
+		args           []string
+		expectedOutput []string
+		expectedError  error
+	}{
+		// Happy Paths
```
PeterSchafer, line 19: Suggestion: The test cases are really helpful understanding the implementation! One thing I noticed was, that they are very similar in their pattern. The are actually many different orders and forms how to provide arguments, so that it is a bit tricky do do that correctly. For example take this command
```
snyk -d test ../folder/ --org=myorg --remote-repo-url https://github.com/snyk
```

pkg/analytics/categories.go
```diff
@@ -0,0 +1,127 @@
...
+//
+// Behavior:
+// - Special Case: If the flag is "-d" (the shorthand for "debug"), it adds "debug" to the flags slice and returns.
+// - General Case:
+//   - Splits the flag on "=" if present to extract the flag name.
+//   - Removes leading dashes from the flag name.
+//   - Checks if the flag name (without dashes) is in the whitelist.
+//   - If whitelisted, appends the flag name (without dashes) to the flags slice.
+func (c *CliArgCategorizer) parseFlag(flag string) {
+	if flag == "-d" {
```
PeterSchafer, line 94: Suggestion: I think the approach of replacing short-hand flags like this is not very well maintainable. I would either ignore them or keep them as they are.

pkg/analytics/categories.go
```diff
@@ -0,0 +1,127 @@
...
+//   - A pointer to a new CliArgCategorizer instance, ready to be used for parsing arguments.
+//
+// Usage:
+// categorizer := NewCategoryParser(knownCommands, whitelistedFlags, knownProducts)
+// categoryVector = categorizer.CategorizeCliArgs(os.Args)
+func NewCategoryParser(knownCommands []string, flagsWhitelist []string, knownProducts []string) *CliArgCategorizer {
+	return &CliArgCategorizer{
+		commands: []string{},
+		flags:    []string{},
+		result:   []string{},
```
PeterSchafer, line 38: Suggestion: caching results like this introduces a state to the structure, this makes it impossible to use a parser concurrently. This seems an unnecessary limitation. Just use parameters instead.

pkg/analytics/categories.go
```diff
@@ -0,0 +1,127 @@
...
+//   - The first element is the product category (e.g., "oss", "code", "iac"), defaulting to "oss" if not explicitly specified.
+//   - The second element (or more) is the main command, potentially followed by any valid subcommands.
+//   - The remaining elements are whitelisted flags (without the leading dashes).
+//
+// Parameters:
+//   - args: A slice of strings representing the command-line arguments.
+//
+// Returns:
+//   - A slice of strings representing the categorized arguments.
+func (c *CliArgCategorizer) CategorizeCliArgs(args []string) []string {
```
PeterSchafer, line 62: nitpick: Naming is hard but since this is framed to be a parser, why not call this `parse()`

pkg/analytics/categories.go
mgyorke: I agree, if ignoring them is an option I would go on that direction.

internal/constants/constants.go
```diff
@@ -2,3 +2,90 @@ package constants
 
 const SNYK_DEFAULT_API_URL = "https://api.snyk.io"
 const SNYK_DEFAULT_API_VERSION = "2024-04-22"
+
+var PRODUCTS = []string{"code", "iac", "container", "sbom"}
+var KNOWN_COMMANDS = []string{"test", "code", "monitor", "iac", "rules", "describe", "sbom", "container", "debug"}
+var WHITELISTED_FLAGS = []string{
```
thisislawatts, line 8: **suggestion (blocking):** Please lets switch to ALLOWED_FLAGS here. It clearly articulates the meaning.  

Supporting material: https://www.ncsc.gov.uk/blog-post/terminology-its-not-black-and-white

PeterSchafer: In addition, I would move this to the `categories.go` file.

PeterSchafer (approved):
LGTM

PR #169, opened 2024-05-14
feat: return error for empty projects [CLI-309]

thisislawatts:
Return an error when the bundle hash is empty, this happens when the submitted folder contains no files supported by the Snyk Code product.

The ideal outcome here is that the workflow provides an easy to reason with and actionable error object for consumption by downstream clients.

Based on the client context, IDE or CLI, the client can choose whether this handled with the default error handling or instead apply some tailored experience.

For example:
* CLI currently exits with code `3` for unsupported projects
* IDE may ignore failure completely, and empty directory is a valid state at certain points in IDE lifecycle, for example creating a new project.

### DIFF
--- a/pkg/local_workflows/code_workflow/native_workflow.go
+++ b/pkg/local_workflows/code_workflow/native_workflow.go
@@ -11,6 +11,7 @@ import (
 	codeclienthttp "github.com/snyk/code-client-go/http"
 	"github.com/snyk/code-client-go/sarif"
 	"github.com/snyk/code-client-go/scan"
+	"github.com/snyk/error-catalog-golang-public/snyk_errors"
 
 	sarif2 "github.com/snyk/go-application-framework/internal/utils/sarif"
 	"github.com/snyk/go-application-framework/pkg/configuration"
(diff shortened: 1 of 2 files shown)


pkg/local_workflows/code_workflow/native_workflow.go
```diff
@@ -95,12 +99,27 @@ func defaultAnalyzeFunction(target scan.Target, httpClientFunc func() *http.Clie
...
+		return result, snyk_errors.Error{
+			ID:             "Snyk-CLI-0002",
+			Title:          "No Supported Projects Detected",
+			Classification: "ACTIONABLE",
+			Level:          "warning",
+			Detail:         "We found 0 supported files\nPlease see our documentation for Snyk Code language and framework support\n",
+			Links: []string{
+				"https://docs.snyk.io/products/snyk-code/snyk-code-language-and-framework-support",
+			},
+		}
```
thisislawatts, line 115: **issue:** Stubbed out error catalog object, introduce error upstream then import and replace here.

PR #171, opened 2024-05-14
fix(auth): automatically switch auth type if an API token is provided

PeterSchafer:


pkg/local_workflows/auth_workflow.go
```diff
@@ -57,19 +56,33 @@ func OpenBrowser(authUrl string) {
 func authEntryPoint(invocationCtx workflow.InvocationContext, _ []workflow.Data) (_ []workflow.Data, err error) {
 	// get necessary objects from invocation context
 	config := invocationCtx.GetConfiguration()
-	logger := invocationCtx.GetLogger()
+	logger := invocationCtx.GetEnhancedLogger()
```
thisislawatts, line 59: **praise:** Nice upgrade

thisislawatts (approved):
Good to get some tests introduced around the auth_workflow

pkg/configuration/constants.go
```diff
@@ -22,6 +22,7 @@ const (
 	RAW_CMD_ARGS                    string = "internal_raw_cmd_args"
 	WEB_APP_URL                     string = "internal_snyk_app"
 	INPUT_DIRECTORY                 string = "targetDirectory"
+	UNNAMED_PARAMETER               string = INPUT_DIRECTORY
```
thisislawatts, line 25: **nitpick(non-blocking):** It may be useful to move the comment around positional arguments to this declaration to improve readability.

pkg/configuration/constants.go
```diff
@@ -22,6 +22,7 @@ const (
 	RAW_CMD_ARGS                    string = "internal_raw_cmd_args"
 	WEB_APP_URL                     string = "internal_snyk_app"
 	INPUT_DIRECTORY                 string = "targetDirectory"
+	UNNAMED_PARAMETER               string = INPUT_DIRECTORY                       //UNNAMED_PARAMETER in the CLI is the positional argument
```
cmars, line 25: It seems weird that the value for this is `"targetDirectory"` when it's not a target directory.

Feels the wrong way around. The value should be `positionalArgument` and the context of that might be a target directory if that's what `positionalArgument` means to a specific command.

cmars: It might be arbitrary now in the way it works, but can easily create bugs or hard to debug situations when we have to reverse the meaning.

cmars (commented):
Concerned with the conflation of unnamed parameter & input directory in the constants.

pkg/configuration/constants.go
cmars: I'd go further and fix this... it's more than a nit to me.

pkg/configuration/constants.go
PeterSchafer: Naming is hard and especially in this case. The problem with positional arguments is, that it is something from the cli domain as it describes the concept of arguments that have a meaning based on their position. In a framework that doesn't necessarily know about the application it is used in, the best naming would be usage specific. A mapping in the cli application of the usage based name to the positional argument would be the best. I will adapt the PR accordingly.
Thank you!

PeterSchafer:
I addressed the previously expressed concerns about the variable naming and refactored the logic a bit to be more clear and support better debug logging.
There is the associated CLI PR here https://github.com/snyk/cli/pull/5240

pkg/local_workflows/auth_workflow.go
```diff
@@ -57,37 +58,42 @@ func OpenBrowser(authUrl string) {
...
-		)
+		logger.Printf("Headless: %v", headless)
+
 		err = authenticator.Authenticate()
 		if err != nil {
-			return nil, err
+			return err
 		}
 
 		fmt.Println(auth.AUTHENTICATED_MESSAGE)
```
j-luong, line 99: question: can we not use the logger here?

j-luong (approved):

PR #176, opened 2024-05-16
feat(instrumentation): add interaction id header for network requests CLI-325

mgyorke:
First PR for CLI-325. 
We generate urn identifier and attach it as a header to the network requests.

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -353,3 +353,7 @@ func DetermineStage(isCiEnvironment bool) string {
 
 	return "dev"
 }
+
+func AssembleUrnFromUUID(uuid string) string {
+	return fmt.Sprintf("urn:snyk:interaction:%s", uuid)
+}


pkg/networking/networking.go
```diff
@@ -126,6 +128,13 @@ func (n *networkImpl) AddHeaderField(key, value string) {
 }
 
 func (n *networkImpl) AddHeaders(request *http.Request) error {
+	// add snyk-request-id header with interaction id
+	uuid, err := GenerateInteractionUrn()
+	if err != nil {
+		n.logger.Warn().Err(err).Msg("Error generating Snyk request ID")
+		uuid = "urn:snyk:interaction:unknown"
+	}
+	n.AddHeaderField("snyk-request-id", uuid)
```
PeterSchafer, line 137: Issues: 
1. This will generate a new request-id for each request, the goal is to have a common request id for each interaction, which can include multiple http requests.
2. snyk-request-id has to be a UUID not an URN, the idea is to use the same UUID from the request-id as for the InteractionID.
3. Having a constant request-id, will break any request tracing, in case of an error, it is better to not have a request-id at all.

pkg/networking/networking.go
```diff
@@ -229,3 +238,11 @@ func (n *networkImpl) Clone() NetworkAccess {
 
 	return clone
 }
+
+func GenerateInteractionUrn() (string, error) {
```
PeterSchafer, line 242: Suggestion: 
* Maybe not the best place to add this method, `utils` might be better.
* GenerateInteractionUrn() could take the uuid as a parameter, so that it can be easier re-used.

pkg/networking/networking.go
PeterSchafer: The integration should be on application level, where the interaction is defined. In the CLI this is when running a command, in and IDE, this might be the click on the scan button.

mgyorke: Thank you! This cleared up things, I pushed the helper function.

PeterSchafer (approved):

PR #177, opened 2024-05-17
feat: introduce a collector for instrumentation data

PeterSchafer:


internal/api/analytics/2024-03-07/gen.go
```diff
@@ -0,0 +1,6 @@
+package v20240307
+
+//go:generate oapi-codegen -package=clients -config spec.config.yaml spec.yaml
+
+// We can't pull the spec from the remote as analytics-service is a private repo
+// //go:generate bash -c "mkdir -p 2024-03-07 && curl -s https://raw.githubusercontent.com/snyk/analytics-service/main/internal/hidden/resources/analytics/2024-03-07/spec.yaml | oapi-codegen -package=v20240307 -generate=types,client,spec -o=v20240307/client-gen.go /dev/stdin"
```
PeterSchafer, line 6: Suggestion: let's remove this!

pkg/analytics/instrumentation_collector.go
```diff
@@ -0,0 +1,243 @@
...
+	userAgent           networking.UserAgentInfo
+	interactionId       string
+	timestamp           time.Time
+	duration            time.Duration
+	stage               string
+	instrumentationType string
+	category            []string // TODO: switch to using enum?
+	status              Status
+	testSummary         json_schemas.TestSummary
+	targetId            string // TODO: switch to using purl lib?
```
PeterSchafer, line 54: Suggestion: let's remove this comment, string is an appropriate representation and parsing it can be done by using a purl lib

pkg/analytics/instrumentation_collector.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+type instrumentationCollectorImpl struct {
+	userAgent           networking.UserAgentInfo
+	interactionId       string
+	timestamp           time.Time
+	duration            time.Duration
+	stage               string
+	instrumentationType string
+	category            []string // TODO: switch to using enum?
```
PeterSchafer, line 51: Suggestion: Let's remove this comment. category contains labels which are an open set of values and can't be expressed through an enum

pkg/analytics/instrumentation_collector.go
```diff
@@ -0,0 +1,242 @@
...
+	if !hasUserAgentAppData {
+		return nil, errors.New("no user agent application data")
+	}
+
+	return &api.Runtime{
+		Application: &api.Application{
+			Name:    ic.userAgent.App,
+			Version: ic.userAgent.AppVersion,
+		},
+		Environment: &api.Environment{
```
PeterSchafer, line 174: Issue: some of the data is optional and might not be available, so we should not add empty strings here.

pkg/analytics/instrumentation_collector.go
```diff
@@ -0,0 +1,242 @@
...
+	for _, result := range testSummary.Results {
+		r = append(r, map[string]interface{}{
+			"name":  result.Severity,
+			"count": result.Total,
+		})
+	}
+	return &r
+}
+
+func toInteractionStage(s *string) *api.InteractionStage {
```
PeterSchafer, line 203: Suggestion: there is no need to have a string pointer as parameter

pkg/analytics/instrumentation_collector.go
```diff
@@ -0,0 +1,242 @@
...
+	var is api.InteractionStage
+
+	switch stage := s; *stage {
```
PeterSchafer, line 206: Suggestion: this looks like it can be made easier. Overall InteractionStage is a string, so a type cast would be sufficient. I don't see much value of enforcing correct strings here, if there are values not acceptable, the api should handle this. Otherwise we are spreading knowledge unnecessarily.

j-luong (approved):

pkg/analytics/instrumentation_collector.go
j-luong: hmm, I'm not sure. We have the [schema](https://github.com/snyk/go-application-framework/blob/661cc24c0edb47aabade15fb050b07168555610f/internal/api/clients/analytics_v20240307.go#L20-L26) which defines the valid values for this property.

IMO, we should do our best to conform to the schema contract, and so where a schema defines valid values, we should  add checks to make sure this is the case.

The alternative is we only let the API handle this validation, which will cost an API call.

PeterSchafer (commented):
LGTM!

PR #178, opened 2024-05-17
feat(instrumentation): prep work for trace level logging cli-325

mgyorke:
This PR sets the stage for implementing logging of the interaction id header at trace level on the CLI.

- Moved zerologConverter.go to a public location, we need it in the CLI side.
- Updated imports
- LogRequest utility function to be used in the CLI side.


pkg/networking/networking.go
```diff
@@ -52,6 +52,15 @@ type networkImpl struct {
 	logger       *zerolog.Logger
 }
 
+func LogRequest(r *http.Request, logger *zerolog.Logger) {
+	if logger.GetLevel() != zerolog.TraceLevel { // Only log if trace level is enabled
```
mgyorke, line 56: I'm not sure if we need this check, I made an assumption here.

PeterSchafer: It probably doesn't hurt.

PeterSchafer (approved):

PR #181, opened 2024-05-22
fix(instrumentation): CLI-335 add missing known commands

mgyorke:
Extended the list of known commands based on the [docs](https://docs.snyk.io/snyk-cli/cli-commands-and-options-summary).

### DIFF
--- a/pkg/instrumentation/categories.go
+++ b/pkg/instrumentation/categories.go
@@ -79,7 +79,28 @@ func DetermineCategoryFromArgs(args []string, knownCommands []string, flagsAllow
 	return result
 }
 
-var KNOWN_COMMANDS = []string{"test", "code", "monitor", "iac", "rules", "describe", "container", "debug", "config"}
+var KNOWN_COMMANDS = []string{
+	"test",
+	"code",
+	"monitor",
+	"iac",
+	"rules",
+	"describe",
+	"container",
+	"debug",
+	"config",
+	"auth",
+	"ignore",
+	"log4shell",
+	"policy",
+	"fix",
+	"apps",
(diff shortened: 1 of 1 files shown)


pkg/instrumentation/categories.go
```diff
@@ -79,7 +79,24 @@ func DetermineCategoryFromArgs(args []string, knownCommands []string, flagsAllow
...
+	"describe",
+	"container",
+	"debug",
+	"config",
+	"auth",
+	"ignore",
+	"log4shell",
+	"policy",
+	"fix",
+	"apps",
```
PeterSchafer, line 97: Question: any reason the commands below are missing?
* iac capture
* iac rules init
* iac rules push
* iac update-exclude-policy

mgyorke: The logic for building the category array parses the command line one argument at a time, it doesn't check "iac capture" as a single command. 

I'm assuming that this slice is used for that purpose only. If we need this, we also need to tweak the DetermineCategoryFromArgs method. 

Does this make sense?

PeterSchafer: I understand that the list of known_commands contains just the individual command names not set in relation. That makes sense. 
My question is more about why the names "capture" or "init" are not there but the "rules" or "describe". Everything in this list is supposed to be known commands, so why not add all subcommands?

mgyorke: Ah! ..you're right, nvm. I pushed the missing commands.

PeterSchafer: Perfect, thanks!

PR #182, opened 2024-05-22
fix: remove potential concurrent write

bastiandoetsch:


pkg/logging/scrubbingLogWriter.go
```diff
@@ -72,18 +83,16 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
...
 }
 
-func scrub(p []byte, scrubDict ScrubbingDict, regexCache map[string]*regexp.Regexp) []byte {
+func scrub(p []byte, regexCache map[string]*regexp.Regexp) []byte {
 	s := string(p)
-	for term := range scrubDict {
+	for term, re := range regexCache {
 		if len(term) > 0 {
-			re := regexCache[term]
 			if re == nil {
```
cmars, line 94: Can `re` still be nil here? Seems like it wouldn't be.

pkg/logging/scrubbingLogWriter.go
```diff
@@ -72,18 +83,16 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
...
+	_, err := w.writer.Write(scrub(p, w.regexCache))
 	return len(p), err // we return the original length, since we don't know the length of the redacted string
 }
 
-func scrub(p []byte, scrubDict ScrubbingDict, regexCache map[string]*regexp.Regexp) []byte {
+func scrub(p []byte, regexCache map[string]*regexp.Regexp) []byte {
 	s := string(p)
-	for term := range scrubDict {
+	for term, re := range regexCache {
 		if len(term) > 0 {
```
cmars, line 93: When would `term` be an empty string? Could we prevent those from getting into the cache in the first place?

bastiandoetsch: Totally.

PR #183, opened 2024-05-23
feat: add runtime support for adding scrub terms, add patterns [IDE-229]

bastiandoetsch:
🚨 Please don't squash, do a merge. Else the gitleaks check will trigger.

Adds support for 
- adding terms at runtime
- definining the matching group to be masked

Adds additional regex patterns.


PeterSchafer (approved):
Praise! 🥇

PR #187, opened 2024-05-24
feat: clear auth config keys when switching auth types

cmars:
Authenticating with token (deprecated) or oauth causes different config keys to be written to the config store. If these are unset, it potentially leaves stale auth information in the config. This also makes it difficult to determine which auth information is being used during CLI interactions.

This change forces the authentication config to be either oauth or token (deprecated). If one is set, the other method's keys are unset.

This required the addition of a Configuration.Unset method. "Unset" is a long-sought after feature of spf13/viper, which will probably never be added by the maintainer, but it is fairly straightforward to implement in coordination with the config file storage.

### DIFF
--- a/pkg/local_workflows/auth_workflow.go
+++ b/pkg/local_workflows/auth_workflow.go
@@ -98,6 +98,9 @@ func entryPointDI(config configuration.Configuration, logger *zerolog.Logger, en
 	logger.Printf("Authentication Type: %s", authType)
 
 	if strings.EqualFold(authType, authTypeOAuth) { // OAUTH flow
+		logger.Printf("Unset legacy token key %q from config", configuration.AUTHENTICATION_TOKEN)
+		config.Unset(configuration.AUTHENTICATION_TOKEN)
+
 		headless := config.GetBool(headlessFlag)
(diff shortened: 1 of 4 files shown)


thisislawatts (approved):

pkg/local_workflows/auth_workflow.go
```diff
@@ -108,9 +110,11 @@ func entryPointDI(config configuration.Configuration, logger *zerolog.Logger, en
 
 		fmt.Println(auth.AUTHENTICATED_MESSAGE)
 	} else { // LEGACY flow
+		config.Unset(auth.CONFIG_KEY_OAUTH_TOKEN)
```
thisislawatts, line 113: **suggestion (non-blocking):** Consider introducing some logging that details the why behind the configuration changes. This may help folks in the future when debugging "unexpected" configuration changes around this.

cmars: Good idea, did you have debug logging in mind?

cmars: Updated, PTAL

PR #190, opened 2024-05-27
feat: introduce error annotations for data objects [CLI-309]

thisislawatts:
* [x] Blocked by https://github.com/snyk/code-client-go/pull/50
* [ ] Verified integration tests on CLI repo: https://github.com/snyk/cli/pull/5261

### DIFF
--- a/internal/utils/sarif/sarif.go
+++ b/internal/utils/sarif/sarif.go
@@ -34,7 +34,6 @@ func CreateCodeSummary(input *sarif.SarifDocument) *json_schemas.TestSummary {
 	summary := json_schemas.NewTestSummary(summaryType)
 	resultMap := map[string]*json_schemas.TestSummaryResult{}
 
-	summary.Artifacts = 0
 	summary.SeverityOrderAsc = []string{"low", "medium", "high"}
 
 	for _, run := range input.Runs {
--- a/internal/utils/sarif/sarif.go
+++ b/internal/utils/sarif/sarif.go
@@ -9,6 +9,8 @@ const (
 	summaryType = "sast"
 )
 
+type CodeSummaryOptions func(s *json_schemas.TestSummary) *json_schemas.TestSummary
+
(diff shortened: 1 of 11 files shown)


pkg/local_workflows/code_workflow/native_workflow.go
```diff
@@ -55,12 +55,17 @@ func EntryPointNative(invocationCtx workflow.InvocationContext, opts ...Optional
 		return nil, err
 	}
 
-	summary := sarif2.CreateCodeSummary(&result.Sarif)
+	summary := sarif2.CreateCodeSummary(&result.Sarif, sarif2.WithCoverage(result.Coverage))
```
PeterSchafer, line 58: Question: Is `result.Coverage` the correct field? There is also `runs[].properties.coverage`.

internal/utils/sarif/sarif.go
```diff
@@ -25,14 +27,26 @@ func SarifLevelToSeverity(level string) string {
...
-func CreateCodeSummary(input *sarif.SarifDocument) *json_schemas.TestSummary {
+func CreateCodeSummary(input *sarif.SarifDocument, options ...CodeSummaryOptions) *json_schemas.TestSummary {
 	if input == nil {
 		return nil
 	}
 
 	summary := json_schemas.NewTestSummary(summaryType)
 	resultMap := map[string]*json_schemas.TestSummaryResult{}
+
+	summary.Artifacts = 0
```
PeterSchafer, line 49: Suggestion: How about moving the default value to `json_schemas.NewTestSummary()`?

pkg/local_workflows/code_workflow/native_workflow.go
thisislawatts: After reviewing the SarifDocument we found the data we need in there. Refactor this to use the SarifDocument > Runs property.

PeterSchafer (approved):

PR #193, opened 2024-05-28
fix: Support to use a configured repository URL when deriving a Target ID

PeterSchafer:


mgyorke (approved):

pkg/instrumentation/get-target-id_test.go
```diff
@@ -155,6 +157,18 @@ func Test_GetTargetId(t *testing.T) {
 
 		assert.Equal(t, actualAbsolute, actualRelative)
 	})
+
+	t.Run("configured repo url", func(t *testing.T) {
```
thisislawatts, line 161: **question:** Do we want to add a check here for behaviour with an empty or malformed git url?

PeterSchafer: Done!

PR #194, opened 2024-05-28
feat: introduce ErrorRender

thisislawatts:


internal/presenters/components.go
```diff
@@ -8,9 +8,58 @@ import (
...
+
+	if err.StatusCode > 0 {
+		httpLabel := lipgloss.NewStyle().Width(8).Render("HTTP:")
+		httpValue := lipgloss.NewStyle().PaddingLeft(1).PaddingRight(1).Width(80).Render(strconv.Itoa(err.StatusCode))
+		body = append(body, lipgloss.JoinHorizontal(lipgloss.Top, httpLabel, httpValue))
+	}
+
+	if len(err.Links) > 0 {
+		helpLabel := lipgloss.NewStyle().Width(8).Render("Help:")
+		helpValue := lipgloss.NewStyle().PaddingLeft(1).PaddingRight(1).Width(80).Render(strings.Join(err.Links, "\n"))
```
PeterSchafer, line 53: Issue: If the width is set, links might get wrapped and are broken then. See below. One solution could be to remove `Width(80)` completely or increase the value.

internal/presenters/components.go
```diff
@@ -8,9 +8,58 @@ import (
...
+
+func RenderError(err snyk_errors.Error) string {
+	level := strings.ToUpper(err.Level)
+
+	backgroundHighlight := errorLevelToStyle(err.Level)
+
+	var body []string
+
+	if len(err.Detail) > 0 {
+		infoLabel := lipgloss.NewStyle().Width(8).Render("Info:")
```
PeterSchafer, line 40: Suggestion: For all values like Width, padding ... that are used in multiple places it makes sense to introduce a constant as a single source of truth.

internal/presenters/components.go
thisislawatts: Flagged with design for review, ⌛ will revisit.

internal/presenters/components.go
thisislawatts: Removed wrapping for link block https://github.com/snyk/go-application-framework/pull/194/commits/e397089af9096bb5991d22c6b54fa88a620df858

internal/presenters/components.go
thisislawatts: Removed wrapping for link block https://github.com/snyk/go-application-framework/pull/194/commits/e397089af9096bb5991d22c6b54fa88a620df858

internal/presenters/components.go
thisislawatts: Addressed in https://github.com/snyk/go-application-framework/pull/194/commits/cae6cfa85c5539abdabf705c177f8057abddb79c

cmars (approved):
LGTM. I feel like there's possibly some opportunity to refactor this to reuse the labels and colors, but maybe that's something we do later when it becomes more necessary.

PR #198, opened 2024-05-31
fix(output_workflow): show org slug instead of org id in output

j-luong:


pkg/configuration/constants.go
```diff
@@ -13,6 +13,7 @@ const (
 	INTEGRATION_ENVIRONMENT_VERSION string = "snyk_integration_environment_version"
 	ANALYTICS_DISABLED              string = "snyk_disable_analytics"
 	ORGANIZATION                    string = "org"
+	ORGANIZATION_SLUG               string = "org_slug"
```
PeterSchafer, line 16: Suggestion: add `internal_` as we are not expecting this to be a explicit externally set value.

pkg/app/app.go
```diff
@@ -38,6 +38,22 @@ func defaultFunc_FF_CODE_CONSISTENT_IGNORES(engine workflow.Engine, config confi
 	return callback
 }
 
+func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Configuration, apiClient api.ApiClient, logger *zerolog.Logger) configuration.DefaultValueFunction {
+	callback := func(existingValue interface{}) interface{} {
+		client := engine.GetNetworkAccess().GetHttpClient()
+		url := config.GetString(configuration.API_URL)
+		apiClient.Init(url, client)
+		orgId := config.Get(configuration.ORGANIZATION)
```
PeterSchafer, line 46: Suggestion: better use `config.GetString()`

PeterSchafer: Also maybe add a case to handle an empty `orgId`

internal/api/contract/GetOrgResponse.go
```diff
@@ -0,0 +1,28 @@
+package contract
```
PeterSchafer, line 1: Question: Does it make sense to generate this contract rather than manually maintaining it?

PeterSchafer: We will postpone this topic a bit until we have a solution we can apply consistently.

j-luong: 💯, but this sounds like another initiative which shouldn't block this PR

PeterSchafer (approved):

PR #201, opened 2024-06-03
fix: empty errors in serialized instrumentation data

PeterSchafer:
* When errors don't have an id, don't add empty error structures to the serialized data
* ensure category vector is lower case
* add GetDuration()

### DIFF
--- a/pkg/analytics/instrumentation_collector.go
+++ b/pkg/analytics/instrumentation_collector.go
@@ -25,6 +25,7 @@ type InstrumentationCollector interface {
 	SetInteractionId(id string)
 	SetTimestamp(t time.Time)
 	SetDuration(duration time.Duration)
+	GetDuration() time.Duration
 	SetStage(s string)
 	SetType(t string)
 	SetInteractionType(t string)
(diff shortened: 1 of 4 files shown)


pkg/instrumentation/categories.go
```diff
@@ -62,14 +62,14 @@ func DetermineCategoryFromArgs(args []string, knownCommands []string, flagsAllow
 			flagName := strings.TrimLeft(flagParts[0], "-")
 
 			if slices.Contains(flagsAllowList, flagName) {
-				flags = append(flags, flagName)
+				flags = append(flags, strings.ToLower(flagName))
```
cmars, line 65: Could we add a test case for this?

pkg/analytics/instrumentation_collector.go
```diff
@@ -228,13 +232,14 @@ func toInteractionErrors(errors []error) *[]api.InteractionError {
 
 func toInteractionError(e error) *api.InteractionError {
```
cmars, line 233: Could we add a test case for this function?

cmars (commented):
LGTM with a couple of test cases added

pkg/analytics/instrumentation_collector.go
PeterSchafer: Done!

pkg/instrumentation/categories.go
PeterSchafer: Done!

PR #203, opened 2024-06-03
fix: line break in links

PeterSchafer:
The issue was mentioned [here](https://github.com/snyk/go-application-framework/pull/194#discussion_r1619071694) before but there seems to have been a regression.
This PR fixes the problem by copying the style before changing it and adding a test to detect a regression.


cmars (approved):
Tricky!

PR #204, opened 2024-06-04
fix: Sending behaviour in analytics report workflow

PeterSchafer:
This fix changes the behaviour of the analytics report workflow to convert v1 data and send it directly. This change is necessary to work in VS and IntelliJ, since a previous assumption didn't work out.

### DIFF
--- a/pkg/local_workflows/report_analytics_workflow.go
+++ b/pkg/local_workflows/report_analytics_workflow.go
@@ -168,14 +168,6 @@ func instrumentScanDoneEvent(invocationCtx workflow.InvocationContext, input wor
 	ic := analytics.NewInstrumentationCollector()
 	ic.SetInteractionId(instrumentation.AssembleUrnFromUUID(uuid.NewString()))
 
-	if config.IsSet(configuration.INPUT_DIRECTORY) {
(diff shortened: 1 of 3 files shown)


pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -200,7 +220,20 @@ func instrumentScanDoneEvent(invocationCtx workflow.InvocationContext, input wor
 	ic.SetTestSummary(toTestSummary(scanDoneEvent.Data.Attributes.UniqueIssueCount, scanDoneEvent.Data.Type))
 	ic.AddExtension("device_id", scanDoneEvent.Data.Attributes.DeviceId)
 
-	return nil
+	data, err := analytics.GetV2InstrumentationObject(ic)
+	if err != nil {
+		return nil, err
+	}
+
+	v2InstrumentationData := utils.ValueOf(json.Marshal(data))
```
cmars, line 228: Please do not discard errors. If propagating this error isn't the right thing, let's debug-log it.

```suggestion
	v2InstrumentationData, err := json.Marshal(data)
	if err != nil {
		return nil, fmt.Errorf("failed to marshal instrumentation: %w", err)
	}
```

PeterSchafer: copy pasted, the error is now handled.

j-luong (approved):

PR #206, opened 2024-06-07
fix: use cli runtime info when translating v1 data

PeterSchafer:
IDEs send app name and version as the IDE name, which makes it redundant, as the same info is in the environment. This PR ensures to keep CLI version information.

### DIFF
--- a/pkg/local_workflows/report_analytics_workflow.go
+++ b/pkg/local_workflows/report_analytics_workflow.go
@@ -190,8 +190,8 @@ func instrumentScanDoneEvent(invocationCtx workflow.InvocationContext, input wor
 
 	// required v2 analytics parameters
 	userAgent := networking.UserAgentInfo{
-		App:                           scanDoneEvent.Data.Attributes.Application,
-		AppVersion:                    scanDoneEvent.Data.Attributes.ApplicationVersion,
+		App:                           invocationCtx.GetRuntimeInfo().GetName(),
+		AppVersion:                    invocationCtx.GetRuntimeInfo().GetVersion(),
 		Integration:                   scanDoneEvent.Data.Attributes.IntegrationName,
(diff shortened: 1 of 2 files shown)


pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -190,8 +190,8 @@ func instrumentScanDoneEvent(invocationCtx workflow.InvocationContext, input wor
 
 	// required v2 analytics parameters
 	userAgent := networking.UserAgentInfo{
-		App:                           scanDoneEvent.Data.Attributes.Application,
-		AppVersion:                    scanDoneEvent.Data.Attributes.ApplicationVersion,
+		App:                           invocationCtx.GetRuntimeInfo().GetName(),
+		AppVersion:                    invocationCtx.GetRuntimeInfo().GetVersion(),
```
j-luong, line 194: question: is it better for the workflow to be ignorant here and keep the data received from the workflow as the source of truth? this looks like it makes assumptions on implementation details of the IDE

PeterSchafer: The problem is that already released versions of the IDEs got it wrong. They set both Environment Name/Version (IDE) and App Name/Version (should be snyk-cli or snyk-ls) to the IDE Name/Version. The problem is that we loose visibility in the application version in use. If we would correct this in new IDE plugin releases, this would only fix the issue for new versions and it would require the IDEs to know the snyk-cli/snyk-ls version.

PeterSchafer: We could of course add a check and only override of the values are the same.

j-luong: or add `invocationCtx.GetRuntimeInfo().GetName()` and `invocationCtx.GetRuntimeInfo().GetVersion()` as part of the extension metadata?

j-luong (approved):

PR #208, opened 2024-06-14
feat: update code-client-go to v1.8.0 [IDE-399]

teodora-sandu:
Includes the following changes:
- https://github.com/snyk/code-client-go/releases/tag/v1.7.2 - refactored the repository URL computation so it doesn't use the hack for SSH repos
- https://github.com/snyk/code-client-go/releases/tag/v1.7.3 - fixes a bug for `snyk-ls` only, where incremental file scanning was broken
- https://github.com/snyk/code-client-go/releases/tag/v1.8.0 - where `SnykCodeAnalysisTimeout` becomes necessary

With this PR we don't limit Snyk Code scans to an arbitrary 2 minutes, which is too low for larger repos. 
I set the default timeout to 12h if it's not configured (as per https://github.com/snyk/code-client-go/pull/56#discussion_r1639395178) but otherwise using the now unused `snyk_timeout_secs` configuration to override.


pkg/local_workflows/code_workflow/code_client_helper.go
```diff
@@ -25,3 +28,18 @@ func (c *codeClientConfig) SnykCodeApi() string {
 func (c *codeClientConfig) SnykApi() string {
 	return c.localConfiguration.GetString(configuration.API_URL)
 }
+
+func (c *codeClientConfig) SnykCodeAnalysisTimeout() time.Duration {
```
PeterSchafer, line 32: Question: the newly added function `SnykCodeAnalysisTimeout()` is not used somewhere, is this expected?

teodora-sandu: What do you mean it's not used? This is where it's used in `code-client-go` https://github.com/snyk/code-client-go/blob/main/internal/analysis/analysis.go#L349

pkg/local_workflows/code_workflow/code_client_helper.go
```diff
@@ -25,3 +28,18 @@ func (c *codeClientConfig) SnykCodeApi() string {
...
+	var snykCodeTimeout time.Duration
+	var err error
+	env := c.localConfiguration.GetString(configuration.TIMEOUT)
```
PeterSchafer, line 35: Suggestion: User `GetInt()` instead of `getString()` and maybe even `isSet()` to keep the code simpler.

pkg/local_workflows/code_workflow/code_client_helper.go
PeterSchafer: Should it be used in the native workflow?

teodora-sandu: `SnykCodeAnalysisTimeout()` should be used for the new code workflow in GAF that uses ignores, which from what I can tell from the code is the native workflow. Did I misunderstand anything?

pkg/local_workflows/code_workflow/code_client_helper.go
teodora-sandu: Great suggestion, all done in https://github.com/snyk/go-application-framework/pull/208/commits/3ea752e21e26a465736cd32b55480c283a299e5f

pkg/local_workflows/code_workflow/code_client_helper.go
PeterSchafer: Sorry I missed that this change is in the injected configuration object 🤦 now the world makes sense again.

PeterSchafer (approved):

PR #209, opened 2024-06-18
fix: log error when calling endpoint [IDE-418]

acke:
LS should not return an error to the IDEs if the analytics service isn't available or the call returns an error. Instead, it should submit a log.

### DIFF
--- a/pkg/local_workflows/report_analytics_workflow.go
+++ b/pkg/local_workflows/report_analytics_workflow.go
@@ -132,8 +132,7 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
 		// send to V2 analytics endpoint
 		callErr := callEndpoint(invocationCtx, input, url)
 		if callErr != nil {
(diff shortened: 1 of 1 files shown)


pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -132,8 +132,7 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
 		// send to V2 analytics endpoint
 		callErr := callEndpoint(invocationCtx, input, url)
 		if callErr != nil {
-			err := fmt.Errorf("error calling endpoint for input at index %d: %w", i, callErr)
-			return nil, err
+			logger.Printf("error calling endpoint for input at index %d: %v", i, callErr)
```
PeterSchafer, line 135: Suggestion: The decision what to do in case of this error should be on the application not on the workflow, I would recommend to do the change in snyk-ls and leave this code unchanged.

acke:
Will close pr, and it must not be merged.

PR #213, opened 2024-06-26
feat: wrap authentication error for matching

cmars:
Wrap authentication failure errors with ErrAuthenticationFailed so that callers of NetworkAccess.AddHeaders can match this particular error condition: whether headers failed to add due to an authentication failure.

CLI-392

### DIFF
--- a/pkg/networking/middleware/auth_header.go
+++ b/pkg/networking/middleware/auth_header.go
@@ -1,6 +1,8 @@
 package middleware
 
 import (
+	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


sandor-trombitas (approved):

pkg/networking/middleware/auth_header.go
```diff
@@ -82,6 +84,15 @@ func ShouldRequireAuthentication(
 	return false, nil
 }
 
+// ErrAuthenticationFailed indicates that authentication failed in the
+// networking middleware.
+var ErrAuthenticationFailed = fmt.Errorf("authentication failed")
```
PeterSchafer, line 89: Suggestion: We should use an error catalog error for this.

cmars: Done!

cmars: Well, it was done, but I had to reconsider this. Error catalog sent me down an interesting rabbit hole.

Error catalog as it exists today is mainly useful for errors "on the way out" to the user, and we should absolutely use them for this purpose.

They're not really great for wrapping and winding through the stack because you can't match them very consistently in their current form. Which is what this error is all about.

I gave them a shot, managed to make them fit, then decided it was a net degradation in code quality and clarity, so I reverted that attempt, though I'll keep the branch around for reference!

Tangential to all this, what I'd like to see improve in the error catalog package over the longer term:
- There needs to be an easy way to match errors, especially "code". Ideally using errors.Is and As.
- All the codes should be consts that can be tested for.

Overall it was an eye-opening and productive rabbit hole :)

PR #216, opened 2024-07-24
feat: add configuration sanity check to configuration environment workflow

PeterSchafer:
* to inform users that the configuration might be inconsistent

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -460,7 +460,7 @@ func (ev *extendedViper) createFileStorage(configPath string) Storage {
 }
 
 // GetAllKeysThatContainValues() returns a list of all keys, including alternative keys, that are set for the given key.
-// This can be used to identify in which way a certain has been set. If the result size is greater 1, this means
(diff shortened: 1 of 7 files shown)


pkg/configuration/configuration.go
```diff
@@ -450,3 +458,35 @@ func (ev *extendedViper) createFileStorage(configPath string) Storage {
 	file := path.Join(configPath, "snyk.json")
 	return NewJsonStorage(file, WithConfiguration(ev))
 }
+
+// GetAllKeysThatContainValues() returns a list of all keys, including alternative keys, that are set for the given key.
+// This can be used to identify in which way a certain has been set. If the result size is greater 1, this means
```
j-luong, line 463: typo: `a certain has been set` = `a certain key has been set`?

j-luong (approved):

pkg/configuration/configuration.go
sandor-trombitas, line 463:
```suggestion
// This can be used to identify in which way a certain key has been set. If the result size is greater 1, this means
```

sandor-trombitas (approved):

PR #219, opened 2024-07-30
fix: use RWMutex to allow sync writes/reads on dict for scrub writer

bastiandoetsch:
Before this PR, iterating over the scrub word dict was not protected from writes to the dict via addTerm() or removeTerm(). I've added a RWMutex to allow unlimited iterating over the dict, but locking everything once a write is occurring. This should fix concurrency issues like in https://github.com/snyk/snyk-ls/actions/runs/10162805720/job/28104241006?pr=608

### DIFF
--- a/pkg/logging/scrubbingLogWriter.go
+++ b/pkg/logging/scrubbingLogWriter.go
@@ -44,13 +44,13 @@ type scrubStruct struct {
 type ScrubbingDict map[string]scrubStruct
 
 type scrubbingLevelWriter struct {
-	m         sync.Mutex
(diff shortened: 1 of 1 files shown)


cmars (approved):
LGTM but a few questions about this:

Is there a need to dynamically add and remove scrubbed terms to/from a scrubbing log writer after it is created?

Or could these be defined up-front when the log writer is created? That could remove the need for a mutex since all access to the map would then be read-only once created.

Was this discovered with `go test -race`? Should we add `-race` to `make test` as a general rule? Go's race detector is pretty awesome.

PR #221, opened 2024-08-02
chore: set default network logging to debug level

PeterSchafer:


pkg/networking/networking.go
```diff
@@ -53,33 +53,35 @@ type networkImpl struct {
 	logger       *zerolog.Logger
 }
 
+const defaultNetworkLogLevel = zerolog.DebugLevel
+
 func LogRequest(r *http.Request, logger *zerolog.Logger) {
-	if logger.GetLevel() != zerolog.TraceLevel { // Only log if trace level is enabled
+	if logger.GetLevel() > defaultNetworkLogLevel { // Only log if trace level is enabled
```
cmars, line 59: I'm not sure this will evaluate true for trace level, given the constants defined in https://pkg.go.dev/github.com/rs/zerolog#Level

It looks like trace is less than debug.

cmars (commented):
Looks good but I think there might be an issue with the log level comparison.

pkg/networking/networking.go
bastiandoetsch: Yeah, looks like Debug = 0, Trace = -1. How about `if logger.GetLevel() <= defaultNetworkLevel`? That way network would be logged with debug & trace

PeterSchafer: It's Friday so maybe I'm wrong 😅  

The condition here is to **not** log anything. 

This means that:
* if the logger level is greater than the constant (debug) it will return and therefore not log (InfoLevel=1, WarnLevel=2, ErrorLevel=3, ...)
* if the logger level is less than equal the constant (debug) it'll continue and log. (debug, trace)

PeterSchafer: @cmars I think the incorrect comment caused some confusion. Please take a look again and let me know if you still think the code itself is incorrect.

cmars (approved):

PR #223, opened 2024-08-06
fix: interprocess file locking for oauth token refresh

cmars:
Addresses an interprocess race condition when multiple concurrent snyk CLI commands are running with an oauth token which needs to be refreshed.

TODOs:
- [x] Test & document configuration package changes
- [x] Lint TODOs

Updated Circle CI to use Docker instead of an ubuntu VM. Nothing in the GAF needs a VM to test and a Docker image pins down the Go toolchain version.

CLI-462


pkg/auth/oauth2authenticator.go
```diff
@@ -450,3 +461,46 @@ func (o *oAuth2Authenticator) AddAuthenticationHeader(request *http.Request) err
...
+	if storage := o.config.GetStorage(); storage != nil {
+		// Lock configuration storage and refresh, to avoid oAuth2Authenticator
+		// racing with itself on concurrently rotating client & refresh tokens.
+		err := storage.Lock(ctx)
+		if err != nil {
+			return cleanup, err
+		}
+		cleanup = func() {
+			if unlockErr := storage.Unlock(); unlockErr != nil {
+				fmt.Println("failed to unlock config storage:", unlockErr)
```
ShawkyZ, line 483: Shouldn't we use a logger here instead? I think fmt.println can affect LS stdin/out stream communication

cmars: How do we get a logger into the authenticator?

cmars: A logger would be ideal, but I'm not seeing a lot of logger integration across the GAF, and it'd be quite tricky to wind a logger through all the different ways this authenticator is instantiated.

For the time being, I'll write it out to stderr instead.

pkg/configuration/storage.go
```diff
@@ -107,3 +127,28 @@ func (s *JsonStorage) Set(key string, value any) error {
...
+	if err != nil {
+		return err
+	}
+	doc := map[string]interface{}{}
+	err = json.Unmarshal(contents, &doc)
+	if err != nil {
+		return err
+	}
+	for k, v := range doc {
+		config.Set(k, v)
```
PeterSchafer, line 142: Possible Issue: Using Set() is an override, and overrides have precedence. This means that setting values like this might cause a change in behaviour. Can we make use of the reloading mechanism inside viper, for example by calling [ReadInConfig()](https://pkg.go.dev/github.com/spf13/viper#ReadInConfig) instead?

cmars: In trying this approach, I think I've found a pre-existing bug in GAF configuration. `ev.viper.ReadInConfig()` fails to update the extendedViper keys in my unit tests. Digging into it, here's what I found:

The JsonStorage created by extendedViper always writes `~/.config/configstore/snyk.json` -- this filename is hard-coded, regardless of what the config name is set to, see https://github.com/snyk/go-application-framework/blob/068ee27b9941a958cce67e11d7eb332302678aaa/pkg/configuration/configuration.go#L465

However ReadInConfig attempts to honor the viper config name ("test" in unit tests), fails to find the file, and fails to refresh as a result. This might not impact production code if we only use `snyk.json` there, but it makes me aware that our storage implementation isn't always consistent with the way we're orchestrating viper.

I'm concerned this is becoming a rabbit hole and something to fix in a followup. For the time being, points to consider:

1. `Set` is used by `Clone`, so the override might be a reasonable thing to use in `Refresh`, given that we're setting from raw JSON parsed values.

2. Regarding keys not in the config, given that we persist only some of the keys, leaving the un-persisted config keys intact (as the current implementation does) seems less disruptive.

3. Making Storage entirely responsible for the serialization in refresh will likely have better results for a LS alternative implementation.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/snyk/go-application-framework
 
-go 1.21
+go 1.21.0
```
bastiandoetsch, line 3: can we potentially increase that to 1.22.x? I think LS already states that it uses 1.22.x, so bumping the go versions consistently would be nice (also for CLI).

cmars: Oh cool! I wasn't sure about that but heck yeah I could use range over integer in the test.

cmars: Done!

.golangci.yaml
```diff
@@ -32,7 +32,6 @@ linters-settings:
   gosimple:
     checks: ['all']
   govet:
-    check-shadowing: true
```
cmars, line 35: This is now default in the newer golangci-lint version, it was saying:

```
WARN [config_reader] The configuration option `linters.govet.check-shadowing` is deprecated. Please enable `shadow` instead, if you are not using `enable-all`. 
```

Makefile
```diff
@@ -3,7 +3,7 @@ GOOS = $(shell go env GOOS)
 GOARCH = $(shell go env GOARCH)
 
 GO_BIN := $(shell pwd)/.bin
-OVERRIDE_GOCI_LINT_V := v1.55.2
+OVERRIDE_GOCI_LINT_V := v1.59.1
```
cmars, line 6: Needed this because I'm using a 1.22-specific feature that the older lint didn't understand (range over int)

.circleci/config.yml
```diff
@@ -3,22 +3,22 @@ version: 2.1
 orbs:
   prodsec: snyk/prodsec-orb@1
 
-executors:
-  linux:
-    machine:
-      image: ubuntu-2204:2023.10.1
```
cmars, line 9: Realized VM is overkill here and it also didn't have the right Go version installed.

internal/utils/cache-dir_test.go
```diff
@@ -28,7 +28,7 @@ func (m *mockCacheDirOSUtil) MkdirAll(path string, perm os.FileMode) error {
 }
 
 func (m *mockCacheDirOSUtil) Stat(_ string) (os.FileInfo, error) {
-	return nil, nil
+	return nil, nil //nolint:nilnil // its a mock
```
cmars, line 31: New lint version complaining about this.

pkg/mocks/config_storage.go
cmars, line 1: This one and the following file are generated by mockgen, which needed a refresh due to interface changes.

Looks like we were a little out of sync.

pkg/configuration/storage.go
PeterSchafer: Trying to separate my concern above from a solution proposal.

**Concern**
Using Set() for all values from the config file changes the precedence. Taking the example where `org` is set in the config file but also as an Environment Variable. The normal behaviour is that the environment variable will be consumed. After udpating the token, the org value from the config file will be Set() and suddenly another org will be used.

I think this concern is generally valid here. The difference to the mentioned usage in `Clone()` is that there it calls `Get()` and uses this value to `Set()`, it basically takes a snapshot of the configuration in time. But the snapshot has the same values. And doesn't change automatically through the update of an unassociated value.

**Solution**
If my previous hint towards a solution doesn't work, let's take a look further.

pkg/configuration/configuration_test.go
```diff
@@ -439,3 +441,95 @@ func Test_Configuration_GetKeyType(t *testing.T) {
...
+			err := storage.Lock(ctx, time.Millisecond)
+			assert.NoError(t, err)
+			defer func() {
+				unlockErr := storage.Unlock()
+				assert.NoError(t, unlockErr)
+			}()
+
+			err = storage.Refresh(config, "n")
+			assert.NoError(t, err)
+			config.Set("n", config.Get("n").(float64)+1)
```
PeterSchafer, line 470: nitpick: There is GetFloat64() ;)

pkg/configuration/configuration_test.go
```diff
@@ -439,3 +441,95 @@ func Test_Configuration_GetKeyType(t *testing.T) {
...
+				assert.NoError(t, err)
+			}
+			defer func() {
+				unlockErr := storage.Unlock()
+				assert.NoError(t, unlockErr)
+			}()
+
+			err = storage.Refresh(config, "n")
+			assert.NoError(t, err)
+			n := config.Get("n").(float64) //nolint:errcheck // let it panic in a test
```
PeterSchafer, line 515: nitpick: Same as above, there is GetFloat64() which probably avoids the linter topic.

PR #226, opened 2024-08-15
feat: Add support for dynamic headers

sandor-trombitas:
* This PR adds a possibility to add http headers dynamically in addition to the existing static headers.
* It also introduces a helper to add a dynamically generated snyk-request-id

### DIFF
--- a/pkg/local_workflows/network_utils/snyk_request_id.go
+++ b/pkg/local_workflows/network_utils/snyk_request_id.go
@@ -1,0 +1,16 @@
+package network_utils
+
+import (
+	"github.com/google/uuid"
+
+	"github.com/snyk/go-application-framework/pkg/networking"
+)
+
+func AddSnykRequestId(n networking.NetworkAccess) {
+	n.AddDynamicHeaderField("snyk-request-id", func(values []string) []string {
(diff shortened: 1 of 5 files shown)


pkg/local_workflows/network_utils/snyk_request_id.go
```diff
@@ -0,0 +1,16 @@
+package network_utils
+
+import (
+	"github.com/google/uuid"
+
+	"github.com/snyk/go-application-framework/pkg/networking"
+)
+
+func AddSnykRequestId(n networking.NetworkAccess) {
```
cmars, line 9: A godoc comment would be nice 🙏. There's a recommended convention for these, see https://go.dev/doc/comment

(At some point, we should turn on the linter rule requiring them for exported symbols.)

pkg/networking/networking.go
```diff
@@ -44,13 +46,16 @@ type NetworkAccess interface {
 	Clone() NetworkAccess
 }
 
+type DynamicHeaderFunc func([]string) []string
```
cmars, line 49: Godoc would be nice here as well.

cmars (approved):
Awesome stuff!

PR #228, opened 2024-08-22
fix: panic when no remote url exists

ShawkyZ:


pkg/instrumentation/get-target-id.go
```diff
@@ -199,10 +199,11 @@ func gitBaseId(path string) (*url.URL, error) {
 	}
 
 	// based on the docs, the first URL is being used to fetch, so this is the one we use
-	repoUrl := remote.Config().URLs[0]
-	if repoUrl == "" {
+	remoteConfig := remote.Config()
```
PeterSchafer, line 202: Suggestion: looks good to me! Do we have a chance to add a test case?

ShawkyZ: looking into it

PeterSchafer: Currently it is unclear to create such a test case, so we will postpone this.

PeterSchafer (approved):

pkg/instrumentation/get-target-id.go
```diff
@@ -199,10 +199,11 @@ func gitBaseId(path string) (*url.URL, error) {
...
+	if remoteConfig == nil || len(remoteConfig.URLs) == 0 || remoteConfig.URLs[0] == "" {
```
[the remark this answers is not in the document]
bastiandoetsch: Would you please file an issue with `go-git` then? They may fix it if they deem it appropriate.

PR #233, opened 2024-09-05
feat: add snyk_tmp_dir_path config

sandor-trombitas:


pkg/app/app.go
```diff
@@ -125,6 +126,29 @@ func defaultInputDirectory() configuration.DefaultValueFunction {
...
+		if existingValue != nil {
+			return existingValue
+		}
+		ri := engine.GetRuntimeInfo()
+		if ri == nil {
+			return false
+		}
+
+		version := ri.GetVersion()
+		tmpDir := pkg_utils.GetTemporaryDirectory(os.TempDir(), version)
```
PeterSchafer, line 140: Issue: The first parameter should be the configured `cache_path`. 

something like this.
```suggestion
		tmpDir := pkg_utils.GetTemporaryDirectory(config.getString(configuration.CACHE_PATH), version)
```

pkg/app/app.go
```diff
@@ -125,6 +126,29 @@ func defaultInputDirectory() configuration.DefaultValueFunction {
...
+	callback := func(existingValue interface{}) interface{} {
+		if existingValue != nil {
+			return existingValue
+		}
+		ri := engine.GetRuntimeInfo()
+		if ri == nil {
+			return false
+		}
+
+		version := ri.GetVersion()
```
PeterSchafer, line 139: Suggestion: let's just check that the version isn't empty and use some default value if it is.

pkg/configuration/constants.go
```diff
@@ -18,6 +18,7 @@ const (
 	DEBUG_FORMAT                    string = "debug_format"
 	INSECURE_HTTPS                  string = "insecure"
 	PROXY_AUTHENTICATION_MECHANISM  string = "proxy_auth"
+	TEMP_DIR_PATH                   string = "snyk_tmp_dir_path"
```
PeterSchafer, line 21: Suggestion: `snyk_` will translate into an environment variable as per naming convention. considering this, we might want to look at consistency with other examples. `SNYK_CACHE_PATH` for example, this would mean `SNYK_TMP_PATH` makes more sense.

pkg/app/app_test.go
```diff
@@ -281,3 +282,33 @@ func Test_initConfiguration_PREVIEW_FEATURES_ENABLED(t *testing.T) {
 	actual = config.GetBool(configuration.PREVIEW_FEATURES_ENABLED)
 	assert.True(t, actual)
 }
+
+func Test_initConfiguration_DEFAULT_TEMP_DIRECTORY(t *testing.T) {
```
PeterSchafer, line 286: Suggestion: add a case where the tmp path is specified in the config, this should return the configured rather then the default value.

pkg/app/app.go
```diff
@@ -125,6 +126,33 @@ func defaultInputDirectory() configuration.DefaultValueFunction {
 	return callback
 }
 
+func defaultTempDirectory(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger) configuration.DefaultValueFunction {
+	callback := func(existingValue interface{}) interface{} {
+		if existingValue != nil {
+			return existingValue
+		}
+		ri := engine.GetRuntimeInfo()
```
PeterSchafer, line 134: Suggestion: Let's combine the case where version is empty and runtimeinfo is nil.

pkg/app/app.go
```diff
@@ -125,6 +126,33 @@ func defaultInputDirectory() configuration.DefaultValueFunction {
...
+		version := ri.GetVersion()
+		if len(version) == 0 {
+			version = "0.0.0"
+		}
+
+		tmpDir := pkg_utils.GetTemporaryDirectory(config.GetString(configuration.CACHE_PATH), version)
+		err := pkg_utils.CreateAllDirectories(tmpDir, version)
+		if err != nil {
+			logger.Err(err)
+			return false
```
PeterSchafer, line 148: Suggestion: Please always return the same type, in this case string. Also it might be worth to think of a better more error tolerant fallback if the folders couldn't be created. Maybe we just return the directory anyways and only keep the log?

PeterSchafer (approved):

PR #234, opened 2024-09-09
feat: conditionally set the workflow data location based on size

j-luong:


pkg/configuration/constants.go
```diff
@@ -18,6 +18,7 @@ const (
 	DEBUG_FORMAT                    string = "debug_format"
 	INSECURE_HTTPS                  string = "insecure"
 	PROXY_AUTHENTICATION_MECHANISM  string = "proxy_auth"
+	IN_MEMORY_THRESHOLD_BYTES       string = "in_memory_threshold_bytes"
```
PeterSchafer, line 21: Suggestion: make this `internal_`

pkg/app/app.go
```diff
@@ -194,6 +195,8 @@ func initConfiguration(engine workflow.Engine, config configuration.Configuratio
 	config.AddDefaultValue(configuration.CACHE_PATH, configuration.StandardDefaultValueFunction(dir))
 	config.AddDefaultValue(configuration.AUTHENTICATION_SUBDOMAINS, configuration.StandardDefaultValueFunction([]string{"deeproxy"}))
 	config.AddDefaultValue(configuration.MAX_THREADS, configuration.StandardDefaultValueFunction(runtime.NumCPU()))
+	// set default filesize threshold to 512MB
+	config.AddDefaultValue(configuration.IN_MEMORY_THRESHOLD_BYTES, configuration.StandardDefaultValueFunction(fmt.Sprintf("%d", 512*1024*1024)))
```
PeterSchafer, line 199: Question: out of curiosity, why do we need `fmt.Sprintf()`? can't we us an integer directly?

pkg/workflow/dataimpl.go
```diff
@@ -2,30 +2,157 @@ package workflow
...
+
+// Deprecated: Use NewData instead
+//
 // NewDataFromInput creates a new data instance from the given input data.
 //
 // It will preserve the headers, metadata and errors
-func NewDataFromInput(input Data, typeIdentifier Identifier, contentType string, payload interface{}) Data {
+func NewDataFromInput(input Data, typeIdentifier Identifier, contentType string, payload interface{}, opts ...Option) Data {
+	// initialise DataImpl
+	output := &DataImpl{
```
PeterSchafer, line 143: Suggestion: don't duplicate this code, just use `newDataWith()`

pkg/workflow/dataimpl.go
```diff
@@ -11,62 +14,158 @@ import (
 
 // DataImpl is the default implementation of the Data interface.
 type DataImpl struct {
-	identifier Identifier
-	header     http.Header
-	payload    interface{}
-	errors     []snyk_errors.Error
+	identifier      Identifier
+	header          http.Header
+	contentType     string
```
PeterSchafer, line 19: Question: content-type is already stored in the header, why do we need an extra variable for it?

pkg/workflow/dataimpl.go
```diff
@@ -11,62 +14,158 @@ import (
...
+	inputData       Data
```
PeterSchafer, line 20: Question: Do we really need to hold the input data?

pkg/workflow/dataimpl.go
```diff
@@ -11,62 +14,158 @@ import (
...
+
+func withPayload(payload *interface{}) Option {
+	return func(d *DataImpl) {
+		d.payload = *payload
+	}
+}
+
+func WithConfiguration(config *configuration.Configuration) Option {
+	return func(d *DataImpl) {
+		d.configuration = *config
```
PeterSchafer, line 69: Suggestion: For all `WithXyz(xyz)` you don't have to hold on to everything that is given to the with function, you can also derive the values in with and not hold onto the configuration, the data ... This way you enable to release memory earlier and don't make a data object container for everything. In some reasons it might be useful but not as a default.

pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -81,6 +81,8 @@ func reportAnalyticsEntrypoint(invocationCtx workflow.InvocationContext, inputDa
 			workflow.NewTypeIdentifier(WORKFLOWID_REPORT_ANALYTICS, reportAnalyticsWorkflowName),
 			"application/json",
 			[]byte(commandLineInput),
+			workflow.WithConfiguration(&config),
```
PeterSchafer, line 84: Suggestion: Let's limit impact of the changes and not use this function everywhere.

pkg/local_workflows/output_workflow.go
```diff
@@ -79,7 +79,14 @@ func filterSummaryOutput(config configuration.Configuration, input workflow.Data
 	}
 
 	workflowId := workflow.NewTypeIdentifier(WORKFLOWID_OUTPUT_WORKFLOW, "FilterTestSummary")
-	return workflow.NewDataFromInput(input, workflowId, content_type.TEST_SUMMARY, bytes), nil
+	output := workflow.NewDataFromInput(
```
PeterSchafer, line 82: Suggestion: Let's limit impact of the changes and not use this function everywhere.

pkg/workflow/dataimpl.go
```diff
@@ -134,3 +251,56 @@ func (d *DataImpl) GetErrorList() []snyk_errors.Error {
...
+		return payloadLocation
+	}
+
+	logger.Debug().Msg("payload is larger than threshold, writing it to disk")
+	return writePayloadToDisk(id, config, bytes, logger)
+}
+
+func writePayloadToDisk(id Identifier, config configuration.Configuration, data []byte, logger *zerolog.Logger) Location {
+	tempDir := config.GetString(configuration.TEMP_DIR_PATH)
+	filepath := fmt.Sprintf("%s/workflow.%s.%d", tempDir, id.Path, time.Now().Unix())
```
PeterSchafer, line 290: Suggestion: Better use [stdlib functionality](https://pkg.go.dev/os#CreateTemp) this is safer than using a unix timestamp.

pkg/workflow/dataimpl.go
```diff
@@ -134,3 +251,56 @@ func (d *DataImpl) GetErrorList() []snyk_errors.Error {
...
+	bytes, ok := payload.([]byte)
+	if !ok {
+		return payloadLocation
+	}
+
+	payloadSize := len(bytes)
+	fileSizeThreshold := config.GetInt(configuration.IN_MEMORY_THRESHOLD_BYTES)
+	logger.Debug().Msgf("payload is []byte, comparing payload size (%d bytes) to threshold (%d bytes)", payloadSize, fileSizeThreshold)
+
+	if payloadSize <= fileSizeThreshold {
```
PeterSchafer, line 279: Question: how can I disable it through configuration. Setting a large number? how about setting the threshold to -1.

pkg/workflow/types.go
```diff
@@ -35,6 +35,8 @@ type Data interface {
 	SetContentLocation(string)
 	AddError(err snyk_errors.Error)
 	GetErrorList() []snyk_errors.Error
+	SetConfiguration(config configuration.Configuration)
```
PeterSchafer, line 38: Question: Do we really need this here? otherwise I would say configuration is to big of a concept to link it to on data instance. better take what you need from the configuration once.

pkg/workflow/dataimpl.go
```diff
@@ -11,62 +14,158 @@ import (
...
-		dataIdentifier.Fragment = fmt.Sprintf("%d", time.Now().Nanosecond())
+		output.identifier.Fragment = fmt.Sprintf("%d", time.Now().Nanosecond())
 	}
 
-	output := &DataImpl{
-		identifier: &dataIdentifier,
-		header:     header,
-		payload:    payload,
-		errors:     errors,
+	output.payloadLocation = setPayloadLocation(output.identifier, output.configuration, output.payload, output.logger)
```
PeterSchafer, line 129: Suggestion: keep logic close together, separating writing to disk and related logic makes it difficult to understand and error prone. Currently there are two places that use `setPayloadLocation()` which can easily cause the errors. Can't we just call a common function? that does the same in both cases?

pkg/workflow/dataimpl.go
```diff
@@ -90,12 +189,30 @@ func (d *DataImpl) GetMetaData(key string) (string, error) {
 
 // SetPayload sets the payload of the given data instance.
 func (d *DataImpl) SetPayload(payload interface{}) {
-	d.payload = payload
+	payloadLocation := setPayloadLocation(d.identifier, d.configuration, payload, d.logger)
```
PeterSchafer, line 192: Suggestion: see https://github.com/snyk/go-application-framework/pull/234/files#r1755160730

pkg/workflow/dataimpl.go
```diff
@@ -134,3 +251,56 @@ func (d *DataImpl) GetErrorList() []snyk_errors.Error {
...
+	logger.Debug().Msg("payload is larger than threshold, writing it to disk")
+	return writePayloadToDisk(id, config, bytes, logger)
+}
+
+func writePayloadToDisk(id Identifier, config configuration.Configuration, data []byte, logger *zerolog.Logger) Location {
+	tempDir := config.GetString(configuration.TEMP_DIR_PATH)
+	filepath := fmt.Sprintf("%s/workflow.%s.%d", tempDir, id.Path, time.Now().Unix())
+
+	logger.Debug().Msgf("Writing payload to file: %v", filepath)
+	if err := os.WriteFile(filepath, data, 0755); err != nil {
```
PeterSchafer, line 293: Question: I remember that we talked about keeping a shasum to at least detect broken files. Did this prove not feasible or is there any reason not to do it?

pkg/workflow/dataimpl.go
```diff
@@ -134,3 +251,56 @@ func (d *DataImpl) GetErrorList() []snyk_errors.Error {
...
+		logger.Error().Msgf("Error writing to file: %v", err)
+		return Location{
+			Path: "",
+			Type: InMemory,
+		}
+	}
+
+	logger.Debug().Msgf("Payload written to file: %v", filepath)
```
PeterSchafer, line 301: Suggestion on logging: There is a lot of debug logs being added here, but I doubt they are always required, how about moving some of them to trace level and maybe double checking that they have value. 
Just imagine that this logic is called multiple times in the dataflow, maybe even in parallel, how can we make sure the log messages make sense? For network requests, we included the a pointer in the log message to be able to separate the same message.

pkg/workflow/dataimpl.go
j-luong: nope, removed

pkg/workflow/types.go
j-luong: removed

pkg/workflow/dataimpl.go
j-luong: We can keep a shasum of the data if we save to disk, is there much value in adding validation when reading from disk?

sandor-trombitas (approved):

PR #236, opened 2024-09-10
feat: import specifications from generated artefacts

thisislawatts:
There are few things in this changeset:

- Import script for inlining generated specification artefacts. Verify by running `./scripts/pull-down-test-api-spec.sh`
- Experimental Cue based workflow for transforming data between specifications.


scripts/pull-down-test-api-spec.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+API_SPEC_PATH=../dragonfly
+# Navigate to repo
+cd $API_SPEC_PATH
```
PeterSchafer, line 5: Question: Do we need an initial clone?

scripts/pull-down-test-api-spec.sh
```diff
@@ -0,0 +1,27 @@
...
+cp -r $API_SPEC_PATH/tsp-output/@typespec/openapi3/ ./internal/cue_utils/source/openapi
+# Import types for go
+cp $API_SPEC_PATH/tsp-output/go/typespec_gen.go ./internal/dragonfly/dragonfly.go
+
+# Rename imported go types
+# In the future the package name will be configurable
+sed -i "" 's/package\ presentation/package\ dragonfly/g' internal/dragonfly/dragonfly.go
+
+cd ./internal/cue_utils
+go generate
```
cmars, line 27:
```suggestion
make generate
```

Then the `generate` target can depend on tooling like `oapi-codegen` and `cue` being installed.

internal/cue_utils/source/openapi/test.spec.yaml
cmars, line 1: I think this might be a duplicate copy of `.../rest/test.spec.yaml`?

internal/cue_utils/embed_gen.go
```diff
@@ -0,0 +1,17 @@
+package cue_utils
```
cmars, line 1: Underscores in Go packages are allowed but not conventional... Suggest renaming the directory & package to 

```suggestion
package cueutils
```

cmars: A package-level Godoc comment might be a useful place to explain what this is and what it's for. It'd go into https://pkg.go.dev/github.com/snyk/go-application-framework

internal/cue_utils/transformer_test.go
```diff
@@ -0,0 +1,95 @@
...
+	cuejson "cuelang.org/go/pkg/encoding/json"
+)
+
+func TestNewTransformer_ValidTransformToTestApiFromCliTestManaged(t *testing.T) {
+	ctx := cuecontext.New()
+
+	transformer, err := NewTransformer(ctx, ToTestApiFromCliTestManaged)
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
```
cmars, line 19: Suggest using https://pkg.go.dev/github.com/attic-labs/testify/assert for assertions, they're much more ergonomic and conventionally used elsewhere in the GAF.

```suggestion
	assert.NoError(t, err)
```

scripts/pull-down-test-api-spec.sh
thisislawatts: Switched to: https://github.com/snyk/go-application-framework/pull/236/commits/a8d513af994df83d33cce76a416363957d9cd12f

internal/cueutils/convert/to_cli/from_testapi.cue
```diff
@@ -0,0 +1,17 @@
...
+input: {
+    test: testapi.#TestResource
+    findings: [...testapi.#FindingResource]
+}
+
+output: {
+    findings: [ for finding in input.findings { finding.attributes } ]
+    summary: input.test.attributes.summary
+    outcome: input.test.attributes.outcome
+    severity_order: ["high", "medium", "low"]
```
PeterSchafer, line 16: IIRC this is supposed to be in ASC order

internal/cue_utils/embed_gen.go
thisislawatts: Renamed in: https://github.com/snyk/go-application-framework/pull/236/commits/86da075ee9401380a34c0a22b036b60518b8c225

thisislawatts: https://github.com/snyk/go-application-framework/pull/236/commits/1769935b2307be2148d36892880f3089155195c6

internal/cueutils/transformer_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if transformer == nil {
+		t.Error("Expected a non-nil transformer")
+	}
+
+	input := loadJsonFile(t, "cli-json-test-npm.json")
+	_, applyError := transformer.Apply(input)
+	if applyError == nil {
+		t.Fatal(err)
+	}
```
sandor-trombitas, line 29: I'm not sure i'm reading this test correct. Are we expecting `transformer.Apply` to return an error for our input?
```suggestion
	if applyError != nil {
        	t.Fatalf("Unexpected error applying transformation: %v", applyError)
    }
```

internal/cueutils/transformer_test.go
```diff
@@ -0,0 +1,95 @@
...
+	transformer, err := NewTransformer(ctx, ToCliFromTestApi)
+
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	if transformer == nil {
+		t.Error("Expected a non-nil transformer")
+	}
+}
```
sandor-trombitas, line 61: Is this test complete? Can we add some positive test cases: The existing tests seem to expect errors We should include positive test cases where we expect successful transformations and verify the results.

```go
expectedOutput := loadJsonFile(t, "expected-test-api-output.json")
    if !reflect.DeepEqual(output, expectedOutput) {
        t.Errorf("Transformation output does not match expected result.\nGot: %v\nWant: %v", output, expectedOutput)
    }
```

internal/cueutils/transformer_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func TestNewTransformer_InvalidTransform(t *testing.T) {
+	ctx := cuecontext.New()
+	_, err := NewTransformer(ctx, "invalid_transform.cue")
+	if err == nil {
+		t.Error("Expected an error for invalid transform")
+	}
+}
```
sandor-trombitas, line 69: ⛏️ _Nitpick_: Can we include some other edge cases, such as empty input, malformed data, or boundary conditions specific to each transformation type.

internal/cueutils/transformer_test.go
thisislawatts: Correct, fixed in: https://github.com/snyk/go-application-framework/pull/236/commits/0d1b77d8990207bcbb3be1602d51af8a3b984d4d

scripts/pull-down-test-api-spec.sh
thisislawatts: Added a clone if the directory does not exist

scripts/pull-down-test-api-spec.sh
```diff
@@ -0,0 +1,27 @@
...
+cd -
+
+# Vendor OpenAPI build artefacts for use in cue
+cp -r $API_SPEC_PATH/tsp-output/@typespec/openapi3/ ./internal/cue_utils/source/openapi
+# Import types for go
+cp $API_SPEC_PATH/tsp-output/go/typespec_gen.go ./internal/dragonfly/dragonfly.go
+
+# Rename imported go types
+# In the future the package name will be configurable
+sed -i "" 's/package\ presentation/package\ dragonfly/g' internal/dragonfly/dragonfly.go
```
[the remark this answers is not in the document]
thisislawatts: Adopted in: https://github.com/snyk/go-application-framework/pull/236/commits/f8d128a6b29b361941618fc300ec6103e2383d73

internal/cueutils/transformer_test.go
thisislawatts: I am definitely up for introducing these test cases @sandor-trombitas but I would like to do that in follow up iteration as we develop a better understanding of the failure modes we want to get from this module.

tools.go
```diff
@@ -0,0 +1,7 @@
+//go:build tools
+
+package go_application_framework
```
PeterSchafer, line 3: Question: What is this file about?

thisislawatts: Adopting this pattern:
https://www.jvt.me/posts/2022/06/15/go-tools-dependency-management/

scripts/pull-down-test-api-spec.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
cmars: Inlined models in a union I think

thisislawatts:
We missing types for transformed Cue data objects, this means the output of `internal/cueutils/convert/to_cli/from_testapi.cue`

scripts/pull-down-test-api-spec.sh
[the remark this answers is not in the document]
thisislawatts: Resolved by recent fix to underlying spec

internal/cueutils/transformer_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestNewTransformer_ValidTransformToTestApiFromCliTestManaged_Malformed(t *testing.T) {
+	ctx := cuecontext.New()
+
+	transformer, err := NewTransformer(ctx, ToTestApiFromCliTestManaged)
+
+	assert.NoError(t, err)
+	assert.NotNil(t, transformer, "Expected a non-nil transformer")
+
+	input := loadJsonFile(t, "cli-json-test-npm.malformed.json")
+	n, applyError := transformer.Apply(input)
```
cmars, line 24: Nitpick: prefer `err` for errors, unless there's enough context or scoping conflicts to require differentiation. It's a preference, but also a prevailing style in Go code. `err` tends to stand out better.

```suggestion
	n, err := transformer.Apply(input)
```

cmars (approved):
LGTM!

Some ideas on how to keep the output schema & Go structs in sync but fine for followups.

internal/cueutils/transformer.go
```diff
@@ -0,0 +1,98 @@
...
+	if err := withOutput.Err(); err != nil {
+		return nil, fmt.Errorf("failed to get output: %w", err)
+	}
+
+	// Convert from Cue.Value to relevant go type
+	codec := gocodec.New(t.inst.Context(), &gocodec.Config{})
+	var localFinding LocalFinding
+
+	// Gate with validation before encoding?
+	encodeErr := codec.Encode(withOutput, &localFinding)
```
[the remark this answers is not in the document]
cmars: Nevermind, looks like we have them... Do tests fail if we remove them?

internal/cueutils/convert/to_cli/from_testapi.cue
```diff
@@ -0,0 +1,18 @@
...
+
+input: _
+
+input: {
+    test: testapi.#TestResource
+    findings: [...testapi.#FindingResource]
+}
+
+// This should align with
+// LocalFinding definition
```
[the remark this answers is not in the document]
PeterSchafer: I think Option 2 would mean we are detached from the dragonfly types. Which I would like to avoid to be honest.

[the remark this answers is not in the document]
PeterSchafer: Is there maybe an **Option 3** that makes use of any of the [encodings in cue](https://pkg.go.dev/cuelang.org/go/encoding@v0.10.0) instead of defining these cue types in code, we should be able to import them from something that comes from dragonfly. 
So that dragonfly would still be the single source of truth.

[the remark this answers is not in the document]
cmars: I think we'd want to do that either way (reference the original Dragonfly types)!

With Option 1, you'd import the test API types into a new Typespec model defining the output schema.

With Option 2, you'd reference the Dragonfly-derived Cue types from `testapi` in the output schema.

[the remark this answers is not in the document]
cmars: Option 2 would look like:

```
output: {
    test: testapi.#TestResource
    findings: [...testapi.#FindingResource]
    outcome: // TODO: define the shape for this one...
}
output: {
    // the concrete values you've already got here...
}
```

internal/cueutils/schemas/cli_test_managed.cue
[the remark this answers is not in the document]
thisislawatts: Added in https://github.com/snyk/go-application-framework/pull/236/commits/e4785b7d7afa3e8feed4c8e926790aa6f19cc256

internal/cueutils/transformer.go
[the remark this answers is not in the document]
thisislawatts: I added these in after out conversation the tests were passing before hand as well. Working out _why_ it was working without the annotations is an investigation for another day :)

internal/cueutils/transformer_test.go
thisislawatts: Aligned https://github.com/snyk/go-application-framework/pull/236/commits/c3ff5f653fbc781dbff2ed21d8683fd8a22cd8a2

sandor-trombitas (approved):
💯

PR #237, opened 2024-09-12
feat: add possibility to overwrite log level in roundtripper

bastiandoetsch:


pkg/networking/networking.go
```diff
@@ -96,20 +97,19 @@ func LogResponse(response *http.Response, logger *zerolog.Logger) {
 type defaultHeadersRoundTripper struct {
 	encapsulatedRoundTripper http.RoundTripper
 	networkAccess            *networkImpl
+	logLevel                 zerolog.Level
```
PeterSchafer, line 100: Suggestion: please assign the new member in Clone() as well and extend the associated test. Thank you!

bastiandoetsch: Which Clone() do you mean? This struct does not have a Clone() func.

ShawkyZ (approved):

PR #239, opened 2024-09-16
fix: config persistence when config file is empty

ShawkyZ:
In some cases the settings file exists but it doesn't have any content.
I'm not sure why this happens.
Nonetheless I changed the Set func to check if file has no content. 
If it doesn't it will set it as empty json so that it won't fail on deserialize


acke (approved):
LGTM!

PR #244, opened 2024-09-19
test: test in memory feature is disabled when no config passed to newData

j-luong:


pkg/workflow/dataimpl_test.go
```diff
@@ -126,4 +134,57 @@ func Test_NewData(t *testing.T) {
 		assert.Equal(t, 1, len(actualFiles))
 		assert.Contains(t, actualFiles[0].Name(), expectedFileName)
 	})
+
+	t.Run("disables feature when config is not passed to NewData", func(t *testing.T) {
```
thisislawatts, line 138: **nit:** consider this as an alternative test name: _when configuration is not provided, filesystem cache is not used_

thisislawatts (approved):

PR #245, opened 2024-09-23
fix: race writing settings file

ShawkyZ:


bastiandoetsch (approved):

pkg/configuration/configuration.go
```diff
@@ -212,11 +212,10 @@ func (ev *extendedViper) Clone() Configuration {
 // Set sets a configuration value.
 func (ev *extendedViper) Set(key string, value interface{}) {
 	ev.mutex.Lock()
+	defer ev.mutex.Unlock()
```
PeterSchafer, line 215: Issue: This change has the potential to create deadlocks! The mutex should not be held when leaving the critical section, in this case the critical section is `Set()` on the wrapped viper instance.

PeterSchafer: The problem arises with `ev.storage.Set()`, which is not part of the critical section anymore.

PeterSchafer: @ShawkyZ which part of writing settings was having a race condition before?

ShawkyZ: There was a race that resulted in corrupt json file. 
Currently jsonStorage is not called from anywhere outside of the ev set.

But I agree that it could result in deadlocks in the future. If jsonStorage.Set called directly.
I will open another PR with a mutex in jsonStorage and reverting this change here.

PeterSchafer: In general, for any change in gaf, it is important to keep in mind that this is a library/sdk that might be used correctly for now but future usage could introduce issues if undocumented assumptions are not met anymore.

ShawkyZ: I would prefer to use the same lock actually. What do you think of passing an optional defer Func that will release the lock when done (in case of extendedViper)

PR #253, opened 2024-10-01
chore: introduce helper function to register a feature flag accessible via the configuration

PeterSchafer:
* helper function to register feature flag
* move existing functionality to use helper.

### DIFF
--- a/pkg/local_workflows/config_utils/feature_flag.go
+++ b/pkg/local_workflows/config_utils/feature_flag.go
@@ -1,0 +1,30 @@
+package config_utils
+
+import (
+	"github.com/snyk/go-application-framework/internal/api"
+	"github.com/snyk/go-application-framework/pkg/configuration"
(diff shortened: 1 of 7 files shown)


pkg/local_workflows/config_utils/feature_flag.go
```diff
@@ -0,0 +1,30 @@
+package config_utils
```
sergiu-snyk, line 1: maybe move this package outside `local_workflows` if it can be used by external workflows as well?

PeterSchafer: In general it can be used outside already as it is in `pkg`. `local_workflows` is a currently misnamed package that collects logic that is less technical platform and more re-usable domain logic. In another discussion we will try to improve this properly, until then I fall back to putting things here if this makes sense.

j-luong (approved):

PR #255, opened 2024-10-07
feat: add func to load shell environment variables

bastiandoetsch:
move func to load shell environment from language server to GAF. 

The command injection finding is a false positive (IMHO).

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -236,7 +236,7 @@ func customConfigFiles(config configuration.Configuration) configuration.Default
 		files = append(files, ".snyk.env."+runtime.GOOS)
 		files = append(files, ".envrc."+runtime.GOOS)
 
-		configFile := config.GetString(configuration.CONFIG_FILE)
+		configFile := config.GetString("configfile")
 		if configFile != "" {
 			files = append(files, configFile)
 		}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -219,10 +219,10 @@ func initConfiguration(engine workflow.Engine, config configuration.Configuratio
 
 	config.AddDefaultValue(configuration.INPUT_DIRECTORY, defaultInputDirectory())
 	config.AddDefaultValue(configuration.PREVIEW_FEATURES_ENABLED, defaultPreviewFeaturesEnabled(engine, logger))
-	config.AddDefaultValue(configuration.CUSTOM_CONFIG_FILES, CustomConfigFiles(config))
+	config.AddDefaultValue(configuration.CUSTOM_CONFIG_FILES, customConfigFiles(config))
 }
 
-func CustomConfigFiles(config configuration.Configuration) configuration.DefaultValueFunction {
+func customConfigFiles(config configuration.Configuration) configuration.DefaultValueFunction {
 	return func(existingValue interface{}) interface{} {
 		var files []string
 		// last file usually wins if the same values are configured
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -233,6 +233,8 @@ func CustomConfigFiles(config configuration.Configuration) configuration.Default
 
 		files = append(files, ".snyk.env")
 		files = append(files, ".envrc")
+		files = append(files, ".snyk.env."+runtime.GOOS)
+		files = append(files, ".envrc."+runtime.GOOS)
 
 		configFile := config.GetString(configuration.CONFIG_FILE)
 		if configFile != "" {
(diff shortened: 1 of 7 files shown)


pkg/app/app.go
```diff
@@ -218,6 +219,36 @@ func initConfiguration(engine workflow.Engine, config configuration.Configuratio
 
 	config.AddDefaultValue(configuration.INPUT_DIRECTORY, defaultInputDirectory())
 	config.AddDefaultValue(configuration.PREVIEW_FEATURES_ENABLED, defaultPreviewFeaturesEnabled(engine, logger))
+	config.AddDefaultValue(configuration.CUSTOM_CONFIG_FILES, CustomConfigFiles(config))
+}
+
+func CustomConfigFiles(config configuration.Configuration) configuration.DefaultValueFunction {
```
PeterSchafer, line 225: Suggestion: I think there is no need to export this.

```suggestion
func customConfigFiles(config configuration.Configuration) configuration.DefaultValueFunction {
```

pkg/configuration/environment.go
```diff
@@ -0,0 +1,122 @@
...
+
+	if !shellWhiteList[shell] {
+		return gotenv.Env{}
+	}
+
+	ctx, cancelFunc := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancelFunc()
+
+	// deepcode ignore CommandInjection: false positive
+	env, err := exec.CommandContext(ctx, shell, "--login", "-i", "-c", "printenv && exit").Output()
```
PeterSchafer, line 84: Question: I'm struggling to understand how this works in a cross platform environment. Especially the windows case seems to be missing.

pkg/configuration/environment.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/subosito/gotenv"
+)
+
+// LoadConfiguredEnvironment updates the environment with local configuration. Precedence as follows:
+//  1. std folder-based config files
+//  2. given command-line parameter config file
+//  3. std config file in home directory
+//  4. global shell configuration
+func LoadConfiguredEnvironment(config Configuration) {
```
PeterSchafer, line 19: Suggestion: while it seems reasonable that this is part of the `configuration` package, I would like to not export any of these functions and load the environment inside the configuration automatically. The reason for this is, that I would like to continue what viper does and hide implementation details about the source of configuration values. In the future we might even have an improved design to add configuration provider besides what is existing in viper. but since this is not available yet, let's try to remove this from the public API of the package.

Having said this, we need a way to enable additional sources of configuration, for example as a configuration value. This should also ensure backwards compatibility for the CLI where we don't want to create regression by reading from an additional source.

pkg/configuration/constants.go
```diff
@@ -28,8 +28,10 @@ const (
 	CACHE_PATH                      string = "snyk_cache_path"
 	TIMEOUT                         string = "snyk_timeout_secs"
 	LOG_LEVEL                       string = "snyk_log_level" // string that defines the log level based on zerolog levels (trace,debug,info,...)
+	CONFIG_FILE                     string = "configfile"     // command line flag from language server
```
PeterSchafer, line 31: Issue: Extension specific configuration shouldn't be reflected here. The config value is already registered [here](https://github.com/snyk/snyk-ls/blob/main/ls_extension/main.go#L50) so the question is how to use it without creating circular dependencies. `configfile` as a name is globally seen not very specific. I think I will need some time to have a good solution.

pkg/configuration/environment.go
bastiandoetsch: I accidentally deleted the windows early return. Oops.

pkg/app/app.go
bastiandoetsch: Let me check - language server had used it in standalone-mode, but maybe it's now obsolete.

pkg/configuration/environment.go
bastiandoetsch: I need to call it from LS before a scan to get folder specific environment settings. As the legacycli workflow would need to do, too. So we'll need it as it needs to be at runtime. Or we need to integrate it into being called each time when a workflow is invoked. 

Regarding the consistent configuration, I'd like that - should we set each loaded variable into the config manually, or how would you propose to handle that?

pkg/configuration/constants.go
bastiandoetsch: removed it for now, so it's back to the status before.

PeterSchafer (approved):

PR #258, opened 2024-10-07
feat: set api url from jwt audience claim

cmars:
Snyk's OAuth implementation is capable of indicating the environment which the user is authenticated into and authorized to access.

This is specified in the audience JWT claim ("aud"). Snyk's implementation of this claim contains an array of strings, per RFC 7519.

If set and non-empty, the first audience URL is taken as the default API URL that the client should use, unless the endpoint was specifically configured.

---

- [ ] TODO: validate w/IDE team
- [ ] TODO: validate w/auth folks, work out env switching rabbit holes

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -14,3 +14,4 @@ b0ba7f6ed181c23a5c6532cf6f124b731d390f86:internal/presenters/testdata/with-ignor
 237c7f05ec087733fa7929ee9fa3db2bd56bdba4:pkg/logging/scrubbingLogWriter_test.go:github-pat:183
 237c7f05ec087733fa7929ee9fa3db2bd56bdba4:pkg/logging/scrubbingLogWriter_test.go:snyk-api-token:208
 0481cdb4d07351149e65a57ebc9ad5b983896849:pkg/auth/oauth2authenticator.go:generic-api-key:30
+0a646dd3b9eeca0463fd8240b87957106f4b71f3:pkg/app/app_test.go:jwt:57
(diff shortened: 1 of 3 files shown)


pkg/auth/oauth2authenticator.go
```diff
@@ -245,6 +247,13 @@ func (o *oAuth2Authenticator) IsSupported() bool {
 }
 
 func (o *oAuth2Authenticator) persistToken(token *oauth2.Token) error {
+	if audience, err := readAudience(token); err == nil && audience != "" {
```
PeterSchafer, line 250: Suggestion: how about making this part of the default function for the API URL instead of adapting the `oAuth2Authenticator` and persisting the api url. To me this logic seems to be more snyk specific then an oauth specific change.

cmars: I think I see what you mean... parse it out of the stored oauth token if it's configured? Yes... that's a much better approach!

cmars: Updated, with a unit test

PeterSchafer:
Very Nice!!!

thisislawatts:
Given the scenario that an `API_URL` has already been set, perhaps in error. If we get a **new** endpoint in the auth_token we should respect the `aud` claim and remove/unset the `API_URL` configuration so it can instead be loaded from the `auth.CONFIG_KEY_OAUTH_TOKEN` by the default Func handler. 

https://github.com/snyk/go-application-framework/blob/main/pkg/auth/oauth2authenticator.go#L250

PR #259, opened 2024-10-09
fix: do not include ignored issues by default [CLI-506]

thisislawatts:
Updates the rendering to better align with the pretty print formatting. Running the test command with and without the fflag is the best way to validate the output. 

You can also review the snapshots captured in:
- internal/presenters/__snapshots__/presenter_local_finding_test.snap
- internal/presenters/__snapshots__/presenter_sarif_results_pretty_test.snap


internal/presenters/presenter_local_finding.go
```diff
@@ -151,24 +151,37 @@ func (p *LocalFindingPresenter) Render() (string, error) {
...
 	}{
-		Summary:        sum,
-		Results:        p.Input,
-		Order:          []string{"low", "medium", "high"},
-		ShowIgnored:    p.ShowIgnored,
-		SeverityFilter: p.SeverityMinLevel,
+		Summary:         sum,
+		Results:         p.Input,
+		OpenFindings:    filterOutIgnoredFindings(p.Input.Findings),
+		IgnoredFindings: p.Input.Findings,
```
thisislawatts, line 165: Issue: This is currently not used. We want to extend the test cases to cover when ignored issues should be included in the output.

internal/cueutils/convert/to_testapi/from_sarif.cue
```diff
@@ -140,7 +140,14 @@ _findings: list.Sort(list.Concat([ for run in input.runs {
                     }
                 }
             }]
+            if result.suppressions != _|_ {
+            	if result.suppressions[0].justification != _|_ {
```
cmars, line 144: Do we care if there's no justification given? I think we'd still want to honor the ignore, if the finding is suppressed.

```suggestion
            if result.suppressions != _|_ {
              if len(result.suppressions) > 0 {
```

internal/cueutils/convert/to_testapi/from_sarif.cue
```diff
@@ -140,7 +140,14 @@ _findings: list.Sort(list.Concat([ for run in input.runs {
...
+								suppression: {
+									kind: "other" // TODO: Hard-coded as this does not map well to kind|state fields from sarif
```
cmars, line 146: Another idea, it might be nice to specify `kind` and `justification` as

```suggestion
								suppression: kind: "ignored"
```

and then below, in an `if len(result.suppressions) > 0`:

```
							suppression: justification: result.suppressions[0].justification
```

sandor-trombitas (approved):

thisislawatts:
Looking into a failure with
```
{
  "ok": false,
  "error": "failed to set input: 0.1.suppression: cannot use null (type null) as (string|bytes|list|struct) in argument 1 to len",
  "path": "/Users/luke/snyk/snyk-goof"
}
```

PR #260, opened 2024-10-09
feat: modify auth token url based on instance param

cmars:
If an instance parameter is provided in the redirect, use it to modify
the URL from where the oauth token is obtained.

The instance provided is the Snyk region domain, minus the api. host
prefix.

Stacked on #258


pkg/auth/oauth2authenticator.go
```diff
@@ -402,6 +423,16 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
 	return err
 }
 
+func redirectAuthHost(instance string) string {
+	return fmt.Sprintf("api.%s", instance)
+}
+
+var redirectAuthHostRE = regexp.MustCompile(`^api\.(.+)\.snyk\.io$`)
```
DarrellMozingo, line 430: snykgov too!

pkg/app/app.go
```diff
@@ -91,6 +101,61 @@ func defaultFuncApiUrl(logger *zerolog.Logger) configuration.DefaultValueFunctio
...
+		// NOTE: The original jws package models audience with a string, not a
+		// []string. This fails to parse Snyk JWTs.
+		Aud []string `json:"aud"`
+	}{}
+	err = json.NewDecoder(bytes.NewBuffer(decoded)).Decode(&c)
+	if err != nil {
+		return "", err
+	}
+	if len(c.Aud) > 0 {
+		return c.Aud[0], nil
```
ShawkyZ, line 154: could we by any change have multiple entries in the `aud` claim ?

pkg/app/app.go
```diff
@@ -91,6 +101,61 @@ func defaultFuncApiUrl(logger *zerolog.Logger) configuration.DefaultValueFunctio
...
+		return "", errors.New("jws: invalid token received")
+	}
+	decoded, err := base64.RawURLEncoding.DecodeString(s[1])
+	if err != nil {
+		return "", err
+	}
+	c := struct {
+		// NOTE: The original jws package models audience with a string, not a
+		// []string. This fails to parse Snyk JWTs.
+		Aud []string `json:"aud"`
```
ShawkyZ, line 147: it doesn't feel correct to use the `aud` claim for fetching tokens.
We are using aud claim to determine where to fetch the access token from. 
Afaik aud claim should only be used to identify the audience/recipients for this token.
https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3

pkg/app/app.go
thisislawatts: From the spec:
>  In the general case, the "aud" value is an array of case-
   sensitive strings, each containing a StringOrURI value.  In the
   special case when the JWT has one audience, the "aud" value MAY be a
   single case-sensitive string containing a StringOrURI value.  The
   interpretation of audience values is generally application specific.
   Use of this claim is OPTIONAL.
 https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3
  
 It is almost expected for there to be multiple values available. We should likely search the list and extract the first one which meets our criteria.

cmars: Guidance from the IdP team is to use the first value. I can provide more context offline.

thisislawatts (approved):

PeterSchafer (approved):

PR #263, opened 2024-10-14
fix: windows support for cueutils

cmars:
Embedded cue module is loaded with a custom loader that resolves absolute paths, even though the paths are "virtual".

Adding a drive-letter seems to resolve problems with this when running on a Windows OS.

CLI-514

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,9 @@ jobs:
       version: "current"
 
     steps:
+      - run:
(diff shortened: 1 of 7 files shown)


thisislawatts (commented):
**praise:** Great work debugging this. Interesting to see the distro specific files here. I had seen that convention within stdlib but had not made the connection that we could adopt the same pattern.

sandor-trombitas (approved):

PR #265, opened 2024-10-15
fix: only read supported env vars (CLI-526)

j-luong:
This PR makes changes so that environment variables that are supported by GAF are configurable as opposed to GAF automatically detecting supported environment variables.

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -1,6 +1,7 @@
 package configuration
 
 import (
+	"fmt"
 	"net/url"
 	"os"
 	"path"
@@ -226,7 +227,9 @@ func (ev *extendedViper) get(key string) interface{} {
 	ev.mutex.Lock()
 	defer ev.mutex.Unlock()
 
-	// try to lookup given key
+	// try to lookup given key with SNYK_ prefix
+	ev.viper.SetEnvPrefix("snyk")
+	//ev.viper.BindEnv(key)
 	result := ev.viper.Get(key)
 	isSet := ev.viper.IsSet(key)
 
@@ -236,6 +239,7 @@ func (ev *extendedViper) get(key string) interface{} {
 	alternativeKeysSize := len(alternativeKeys)
 	for !isSet && index < alternativeKeysSize {
 		altKey := alternativeKeys[index]
+		fmt.Println(altKey, ": ", os.Getenv(strings.ToUpper(altKey)))
 		result = ev.viper.Get(altKey)
 		isSet = ev.viper.IsSet(altKey)
 		index++
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -1,8 +1,6 @@
 package configuration
 
 import (
-	"fmt"
-	"github.com/snyk/error-catalog-golang-public/snyk_errors"
 	"net/url"
 	"os"
 	"path"
@@ -62,17 +60,26 @@ type Configuration interface {
 // extendedViper is a wrapper around the viper library.
 // It adds support for default values and alternative keys.
 type extendedViper struct {
-	viper           *viper.Viper
-	alternativeKeys map[string][]string
-	defaultValues   map[string]DefaultValueFunction
-	configType      configType
-	flagsets        []*pflag.FlagSet
+	viper               *viper.Viper
+	alternativeKeys     map[string][]string
+	defaultValues       map[string]DefaultValueFunction
+	configType          configType
+	flagsets            []*pflag.FlagSet
+	storage             Storage
+	mutex               sync.Mutex
+	automaticEnvEnabled bool
+	configFiles         []string
 
 	// persistedKeys stores the keys that need to be persisted to storage when Set is called.
 	// Only specific keys are persisted, so viper's native functionality is not used.
 	persistedKeys map[string]bool
-	storage       Storage
-	mutex         sync.Mutex
+
+	// supportedEnvVarPrefixes store the namespace prefixes that should be supported.
+	// Any env var without these prefixes, will be ignored by the configuration.
+	supportedEnvVarPrefixes []string
+
+	// supportedEnvVars store the env vars that should be supported REGARDLESS of its prefix. e.g. NODE_EXTRA_CA_CERTS
+	supportedEnvVars []string
 }
 
 // StandardDefaultValueFunction is a default value function that returns the default value if the existing value is nil.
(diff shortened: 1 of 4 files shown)


pkg/configuration/configuration.go
```diff
@@ -222,11 +223,27 @@ func (ev *extendedViper) Set(key string, value interface{}) {
 	}
 }
 
+func bindEnvSnykErr(key string, err error) snyk_errors.Error {
+	return snyk_errors.Error{
+		Title:          "Unable to bind env",
+		Classification: "ACTIONABLE",
```
PeterSchafer, line 229: Question: how is it possible for a user to action on this error?

pkg/configuration/configuration.go
```diff
@@ -222,11 +223,27 @@ func (ev *extendedViper) Set(key string, value interface{}) {
...
 func (ev *extendedViper) get(key string) interface{} {
 	ev.mutex.Lock()
 	defer ev.mutex.Unlock()
 
-	// try to lookup given key
+	// ensure we only get Snyk related keys (prefixed with SNYK_)
+	if ev.GetKeyType(key) == EnvVarKeyType {
+		err := ev.viper.BindEnv(key, strings.ToUpper(key))
+		if err != nil {
+			return bindEnvSnykErr(key, err)
```
PeterSchafer, line 243: Issue: I'm a bit concerned about this error handling here. Returning an error instead of a value is misleading.

PeterSchafer: Looking at the error cases in `BindEnv()` it can be avoided or ignored since the only error case is, that there is no input. 

But we should definitely not stop further processing here if something fails.

pkg/configuration/configuration.go
```diff
@@ -222,11 +223,27 @@ func (ev *extendedViper) Set(key string, value interface{}) {
...
+		}
+	}
+	// try to lookup given key with SNYK_ prefix
```
PeterSchafer, line 246: Suggestion: instead of commenting on an implementation limitation, why don't we make supported prefixes configurable. That shouldn't be too much work.

pkg/configuration/configuration.go
```diff
@@ -222,11 +223,27 @@ func (ev *extendedViper) Set(key string, value interface{}) {
...
+}
+
 func (ev *extendedViper) get(key string) interface{} {
 	ev.mutex.Lock()
 	defer ev.mutex.Unlock()
 
-	// try to lookup given key
+	// ensure we only get Snyk related keys (prefixed with SNYK_)
+	if ev.GetKeyType(key) == EnvVarKeyType {
+		err := ev.viper.BindEnv(key, strings.ToUpper(key))
```
PeterSchafer, line 241: Suggestion: The docs for the function indicate that a single value does exactly what this code does, so let's keep the code clean and not add two params.

PeterSchafer, line 241: BTW, what happens when we call this multiple times? looks like variables are appended each time, do we want this? Smells like a memory leak to me.

pkg/configuration/configuration.go
j-luong: Alternative take, we support only the `SNYK_` prefix and use this as an opportunity to implement consistency by requiring all env vars we want to support to be prefixed with `SNYK_`?

pkg/configuration/configuration.go
j-luong: > Issue: I'm a bit concerned about this error handling here. Returning an error instead of a value is misleading.

Yeh, this is abusing the `interface{}` return type a bit here, we should be consistent in what we return.

> Looking at the error cases in BindEnv() it can be avoided or ignored since the only error case is, that there is no input.

We can't be sure this will always be the case, would logging the error and returning nil be a good compromise?

pkg/configuration/configuration.go
j-luong: I think given the feedback below, it's not necessary to have an error catalog error here?

pkg/configuration/configuration.go
j-luong: That would be a problem!

pkg/configuration/configuration.go
```diff
@@ -55,22 +55,41 @@ type Configuration interface {
 	PersistInStorage(key string)
 	SetStorage(storage Storage)
 	GetStorage() Storage
+
+	AutomaticEnv()
+	SetAutomaticEnv(b bool)
```
PeterSchafer, line 60: Suggestion: I think the setter is a bit problematic as it creates the possibility for ambiguity, since you can set the parameter to false, which will have no effect. Better remove the setting and only have the getter `GetAutomaticEnv()` and `AutomaticEnv()`

pkg/configuration/configuration.go
```diff
@@ -197,6 +268,11 @@ func (ev *extendedViper) Clone() Configuration {
 		clone.AddDefaultValue(k, v)
 	}
 
+	clone.SetAutomaticEnv(ev.GetAutomaticEnv())
```
PeterSchafer, line 271: Suggestion: based on my previous comment about the setter.
```suggestion
	if ev.GetAutomaticEnv() {
	  clone.AutomaticEnv()
	}
```

pkg/configuration/configuration.go
```diff
@@ -484,9 +575,87 @@ func (ev *extendedViper) GetAllKeysThatContainValues(key string) []string {
...
+// SetFiles sets the files from which the Configuration instance uses
+func (ev *extendedViper) SetFiles(files ...string) {
+	ev.mutex.Lock()
+	defer ev.mutex.Unlock()
+
+	ev.configFiles = append(ev.configFiles, files...)
+}
+
+func (ev *extendedViper) GetFiles() []string {
+	return ev.configFiles
```
PeterSchafer, line 660: Suggestion: I know it is not consistent in the existing code, but you should be locking the mutex even in a getter.

j-luong: I used `ev.mutex.TryLock()` in these instances to avoid hitting a deadlock with the calling method's (`Clone()`) `mutex.Lock`. 

Open to alternatives, if `TryLock` is not appropriate here

pkg/configuration/configuration.go
```diff
@@ -484,9 +577,99 @@ func (ev *extendedViper) GetAllKeysThatContainValues(key string) []string {
...
+func (ev *extendedViper) AutomaticEnv() {
+	ev.mutex.Lock()
+	defer ev.mutex.Unlock()
+
+	ev.viper.AutomaticEnv()
+	ev.automaticEnvEnabled = true
+}
+
+func (ev *extendedViper) GetAutomaticEnv() bool {
+	if ev.mutex.TryLock() {
```
PeterSchafer, line 608: TryLock is making the situation worse, as a will also fail if someone else locked the mutex, which is the opposite of what we want. Let's sync on a solution.

pkg/configuration/configuration.go
PeterSchafer: I think TryLock() is making the situation worse to be honest, as it'll fail if the mutex is already locked, since it is unknown who locked the mutex this effectively ignoring if someone else holds it. Which is exactly what we want to avoid using the mutex.

Some additional thoughts on solutions:
* The mutex we acquire during Clone() is supposed to ensure that the configuration to clone `ev` isn't changed while cloning.
* The deadlocks only happen when we use the getter of `ev`, possible solution access the fields directly without the locking public accessors.
* Use a [RWMutex](https://go.dev/pkg/sync/?m=old#RWMutex), which allows multiple reader but just a single writer.

From the complexity of the change I think avoiding getter in clone and still having all public methods use the mutex is the best.

pkg/configuration/configuration.go
```diff
@@ -115,32 +133,73 @@ func CreateConfigurationFile(filename string) (string, error) {
...
+		WithAutomaticEnv(),
+	)
 	return config
 }
 
 func (ev *extendedViper) getConfigType() configType {
 	return ev.configType
 }
 
+// Deprecated: Use NewWithOpts with configuration.WithAutomaticEnv() option instead
```
PeterSchafer, line 192: Question: I'm still a bit undecided myself, but let me still raise this question, do you think it is essential to deprecate these helpers? I'm asking since I understand the wish to keep the interface clean but on the other hand it creates quite some noise here and in the other PR. Wouldn't it be better to have the tests for example still use the old methods and ensure that the changes don't cause a change in behaviour.
From a reviewer perspective it would make it easier for sure.

pkg/configuration/configuration.go
PeterSchafer: BTW, taking the trylock change as an proof that we are missing concurrency tests, let's add one that fails. We have -race enabled so that should be easily discoverable.

pkg/configuration/configuration.go
j-luong: It's not essential, but IMO it's better to identify and ideally address these issues early, rather than defer them. Marking as deprecated seems a good compromise between identifying and fixing now vs keeping the status quo; it gives us the opportunity to point people to the desired implementation and hopefully avoids building tech debt.

Ideally, marking as deprecated would've been the only change, but the linter interpreted it as a fail, perhaps it would've been better to see if it was possible to change the lint error to a warn.

sandor-trombitas (approved):

PR #266, opened 2024-10-24
chore: specify config and writer for renderer

PeterSchafer:


internal/presenters/templates/local_finding.tmpl
```diff
@@ -22,7 +22,7 @@
 
 {{define "summary"}}{{ "Test Summary" | bold }}
 
-  Organization:      {{.Summary.Org}}
+  Organization:      {{ getValueFromConfig "internal_org_slug" }}
```
thisislawatts, line 25: **question:** Do we want to normalise use of `internal_` prefixed properites within the templates if we are aiming for this to become an external interface? 

Or do I misunderstand the purpose of "internal_org_slug"?

PeterSchafer: Sorry, this change can actually undone. It was meant for testing reading from the config. There is an explicit ticket to implement this.

PeterSchafer: Reverted!

thisislawatts: Thats what I thought :)

internal/presenters/presenter_local_finding.go
```diff
@@ -139,34 +129,68 @@ func LoadTemplates(files []string, tmpl *template.Template) error {
 	return nil
 }
 
-func (p *LocalFindingPresenter) Render() (string, error) {
-	localFindingsTemplate, err := template.New("local_finding").Parse("")
+func mergeMaps[K comparable, V any](mapA map[K]V, mapB map[K]V) map[K]V {
```
thisislawatts, line 132: **nit:** Could we move this to a util method to improve discoverability to encourage reuse in future

j-luong, line 132: question: is this better placed in the `util` package?

internal/presenters/presenter_local_finding.go
```diff
@@ -139,34 +129,68 @@ func LoadTemplates(files []string, tmpl *template.Template) error {
 	return nil
 }
```
j-luong, line 131: nitpick: would be good to describe the method in a comment

internal/presenters/presenter_local_finding.go
PeterSchafer: sure, I can move it

PR #267, opened 2024-10-25
CLI-545: Extract severity filter into a new workflow

sandor-trombitas:


pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/local_workflows/content_type"
+	"github.com/snyk/go-application-framework/pkg/local_workflows/local_models"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+	"github.com/spf13/pflag"
+)
+
+const (
+	FilterFindingsWorkflowName = "filterfindings"
```
j-luong, line 17:
```suggestion
	FilterFindingsWorkflowName = "findings.filter"
```

suggestion: consider following the pattern established in other workflows for naming `[subject].[action]`

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+var WORKFLOWID_FILTER_FINDINGS = workflow.NewWorkflowIdentifier(FilterFindingsWorkflowName)
+
+func InitFilterFindingsWorkflow(engine workflow.Engine) error {
+	flags := pflag.NewFlagSet(FilterFindingsWorkflowName, pflag.ExitOnError)
+	_, err := engine.Register(WORKFLOWID_FILTER_FINDINGS, workflow.ConfigurationOptionsFromFlagset(flags), filterFindingsEntryPoint)
+
+	// engine.GetConfiguration().AddDefaultValue(configuration.FF_TRANSFORMATION_WORKFLOW, configuration.StandardDefaultValueFunction(false))
```
j-luong, line 26: issue: leftover comment?

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,123 @@
...
+	return filteredFindings
+}
+
+func filterFindingsEntryPoint(invocationCtx workflow.InvocationContext, input []workflow.Data) (output []workflow.Data, err error) {
+	config := invocationCtx.GetConfiguration()
+	logger := invocationCtx.GetEnhancedLogger()
+	ff_transform_enabled := config.GetBool(configuration.FF_TRANSFORMATION_WORKFLOW)
+	output = []workflow.Data{}
+
+	logger.Println("Workflow Name:", FilterFindingsWorkflowName)
```
j-luong, line 85: question: will this set the appropriate log level? if not, consider using 

`logger.Debug().Msgf("Workflow Name: %s", FilterFindingsWorkflowName)`

sandor-trombitas: It will, Println sends a log event using debug level

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,122 @@
...
+		if strings.HasPrefix(data.GetContentType(), content_type.LOCAL_FINDING_MODEL) {
+			findingsBytes, ok := data.GetPayload().([]byte)
+			if !ok {
+				logger.Err(nil).Msg("invalid payload type")
+				return input, fmt.Errorf("invalid payload type")
+			}
+			unmarshalErr := json.Unmarshal(findingsBytes, &findingsModel)
+			if unmarshalErr != nil {
+				logger.Err(unmarshalErr).Msg(unmarshalErr.Error())
+				return input, unmarshalErr
```
j-luong, line 102: suggestion: similar here, probably want to wrap this in an error catalog error

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,122 @@
...
+			output = append(output, data)
+		}
+	}
+	severityOrder := findingsModel.Summary.SeverityOrderAsc
+	filter := filterFindingsBySeverity(config.GetString(configuration.FLAG_SEVERITY_THRESHOLD), severityOrder)
+	filteredFindings := filterFindings(findingsModel, []FindingsFilterFunc{filter})
+	filteredFindingsBytes, err := json.Marshal(filteredFindings)
+	if err != nil {
+		logger.Err(err).Msg(err.Error())
+		return output, err
```
j-luong, line 114: question: did we mean to return an `output, err` rather than `nil err`?

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	output = append(output, workflow.NewData(
```
j-luong, line 116: question: if the intention is to return a filtered list of findings, do we want to append to the existing `output`?

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,122 @@
...
+type FindingsFilterFunc func(local_models.FindingResource) bool
+
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
```
j-luong, line 38: suggestion: consider moving to the `utils` package, there's a duplicate method in `presenters_local_findings`

thisislawatts:
@sandor-trombitas is this ready for review, I see that in the ticket we state as definition of done. 
> Update existing local findings presenter flow to use the new filtering workflow

Or will this be tackled as a follow-up item.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+			filteredFindings.Findings = append(filteredFindings.Findings, finding)
+		}
+	}
+	return filteredFindings
+}
+
+func filterFindingsEntryPoint(invocationCtx workflow.InvocationContext, input []workflow.Data) (output []workflow.Data, err error) {
+	config := invocationCtx.GetConfiguration()
+	logger := invocationCtx.GetEnhancedLogger()
+	ff_transform_enabled := config.GetBool(configuration.FF_TRANSFORMATION_WORKFLOW)
```
j-luong, line 73: nitpick: let's stick to snakeCase for consistency

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+
+		return utils.Contains(allowed_severities, string(finding.Attributes.Rating.Severity.Value))
+	}
+}
+
+func filterFindings(findingsModel local_models.LocalFinding, filter []FindingsFilterFunc) local_models.LocalFinding {
+	filteredFindings := local_models.LocalFinding{}
+	for _, finding := range findingsModel.Findings {
+		match := true
+		for _, filter := range filter {
```
PeterSchafer, line 58: Suggestion: remove this shadow declaration.

```suggestion
		for _, filter := range listOfFilter {
```

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+			if match {
```
PeterSchafer, line 59: Suggestion: This logic seems like it can be improved a bit in order to be more clear and optimize performance. 

I think what we are looking to achieve is for each finding to see if any filter matches. With this assumption, we can just generally stop looping through filters if the first match is found.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+func filterFindingsBySeverity(severityThreshold string, severityOrder []string) FindingsFilterFunc {
+	return func(finding local_models.FindingResource) bool {
+		allowed_severities := filterSeverityASC(severityOrder, severityThreshold)
+
+		return utils.Contains(allowed_severities, string(finding.Attributes.Rating.Severity.Value))
+	}
+}
+
+func filterFindings(findingsModel local_models.LocalFinding, filter []FindingsFilterFunc) local_models.LocalFinding {
+	filteredFindings := local_models.LocalFinding{}
```
PeterSchafer, line 55: Issue: `filteredFindings` is a new instance of `local_models.LocalFinding` this means all other fields from the original findingsModel are dropped, for example the summary ...

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+	minLevelPointer := slices.Index(original, severityMinLevel)
+
+	if minLevelPointer >= 0 {
+		return original[minLevelPointer:]
+	}
+
+	return original
+}
+
+func filterFindingsBySeverity(severityThreshold string, severityOrder []string) FindingsFilterFunc {
```
PeterSchafer, line 46: Suggestion: I know naming is hard, but I stumbled across the name as it doesn't actually filter something like the name suggests but it returns a filter function to be used for filtering. I would suggest naming it something like `getSeverityThresholdFilter()`

pkg/local_workflows/filter_workflow.go
PeterSchafer: Yes please ❤️

PeterSchafer: Question: actually do we really need to do this? Isn't it enough to search for finding models and filter them i available? This seems in general an unnecessary dependency between the different workflows.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+		} else {
+			output = append(output, data)
+		}
+	}
+	if len(findingsModels) == 0 {
+		return output, nil
+	}
+	for _, findingsModel := range findingsModels {
+		severityOrder := findingsModel.Summary.SeverityOrderAsc
+		filter := filterFindingsBySeverity(config.GetString(configuration.FLAG_SEVERITY_THRESHOLD), severityOrder)
```
PeterSchafer, line 119: Suggestion: If the severity-threshold is not set, there will be a filter with all severities. This seems unnecessary effort. What I want to say is, we don't need to filter of no filter parameter is specified.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+			findingsBytes, ok := data.GetPayload().([]byte)
+			if !ok {
+				var findingsError snyk_errors.Error = snyk_errors.Error{
+					Title:          "Invalid Payload Type",
+					Classification: "Internal",
+					Level:          "warning",
+					Detail:         "Invalid payload type",
+				}
+				logger.Err(findingsError).Msg(findingsError.Error())
+				return nil, findingsError
```
PeterSchafer, line 96: Question: Do we really want to stop the data flow if this workflow can't process all the data? I know this goes in the direction of panic vs not. But I feel there is a bit of uncertainty here as the error we create says "warning", but effectively we return nil so we swallow all data. And the message "Invalid Payload Type" is not helping to resolve the issue easily if it at all happens.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,139 @@
...
+				var unmarshallError snyk_errors.Error = snyk_errors.Error{
+					Title:          "Failed to unmarshall findings",
+					Classification: "Internal",
+					Level:          "warning",
+					Detail:         err.Error(),
+				}
+				logger.Err(unmarshallError).Msg(unmarshallError.Error())
+				return nil, unmarshallError
+			}
+			findingsModels = append(findingsModels, findingsModel)
```
PeterSchafer, line 109: nitpick: `readability` It is more difficult to follow the control flow, if we put decoded findings in a list, filter them later and put them in the output again. Did you consider just having a single function to filter a single findings model and put it back in the output immediately? This would more follow IOSP.

sandor-trombitas:
> @sandor-trombitas is this ready for review, I see that in the ticket we state as definition of done.
> 
> > Update existing local findings presenter flow to use the new filtering workflow
> 
> Or will this be tackled as a follow-up PR?

Cleanup has already been merged in as part of a previous PR

pkg/local_workflows/filter_workflow.go
sandor-trombitas: thank you for catching that

pkg/local_workflows/filter_workflow.go
sandor-trombitas: Not quite, we append a finding if it matches all the filters.
e.g. we can have two filters, `isBelowThreshold` and `isNotIgnored`, and to show the finding, it should match both

But yes, a small optimization can be made on the else branch 👍

pkg/local_workflows/filter_workflow.go
sandor-trombitas: has been removed, thank you

pkg/local_workflows/filter_workflow.go
PeterSchafer: Oh, this intention wasn't clear to me. Can we please write this in a comment for `applyFilter` that mentions that multiple filters are ANDed. That would be really helpful.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,152 @@
...
+	return func(finding local_models.FindingResource) bool {
+		allowed_severities := filterSeverityASC(severityOrder, severityThreshold)
+
+		return utils.Contains(allowed_severities, string(finding.Attributes.Rating.Severity.Value))
+	}
+}
+
+func applyFilters(findingsModel *local_models.LocalFinding, filters []FindingsFilterFunc) {
+	// filteredFindings := local_models.LocalFinding{}
+	filteredFindings := []local_models.FindingResource{}
```
PeterSchafer, line 56: Issue: I'm just realizing that we need to filter the summary as well, not just only the findings. 
This means, the summary contains ignored and open issues by severity, if we apply a severity filter, it needs to adapt both (findings and summary). 

For the sarif based flow, this is already done in the [output workflow](https://github.com/snyk/go-application-framework/blob/main/pkg/local_workflows/output_workflow.go#L50).

PeterSchafer: I think this is related to the bug ticket CLI-559 that I created.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+type FindingsFilterFunc func(local_models.FindingResource) bool
+
+func filterSeverityASC(original []string, severityMinLevel string) []string {
+	if severityMinLevel == "" {
+		return original
+	}
+
+	minLevelPointer := slices.Index(original, severityMinLevel)
```
PeterSchafer, line 37: Suggestion: currently the implementation is relying on the correct writing, consider converting `severityMinLevel` to lower case

pkg/local_workflows/filter_workflow.go
PeterSchafer: I think the solution is relatively easy, as it would just require to change the filter interface a bit to accept the complete findings model. Iterating through the different data fields can than be done in the filter.
Positive side effect, this filter interface is even more powerful and suitable for other filter types as well.

pkg/local_workflows/filter_workflow.go
PeterSchafer: I know, I'm late to the game but there is also a [stdlib function](https://pkg.go.dev/slices#Contains) for this. `slices.Contains()`. Reducing our code by using stdlib functions is always a good idea.

pkg/local_workflows/filter_workflow.go
PeterSchafer: You might want to do this [here](https://github.com/snyk/go-application-framework/pull/267/files#diff-e047f4d01ea83c84174b4142785d23ac66b49c2d8ca040b60f3bb4fa85a4c29bR78).

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, data := range input {
+		if strings.HasPrefix(data.GetContentType(), content_type.LOCAL_FINDING_MODEL) {
+			var findingsModel local_models.LocalFinding
+			findingsBytes, ok := data.GetPayload().([]byte)
+			if !ok {
+				var findingsError snyk_errors.Error = snyk_errors.Error{
+					Title:          "Invalid Payload Type",
+					Classification: "Internal",
+					Level:          "warning",
+					Detail:         "Invalid payload type",
```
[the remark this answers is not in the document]
PeterSchafer: Something like: `Failed to apply filter due to an unexpected payload type.`

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+				logger.Err(findingsError).Msg(findingsError.Error())
```
PeterSchafer, line 98: Suggestion: make the message more meaningful, the data object has information like the content location or the identifier, having this information in the logs can help trace the data through the system

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,152 @@
...
+				output = append(output, data)
+				continue
+			}
+			err := json.Unmarshal(findingsBytes, &findingsModel)
+			if err != nil {
+				var unmarshallError snyk_errors.Error = snyk_errors.Error{
```
PeterSchafer, line 104: Suggestion: Use https://pkg.go.dev/fmt#Errorf to wrap errors, this enables to keep the original error alongside any higher level error.

pkg/local_workflows/filter_workflow.go
```diff
@@ -0,0 +1,152 @@
...
+					Title:          "Invalid Severity Threshold",
+					Classification: "Internal",
+					Level:          "warning",
+					Detail:         "Severity threshold is not valid",
+				}
+				logger.Err(severityError).Msg(severityError.Error())
+				output = append(output, data)
+				continue
+			}
+			isBelowThreshold := getSeverityThresholdFilter(severityThreshold, severityOrder)
```
PeterSchafer, line 126: nitpick: I think you forgot to rename this variable

pkg/local_workflows/filter_workflow_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func setupMockFilterContext(t *testing.T, fflagEnabled bool, severityThreshold string) *mocks.MockInvocationContext {
+	t.Helper()
+
+	// setup
+	logger := zerolog.Logger{}
+	config := configuration.New()
+	config.Set(configuration.FF_TRANSFORMATION_WORKFLOW, fflagEnabled)
```
PeterSchafer, line 24: Suggestion: I think this is not required anymore

pkg/local_workflows/filter_workflow_test.go
```diff
@@ -0,0 +1,90 @@
...
+			findingsBytes,
+		)}
+		output, err := filterFindingsEntryPoint(ctx, input)
+		assert.NoError(t, err)
+		assert.Equal(t, 1, len(output))
+		var filteredFindings local_models.LocalFinding
+		err = json.Unmarshal(output[0].GetPayload().([]byte), &filteredFindings)
+
+		assert.NoError(t, err)
+		for _, finding := range filteredFindings.Findings {
```
PeterSchafer, line 85: Suggestion: You may also want to ensure that the other data in the findings model is correct.

pkg/local_workflows/filter_workflow.go
sandor-trombitas: we'll do this as a separate PR as part of CLI-559

thisislawatts (approved):

PR #271, opened 2024-10-30
chore: use rwlocks in configuration

PeterSchafer:
This PR extends the locking in the Configuration and improves it by using an RWMutex.


thisislawatts:
@PeterSchafer are there tests we can introduce which verify the desired behaviour which is enabled through the use of RWMutex are available? I am guessing that would be multiple goroutines can hold read locks simultaneously.

thisislawatts (approved):

PR #272, opened 2024-10-31
fix: fix thread safety issue in engine invoke

PeterSchafer:


bastiandoetsch (approved):

thisislawatts:
**issue(blocking):** Similar to my comment https://github.com/snyk/go-application-framework/pull/271#issuecomment-2449705882, can we introduce a test which captures the desired behaviour before merging this item.

thisislawatts (dismissed):
**issue(blocking):** Similar to my comment https://github.com/snyk/go-application-framework/pull/271#issuecomment-2449705882, can we introduce a test which captures the desired behaviour before merging this item.

bastiandoetsch:
This is stalled. Are we planning on continuing this PR?

PeterSchafer:
@bastiandoetsch I added a test and adapted the critical section a bit. Could you please re-review? I think it would be good to get this merged.

pkg/workflow/engineimpl.go
```diff
@@ -247,21 +251,25 @@ func (e *EngineImpl) InvokeWithInputAndConfig(
 
 			// prepare logger
 			prefix := fmt.Sprintf("%s:%d", id.Host, e.invocationCounter)
-			e.mu.Unlock()
```
bastiandoetsch, line 250: I think, we'd in general only need a read-lock (and an RWMutex). Nothing blocking though. Should the lock maybe even start earlier, e.g. in L240 to avoid init races?

pkg/workflow/engineimpl.go
```diff
@@ -247,21 +251,25 @@ func (e *EngineImpl) InvokeWithInputAndConfig(
...
-
 			zlogger := e.logger.With().Str("ext", prefix).Logger()
 
+			localConfig := e.config
+			localNetwork := e.networkAccess
```
bastiandoetsch, line 257: I'd find it more intuitive, to have `localNetwork` never contain `e.networkAccess`, but the clone directly on assignment - that way it is always `local`. Same for config.

pkg/workflow/engineimpl.go
```diff
@@ -247,21 +251,25 @@ func (e *EngineImpl) InvokeWithInputAndConfig(
...
 
 			// prepare networkAccess
-			networkAccess := e.networkAccess.Clone()
+			networkAccess := localNetwork.Clone()
 			networkAccess.SetConfiguration(config)
+			e.mu.Unlock()
 
 			// create a context object for the invocation
-			context := NewInvocationContext(id, config, e, networkAccess, zlogger, e.analytics, e.ui)
+			context := NewInvocationContext(id, config, e, networkAccess, zlogger, localAnalytics, localUi)
```
bastiandoetsch, line 272: why a new network access, if you already have the `localNetwork`? Confused me briefly, why `networkAccess` is passed instead of `localNetwork`

pkg/workflow/engineimpl.go
```diff
@@ -247,21 +251,25 @@ func (e *EngineImpl) InvokeWithInputAndConfig(
...
 			if config == nil {
-				config = e.config.Clone()
+				config = localConfig.Clone()
 			}
 
 			// prepare networkAccess
-			networkAccess := e.networkAccess.Clone()
+			networkAccess := localNetwork.Clone()
 			networkAccess.SetConfiguration(config)
+			e.mu.Unlock()
```
bastiandoetsch, line 269: do we need the additional networkAccess variable?

bastiandoetsch (approved):

PR #276, opened 2024-11-08
feat: render description for Error Catalog errors

CatalinSnyk:
- Updates the dependency for the Error Catalog.
- Includes the description field in the RenderError output for errors from the Error Catalog, shortening the field to the first 2 paragraphs in case of longer descriptions. 
- Changed the label for the HTTP status and added the status code text next to the value.
- Changed the label for the links and added a default link for the error catalog page when the error code is present.
- Renders the title without the parentheses in case of no defined error code.

Ticket: [CLI-551](https://snyksec.atlassian.net/browse/CLI-551)

[CLI-551]: https://snyksec.atlassian.net/browse/CLI-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -215,6 +215,8 @@ github.com/snyk/code-client-go v1.8.0 h1:6H883KAn7ybiSIxhvL2QR9yEyHgAwA2+9WVHMDN
 github.com/snyk/code-client-go v1.8.0/go.mod h1:orU911flV1kJQOlxxx0InUQkAfpBrcERsb2olfnlI8s=
 github.com/snyk/error-catalog-golang-public v0.0.0-20240809094525-c48d19c27edb h1:w9tJhpTFxWqAhLeraGsMExDjGK9x5Dwj1NRFwb+t+QE=
 github.com/snyk/error-catalog-golang-public v0.0.0-20240809094525-c48d19c27edb/go.mod h1:Ytttq7Pw4vOCu9NtRQaOeDU2dhBYUyNBe6kX4+nIIQ4=
+github.com/snyk/error-catalog-golang-public v0.0.0-20241030160523-0aa643bb7069 h1:Oj/BJAEMEuBjTAQ72UYB4tR0IZKOB2ZtdDnAnJDL1BM=
+github.com/snyk/error-catalog-golang-public v0.0.0-20241030160523-0aa643bb7069/go.mod h1:Ytttq7Pw4vOCu9NtRQaOeDU2dhBYUyNBe6kX4+nIIQ4=
 github.com/snyk/go-httpauth v0.0.0-20231117135515-eb445fea7530 h1:s9PHNkL6ueYRiAKNfd8OVxlUOqU3qY0VDbgCD1f6WQY=
 github.com/snyk/go-httpauth v0.0.0-20231117135515-eb445fea7530/go.mod h1:88KbbvGYlmLgee4OcQ19yr0bNpXpOr2kciOthaSzCAg=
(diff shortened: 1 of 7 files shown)


go.sum
```diff
@@ -215,6 +215,8 @@ github.com/snyk/code-client-go v1.8.0 h1:6H883KAn7ybiSIxhvL2QR9yEyHgAwA2+9WVHMDN
 github.com/snyk/code-client-go v1.8.0/go.mod h1:orU911flV1kJQOlxxx0InUQkAfpBrcERsb2olfnlI8s=
 github.com/snyk/error-catalog-golang-public v0.0.0-20240809094525-c48d19c27edb h1:w9tJhpTFxWqAhLeraGsMExDjGK9x5Dwj1NRFwb+t+QE=
 github.com/snyk/error-catalog-golang-public v0.0.0-20240809094525-c48d19c27edb/go.mod h1:Ytttq7Pw4vOCu9NtRQaOeDU2dhBYUyNBe6kX4+nIIQ4=
+github.com/snyk/error-catalog-golang-public v0.0.0-20241030160523-0aa643bb7069 h1:Oj/BJAEMEuBjTAQ72UYB4tR0IZKOB2ZtdDnAnJDL1BM=
```
PeterSchafer, line 218: Suggestion: You have to run go mod tidy.

internal/presenters/components.go
```diff
@@ -47,19 +49,45 @@ func RenderError(err snyk_errors.Error) string {
...
+			lines = lines[0:2]
+			for i, l := range lines {
+				lines[i] = strings.Trim(l, " \n")
+			}
+			desc = strings.Join(lines, " ")
+		}
+
+		body = append(body, lipgloss.JoinHorizontal(lipgloss.Top,
+			label.Render("Details:"),
+			value.Copy().Width(80).Render(desc),
```
PeterSchafer, line 72: Suggestion: Let's move the hard-coded 80 into a constant and re-use it.

PeterSchafer:
Looks good to me! Please take a look at the small comments.

Generally I think this task could have been a great possibility to make use of a template, just to get more used to them. But definitely a scope creep.

internal/presenters/components.go
```diff
@@ -47,19 +49,45 @@ func RenderError(err snyk_errors.Error) string {
...
+		))
+	}
+
+	title := strings.TrimSpace(err.Title)
+	if len(err.ErrorCode) > 0 {
+		link := "https://docs.snyk.io/scan-with-snyk/error-catalog#" + strings.ToLower(err.ErrorCode)
```
PeterSchafer, line 78: Suggestion: let's move the url to a constant, it might not change but it might be good to have it re-usable.

PeterSchafer: maybe even two constants, one for the docs base URL and one for the path

sandor-trombitas:
> I tested this locally but it looked to work without issues, but maybe you guys know more about the module system in Go. I was wondering if accessing the Description field would perhaps cause an issues for errors coming from an older version of the Error Catalog.
> 
> In out case specifically for the CLI, this could be a workflow that didn't update the Error Catalog dependency to the newer version that has the Description field. This just popped in my head yesterday, and from my testing the error catalog version would get overridden by the newer one.

Nothing to worry about, gaf uses its own Error Catalog, so the CLI's version won't impact accessing the Description field.

sandor-trombitas (approved):
looks good 🚀

PR #279, opened 2024-11-15
feat: handle non 2xx responses in GAF network stack

CatalinSnyk:
- Adds a new method to the NetworkAccess interface for registering an error handler
- Implements the handler for 401 unauthorized status code

Ticket: [CLI-575](https://snyksec.atlassian.net/browse/CLI-575)

[CLI-575]: https://snyksec.atlassian.net/browse/CLI-575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/utils/convert.go
+++ b/internal/utils/convert.go
@@ -1,0 +1,24 @@
+package utils
+
+import (
+	"strconv"
+)
+
+func ToBool(value interface{}) bool {
+	if value == nil {
+		return false
+	}
+
+	switch v := value.(type) {
+	case bool:
+		return v
+	case string:
+		boolResult, err := strconv.ParseBool(v)
+		if err != nil {
+			return false
+		}
+		return boolResult
+	}
+
+	return false
+}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -27,25 +27,25 @@ import (
 )
 
 func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger, apiClientFactory func(url string, client *http.Client) api.ApiClient) configuration.DefaultValueFunction {
-	callback := func(existingValue interface{}) interface{} {
+	callback := func(existingValue interface{}) (interface{}, error) {
 		client := engine.GetNetworkAccess().GetHttpClient()
 		url := config.GetString(configuration.API_URL)
 		apiClient := apiClientFactory(url, client)
 		orgId := config.GetString(configuration.ORGANIZATION)
 		if len(orgId) == 0 {
-			return existingValue
+			return existingValue, nil
 		}
 		slugName, err := apiClient.GetSlugFromOrgId(orgId)
 		if err != nil {
 			logger.Print("Failed to determine default value for \"ORGANIZATION_SLUG\":", err)
 		}
-		return slugName
+		return slugName, nil
 	}
 	return callback
 }
 
 func defaultFuncOrganization(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger, apiClientFactory func(url string, client *http.Client) api.ApiClient) configuration.DefaultValueFunction {
-	callback := func(existingValue interface{}) interface{} {
+	callback := func(existingValue interface{}) (interface{}, error) {
 		client := engine.GetNetworkAccess().GetHttpClient()
 		url := config.GetString(configuration.API_URL)
 		apiClient := apiClientFactory(url, client)
@@ -59,10 +59,10 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 				if err != nil {
 					logger.Print("Failed to determine default value for \"ORGANIZATION\":", err)
 				} else {
-					return orgId
+					return orgId, nil
 				}
 			} else {
-				return orgId
+				return orgId, nil
 			}
 		}
 
@@ -71,13 +71,13 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 			logger.Print("Failed to determine default value for \"ORGANIZATION\":", err)
 		}
 
-		return orgId
+		return orgId, nil
 	}
 	return callback
 }
 
 func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logger) configuration.DefaultValueFunction {
-	callback := func(existingValue interface{}) interface{} {
+	callback := func(existingValue interface{}) (interface{}, error) {
 		urlString := constants.SNYK_DEFAULT_API_URL
 
 		urlFromOauthToken, err := auth.GetAudienceClaimFromOauthToken(config.GetString(auth.CONFIG_KEY_OAUTH_TOKEN))
@@ -97,28 +97,28 @@ func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logge
 		if err != nil {
 			logger.Warn().Err(err).Str(configuration.API_URL, urlString).Msg("failed to get api url")
 		}
-		return apiString
+		return apiString, nil
 	}
 	return callback
 }
 
 func defaultInputDirectory() configuration.DefaultValueFunction {
-	callback := func(existingValue interface{}) interface{} {
+	callback := func(existingValue interface{}) (interface{}, error) {
 		if existingValue == nil {
 			path, err := os.Getwd()
 			if err != nil {
-				return ""
+				return "", err
 			}
-			return path
+			return path, nil
 		} else {
-			return existingValue
+			return existingValue, nil
 		}
 	}
 	return callback
 }
 
 func defaultTempDirectory(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger) configuration.DefaultValueFunction {
-	callback := func(existingValue interface{}) interface{} {
+	callback := func(existingValue interface{}) (interface{}, error) {
 		version := "0.0.0"
 		ri := engine.GetRuntimeInfo()
 		if ri != nil && len(ri.GetVersion()) > 0 {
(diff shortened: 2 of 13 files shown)


pkg/networking/networking.go
```diff
@@ -121,9 +126,27 @@ func (rt *defaultHeadersRoundTripper) RoundTrip(request *http.Request) (*http.Re
 
 	rt.logRoundTrip(newRequest, response, err)
 
+	if err != nil {
+		return nil, err
+	}
+
+	if rt.networkAccess.errorHandler != nil {
+		errFromStatus := errFromStatusCode(response.StatusCode)
```
PeterSchafer, line 134: Suggestion: How about `errFromResponse()` this would enable to parse jsonapi errors into error catalog errors.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// RoundTrip should return one or the other.
+	if err != nil {
+		res = nil
+	}
+
+	return res, err
+}
+
+type jsonAPIErroResponse struct {
```
PeterSchafer, line 94: Suggestion: The logic and the data structures to decode error catalog errors should be living in error catalog.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+	Code   string `json:"code,omitempty"`
+}
+
+func (jar *jsonAPIErroResponse) ToErrorCatalog() error {
+	if len(jar.Errors) == 0 {
+		return nil
+	}
+
+	jsonErr := jar.Errors[0]
+	fmt.Println(jsonErr)
```
PeterSchafer, line 111: Issue: watch out for pushing `fmt.Println()`!

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+func errFromResponse(res *http.Response) error {
+	if res.StatusCode >= 200 && res.StatusCode < 300 {
+		return nil
+	}
+
+	if err := errFromStatusCode(res.StatusCode); err != nil {
+		return err
+	}
+
+	if err := errFromJSONAPI(res); err != nil {
```
PeterSchafer, line 35: Suggestion: Let's switch the order to prefer errors received from the backend over generic errors derived from status codes.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return snykErr.ToErrorCatalog()
+}
+
+func (rm ResponseMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
+	res, err := rm.next.RoundTrip(req)
+
+	if err != nil {
+		return nil, err
```
PeterSchafer, line 77: Suggestion: Return res as well, so that you don't change the behaviour.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+
+func NewReponseMiddleware(roundTriper http.RoundTripper, errHandler func(error, context.Context) error) *ResponseMiddleware {
+	return &ResponseMiddleware{
+		next:       roundTriper,
+		errHandler: errHandler,
+	}
+}
+
+func errFromResponse(res *http.Response) error {
+	if res.StatusCode >= 200 && res.StatusCode < 300 {
```
PeterSchafer, line 27: Suggestion: use constants in http package

PeterSchafer: Actually you might be able to remove this logic at all, if you catch this cases in the code below.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if err := errFromStatusCode(res.StatusCode); err != nil {
+		return err
+	}
+
+	if err := errFromJSONAPI(res); err != nil {
+		return err
+	}
+
+	return snyk_errors.Error{
```
PeterSchafer, line 39: Suggestion: catching unhandled cases as an error seems like a very strong approach. I think we might want to be less strong, as we don't want to change the default behaviour too much.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,123 @@
...
+		Description: "A network request failed because with the resulting status code.",
+		StatusCode:  res.StatusCode,
+		Detail:      "Use the `-d` flag to inspect the request and response.",
+		Level:       "error",
+	}
+}
+
+func errFromStatusCode(code int) error {
+	switch code {
+	case http.StatusUnauthorized:
```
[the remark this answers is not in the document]
PeterSchafer: There are other ways as well, like using resp.Request.Header to determine if there is any auth material, so that we can give more helpful errors.

pkg/networking/middleware/response.go
CatalinSnyk: I looked and found out the logic for this is [included](https://github.com/snyk/error-catalog/blob/2d9a7fd6f71d7eddb940a4d00e6bd9269348b864/packages/error-catalog-golang/src/snyk_errors/json_api_error.go#L20) in the Error Catalog, but for some reason it's private. I could raise a PR to make this public.

pkg/app/app.go
```diff
@@ -25,25 +25,25 @@ import (
...
 		url := config.GetString(configuration.API_URL)
 		apiClient := apiClientFactory(url, client)
 		orgId := config.GetString(configuration.ORGANIZATION)
 		if len(orgId) == 0 {
-			return existingValue
+			return existingValue, nil
 		}
 		slugName, err := apiClient.GetSlugFromOrgId(orgId)
 		if err != nil {
 			logger.Print("Failed to determine default value for \"ORGANIZATION_SLUG\":", err)
```
CatalinSnyk, line 38: Just a note, I tried to keep most of the returns untouched as not to alter behaviour, and I could always come back and add the error to the return if necessary. The returns that changed were because of the linter.

internal/utils/convert.go
```diff
@@ -0,0 +1,24 @@
+package utils
+
+import (
+	"strconv"
+)
+
+func ToBool(value interface{}) bool {
```
CatalinSnyk, line 7: I couldn't find a place for this kind of functions, but I could move all of the logic from the other getters as well to keep things clean. Just wanted to confirm this if this place is ok for them.

pkg/configuration/configuration.go
```diff
@@ -375,12 +377,28 @@ func (ev *extendedViper) Get(key string) interface{} {
...
 
+// GetE returns a configuration value and and error if the value is not set.
+func (ev *extendedViper) GetE(key string) (value interface{}, err error) {
+	value = ev.get(key)
+
+	if ev.defaultValues[key] != nil {
+		value, err = ev.defaultValues[key](value)
+	}
+
+	if value == nil && err == nil {
```
PeterSchafer, line 395: Suggestion: I think this is actually a valid case and not an error

pkg/networking/networking.go
```diff
@@ -55,6 +60,7 @@ type networkImpl struct {
 	staticHeader   http.Header
 	dynamicHeaders map[string]DynamicHeaderFunc
 	proxy          func(req *http.Request) (*url.URL, error)
+	errorHandler   func(error, context.Context) error
```
PeterSchafer, line 63: Suggestion: Let's introduce a type for the error handler func

PeterSchafer: Issue: please consider copying the error handler in the clone function.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,101 @@
...
+func HandleResponse(res *http.Response, config configuration.Configuration) error {
+	if res == nil {
+		return nil
+	}
+
+	apiUrl := config.GetString(configuration.API_URL)
+	additionalSubdomains := config.GetStringSlice(configuration.AUTHENTICATION_SUBDOMAINS)
+	additionalUrls := config.GetStringSlice(configuration.AUTHENTICATION_ADDITIONAL_URLS)
+	//nolint:errcheck // discarded error since it's returned only with the value of false
+	requiresAuth, _ := ShouldRequireAuthentication(apiUrl, res.Request.URL, additionalSubdomains, additionalUrls)
```
PeterSchafer, line 58: Suggestion: Even with your comment, I think it is worth to use the error in the next condition. The reason is, that what you describe is the current behaviour, but there might be changes along the way that change this.

pkg/networking/middleware/response.go
```diff
@@ -0,0 +1,101 @@
...
+func (rm ResponseMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
+	res, err := rm.next.RoundTrip(req)
+
+	if err != nil {
+		return res, err
+	}
+
+	err = HandleResponse(res, rm.config)
+
+	if rm.errHandler != nil {
```
PeterSchafer, line 36: Suggestion: To be more clear on what happens, I think we should remove the check if errHandler is nil, since this middleware is [only used](https://github.com/snyk/go-application-framework/pull/279/files#diff-99beec8232d7e64bc806ee6a0d9ee88bbff437a1a93094027686253ab80725a5R216) when an error handler is available.

pkg/networking/middleware/response_test.go
```diff
@@ -0,0 +1,133 @@
...
+	req.Header.Set("snyk-request-id", "1234")
+	if err != nil {
+		panic(err)
+	}
+
+	return req
+}
+
+func getBaseConfig() configuration.Configuration {
+	config := configuration.NewWithOpts(configuration.WithAutomaticEnv())
```
PeterSchafer, line 130: Suggestion: I think this test doesn't require environment variables, so you should be good to remove `configuration.WithAutomaticEnv()`

PeterSchafer:
Looks really great, Thank you! I only left some minor comments.

PeterSchafer (approved):

PR #282, opened 2024-11-27
feat: support URL to Region Alias

thisislawatts:
[CLI-568]

[CLI-568]: https://snyksec.atlassian.net/browse/CLI-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/local_workflows/config_workflow.go
+++ b/pkg/local_workflows/config_workflow.go
@@ -35,6 +35,35 @@ func InitConfigWorkflow(engine workflow.Engine) error {
 	return err
 }
 
+type SnykRegion struct {
+	alias string
+	url   string
+}
+
+func DetermineRegionFromUrl(url string) (string, error) {
+	if len(url) == 0 {
+		return "", fmt.Errorf("url must not be empty")
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/local_workflows/config_workflow.go
```diff
@@ -35,6 +44,27 @@ func InitConfigWorkflow(engine workflow.Engine) error {
 	return err
 }
 
+type SnykRegion struct {
+	alias string
+	url   string
+}
```
j-luong, line 50: question: what's the reason for changing from a map to a struct?

thisislawatts: How would the map help with that?

j-luong: nvm, I think I convinced myself out of it. It may actually complicate things if the map has to contain both the alias and url as keys to be searchable.

j-luong (approved):

PR #290, opened 2024-12-06
fix: include debug shorthand flag

thisislawatts:
This pull request enhances the accuracy of analytics by ensuring correct categorization of CLI commands, even when using shorthand flags.

### Key Changes

- Shorthand Flag Recognition: The determineCategoryFromArgs function is updated to recognize the -d shorthand flag as equivalent to the --debug flag.
- Test Case Updates: Test cases are modified to validate the correct categorization of commands with shorthand flags.
- Constant Introduction: A new constant DEBUG_FLAG_SHORTHAND is added for clarity and maintainability.
- Flag Registration: The GetGlobalConfiguration function registers the -d shorthand flag for the DEBUG configuration option.

By accurately capturing command-line arguments, including shorthand flags, this change improves the reliability and precision of analytics data.

### DIFF
--- a/pkg/instrumentation/categories.go
+++ b/pkg/instrumentation/categories.go
@@ -8,6 +8,8 @@ import (
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
+const dEBUG_FLAG_SHORTHAND string = "d"
+
 // DetermineCategoryFromArgs categorizes command-line arguments into a structured slice.
 //
 // This function receives the entire os.Args slice (including the program name at index 0)
@@ -69,7 +71,7 @@ func determineCategoryFromArgs(args []string, knownCommands []string, flagsAllow
 			}
 
 			// a shorthand variant for debug
-			if flagName == configuration.DEBUG_FLAG_SHORTHAND {
+			if flagName == dEBUG_FLAG_SHORTHAND {
 				flags = append(flags, configuration.DEBUG)
 			}
 		} else if slices.Contains(knownCommands, arg) {
--- a/pkg/instrumentation/categories.go
+++ b/pkg/instrumentation/categories.go
@@ -4,6 +4,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/instrumentation/categories.go
```diff
@@ -66,6 +67,11 @@ func determineCategoryFromArgs(args []string, knownCommands []string, flagsAllow
 			if slices.Contains(flagsAllowList, flagName) {
 				flags = append(flags, strings.ToLower(flagName))
 			}
+
+			// a shorthand variant for debug
+			if flagName == configuration.DEBUG_FLAG_SHORTHAND {
```
PeterSchafer, line 72: Suggestion: While this solution solves the missing support for the `-d` short hand flag, there is a general missing support for shorthand flags and I think we can adapt the implementation slightly to make this work. Happy to hop on a call for this, otherwise I can push a commit here with my proposal.

thisislawatts: @PeterSchafer feel free to push your proposal into this branch :)

thisislawatts: I opted for this lightweight approach as this is the only shorthand flag we are using at the moment.

PeterSchafer: Makes sense! But I think there are actually more shorthand flags in the wild than we expect.

pkg/instrumentation/categories.go
```diff
@@ -66,6 +67,11 @@ func determineCategoryFromArgs(args []string, knownCommands []string, flagsAllow
 			if slices.Contains(flagsAllowList, flagName) {
 				flags = append(flags, strings.ToLower(flagName))
 			}
+
+			// a shorthand variant for debug
```
bastiandoetsch, line 71: nitpick: do we really need this comment?

pkg/instrumentation/categories.go
PeterSchafer: I just pushed a minor change that moves the shorthand flag out of the configuration package, since the configuration package is meant to hold keys that are used to access the configuration. a shorthand flag is a cmd specific thing though.

PeterSchafer:
I think now the PR is good to merge, if we think it adds enough value. I'm a bit unsure, if we would actually want to cover short hand flags more generally, as mentioned above. But this eventually boils down to our appetite.

PeterSchafer (approved):

pkg/instrumentation/categories.go
thisislawatts: Removed

PR #298, opened 2025-01-22
chore: CLI-598 migrate transformation workflow to go native with benchmark

sandor-trombitas:
### Data Transformation Workflow Enhancements:
* Migrated `TransformToLocalFindingModel` to handle data transformation without using CUE language. This function includes detailed logic for transforming SARIF and summary bytes into a `local_models.LocalFinding` object using native go.

### Removal of cuelang
* Removed cue integration after migrating `TransformToLocalFindingModel` to a native implementation.

### Benchmarking:
* Added a benchmarking test `BenchmarkTransformationAndOutputWorkflow` to measure the performance of the native transformations. This test logs memory usage and duration.

Initial benchmark results for both cue and native can be found below for 10000 findings.
Benchmark was run on a m3 pro macbook pro with 32gb of ram
| Implementation | Memory | Duration |
|--------|--------|--------|
| Cue | 47212 MBl | 1m10s |
| Native | 590 MB | 1.1s | 

These changes aim to improve the performance of the data transformation workflow by replacing the CUE-based transformation with a native implementation ensuring 1:1 compatibility and benchmarking.


pkg/local_workflows/output_workflow_test.go
```diff
@@ -187,7 +190,12 @@ func sarifToLocalFinding(t *testing.T, filename string, projectPath string) (loc
 	summaryBytes, err := json.Marshal(summaryData)
 	assert.NoError(t, err)
 
-	tmp, err := TransformToLocalFindingModel(sarifBytes, summaryBytes)
+	t.Log("transform start", time.Now())
+
+	var tmp local_models.LocalFinding
+	tmp, err = TransformToLocalFindingModel(sarifBytes, summaryBytes)
+
+	t.Log("transform end", time.Now())
```
thisislawatts, line 198: **nit:** Clean up the logs, or add assertions around the values?

pkg/local_workflows/output_workflow_test.go
```diff
@@ -662,3 +675,61 @@ func TestLocalFindingsHandling_renderSarifFile_renderUI(t *testing.T) {
...
+	// invocation context mocks
+	invocationContextMock.EXPECT().GetConfiguration().Return(config).AnyTimes()
+	invocationContextMock.EXPECT().GetEnhancedLogger().Return(logger).AnyTimes()
+	invocationContextMock.EXPECT().GetRuntimeInfo().Return(runtimeinfo.New(runtimeinfo.WithName("SnykCode"), runtimeinfo.WithVersion("1.0.0"))).AnyTimes()
+	outputDestination.EXPECT().GetWriter().Return(writer).AnyTimes()
+
+	config.Set(output_workflow.OUTPUT_CONFIG_KEY_SARIF, true)
+	testfile := "testdata/10000Findings.json"
+
+	b.Run("native", func(b *testing.B) {
```
thisislawatts, line 699: **nit:** change the name for this now that we no longer have a cuelang implementation.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -115,56 +114,330 @@ func TransformToLocalFindingModel(sarifBytes []byte, summaryBytes []byte) (local
 		return localFinding, err
 	}
 
-	input, errUnJson := cuejson.Unmarshal(sarifBytes)
-	if errUnJson != nil {
-		return localFinding, fmt.Errorf("failed to unmarshal input: %w", err)
+	var sarifDoc sarif.SarifDocument
+	err = json.Unmarshal(sarifBytes, &sarifDoc)
+	if err != nil {
+		return localFinding, err
```
thisislawatts, line 120: **question:** Why have we dropped the wrapped error for passing through the err from json.Unmarshal directly.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -115,56 +114,330 @@ func TransformToLocalFindingModel(sarifBytes []byte, summaryBytes []byte) (local
...
+		for _, cov := range run.Properties.Coverage {
+			coverage = append(coverage, local_models.TypesCoverage{
+				Files:       cov.Files,
+				IsSupported: cov.IsSupported,
+				Lang:        cov.Lang,
+				Type:        cov.Type,
+			})
+		}
 	}
+	localFinding.Summary.Coverage = coverage
```
thisislawatts, line 135: **suggestion:** Move the summary computation/composition into a single function that takes testSummary and coverage as params.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -115,56 +114,330 @@ func TransformToLocalFindingModel(sarifBytes []byte, summaryBytes []byte) (local
...
-	if transformerError != nil {
-		return localFinding, transformerError
-	}
+	rules := mapRules(sarifDoc)
+	localFinding.Rules = rules
 
-	apiOutput, applyError := sarif2apiTransformer.Apply(input)
-	if applyError != nil {
-		return localFinding, applyError
+	for _, res := range sarifDoc.Runs[0].Results {
```
thisislawatts, line 140: **suggestion:** move this out to a `mapFindings` method. So that the `TransformToLocalFindingModel` becomes a shorter composition method.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -115,56 +114,330 @@ func TransformToLocalFindingModel(sarifBytes []byte, summaryBytes []byte) (local
...
+		finding.Attributes.Rating.Priority = &local_models.TypesFindingNumericalRating{
+			Score: res.Properties.PriorityScore,
+			Factors: func() (factors []local_models.RiskFactors) {
+				for _, v := range res.Properties.PriorityScoreFactors {
+					factor := &local_models.RiskFactors{}
+					err = factor.FromTypesVulnerabilityFactRiskFactor(local_models.TypesVulnerabilityFactRiskFactor{
+						Name:  v.Type,
+						Value: v.Label,
+					})
+					factors = append(factors, *factor)
```
thisislawatts, line 201: **issue:** We are missing error handling within this map.

pkg/local_workflows/output_workflow_test.go
```diff
@@ -606,7 +611,10 @@ func Test_Output_outputWorkflowEntryPoint(t *testing.T) {
 		assert.NoError(t, err)
 
 		expectedString := string(prettyExpectedSarif)
+		fmt.Println(expectedString)
```
PeterSchafer, line 614: nit: maybe remove these log lines

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
 
-	localFinding.Summary.Path = testSummary.Path
-	localFinding.Summary.Artifacts = testSummary.Artifacts
-	localFinding.Summary.Type = testSummary.Type
-	localFinding.Summary.Counts.CountKeyOrderAsc.Severity = testSummary.SeverityOrderAsc
-	localFinding.Summary.Counts.Count = 0
-	localFinding.Summary.Counts.CountAdjusted = 0
-	localFinding.Summary.Counts.CountSuppressed = 0
+func transformTestSummary(testSummary *json_schemas.TestSummary, sarifDoc *sarif.SarifDocument) *local_models.TypesFindingsSummary {
+	var summary local_models.TypesFindingsSummary
```
PeterSchafer, line 405: Suggestion: look into using the existing [NewFindingsCounts()](https://github.com/snyk/go-application-framework/blob/main/internal/utils/findings/findings.go#L33)

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
+	err = json.Unmarshal(sarifBytes, &sarifDoc)
+	if err != nil {
 		return localFinding, fmt.Errorf("failed to unmarshal input: %w", err)
 	}
 
-	ctx := cuecontext.New()
-	sarif2apiTransformer, transformerError := cueutil.NewTransformer(ctx, cueutil.ToTestApiFromSarif)
-	if transformerError != nil {
-		return localFinding, transformerError
+	localFinding.Summary = *transformTestSummary(&testSummary, &sarifDoc)
```
PeterSchafer, line 124: Suggestion: look into removing de-referencing

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
+			codeFlow.ThreadFlows = append(codeFlow.ThreadFlows, threadFlow)
+		}
+		codeFlows = append(codeFlows, codeFlow)
+	}
+	return codeFlows
+}
 
-	if encodeErr != nil {
-		return localFinding, fmt.Errorf("failed to convert to type: %w", encodeErr)
+func mapRules(sarifDoc sarif.SarifDocument) []local_models.TypesRules {
```
PeterSchafer, line 300: Suggestion: Use the sarif rules as an input parameter, this way, `mapRules` accepts a sarif rule and outputs a findingsmodel rule. It basically moves the decision of just using a single sarif rule one level up. See [IOSP](https://clean-code-developer.de/en/the-straight/red-degree/#elementor-toc__heading-anchor-5) for a reference.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
+	var rules []local_models.TypesRules
+	for _, rule := range sarifDoc.Runs[0].Tool.Driver.Rules {
+		rules = append(rules, local_models.TypesRules{
```
PeterSchafer, line 303: Suggestion: This part of the code is not easy to read as it combines declaration and initialization. what do you think of separating these concerns more by introducing `New` functions.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
-	cliOutput, applyError := api2cliTransformer.ApplyValue(apiOutput)
-	if applyError != nil {
-		return localFinding, applyError
+func mapFingerprints(res sarif.Result) []local_models.Fingerprint {
+	var fingerprints []local_models.Fingerprint
+	if res.Fingerprints.Identity != "" {
+		var fp local_models.Fingerprint
+		rawIdentity := []byte(`{"scheme":"` + string(local_models.Identity) + `","value":"` + res.Fingerprints.Identity + `"}`)
+		if err := json.Unmarshal(rawIdentity, &fp); err != nil {
+			log.Warn().Msg("Failed to unmarshal identity fingerprint")
```
PeterSchafer, line 256: Suggestion: this uses a global logger which shouldn't be used, either use parameterized logger or return errors.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
+		},
+		Justification: &suppression.Justification,
+		Kind:          "ignored",
+	}
+}
 
-	cliOutput, applyError := api2cliTransformer.ApplyValue(apiOutput)
-	if applyError != nil {
-		return localFinding, applyError
+func mapFingerprints(res sarif.Result) []local_models.Fingerprint {
```
PeterSchafer, line 250: Suggestion: There is quite some repetition in this method, I would recommend to have the logic for converting logic only once and iterate over a list of fingerprints. This will also make future extensions easier.

PeterSchafer: Suggestion: similar to a previous comment, I would recommend to use sarif fingerprints as an input and local findings fingerprints as an output.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -108,63 +108,353 @@ func dataTransformationEntryPoint(invocationCtx workflow.InvocationContext, inpu
...
+			})
+		}
 	}
+	summary.Coverage = coverage
 
-	return localFinding, nil
+	return &summary
+}
+
+func mapLocations(res sarif.Result) []local_models.IoSnykReactiveFindingLocation {
```
PeterSchafer, line 446: Suggestion: There are two places where we map PhysicalLocations, here and in `mapCodeFlows` following DRY, I would recommend to unite them. Either we also try to combine the types or use generics to handle different return types. But given that the field anems are the same and the input structure is the same, we should try to avoid this code duplication.

pkg/local_workflows/data_transformation_workflow.go
```diff
@@ -253,28 +253,32 @@ func mapSuppressions(res sarif.Result) *local_models.TypesSuppression {
 	}
 }
 
+func createFingerprint(scheme string, value string) (local_models.Fingerprint, error) {
+	var fp local_models.Fingerprint
+	raw := []byte(`{"scheme":"` + string(scheme) + `","value":"` + value + `"}`)
```
PeterSchafer, line 258: nitpick: unnecessary string() around scheme.

PeterSchafer (approved):

PR #307, opened 2025-02-13
chore: render interaction id

sandor-trombitas:


internal/presenters/components.go
```diff
@@ -40,7 +40,7 @@ func RenderError(err snyk_errors.Error) string {
 
 	level := strings.ToUpper(err.Level)
 	backgroundHighlight := errorLevelToStyle(err.Level)
-	label := lipgloss.NewStyle().Width(8)
+	label := lipgloss.NewStyle().Width(15)
```
PeterSchafer, line 43: Suggestion: maybe get design input on this change, as it changes the overall appearance of errors. Alternatively we could shorten the Label.
image

pkg/local_workflows/output_workflow.go
```diff
@@ -124,7 +124,7 @@ func outputWorkflowEntryPoint(invocation workflow.InvocationContext, input []wor
 			if err != nil {
 				return output, err
 			}
-		} else { // handle text/pain and unknown the same way
+		} else { // handle text/plain and unknown the same way
```
thisislawatts, line 127: **praise:** Good catch 🦅 👁️

internal/presenters/components.go
thisislawatts: Do we want the interaction ID bundled with the error or displayed separately as a footer? In theory a single interaction (command) could result in multiple errors being returned which would result in repetition of the Interaction Id.

thisislawatts: I see there is a thread on the PR in CLI, let me [move the conversation there](https://github.com/snyk/cli/pull/5737#discussion_r1956105436)

pkg/networking/middleware/response.go
```diff
@@ -83,8 +83,19 @@ func errFromStatusCode(code int) error {
...
+	reqPath := res.Request.URL.Path
+
+	return AddMetaDataToErr(err, map[string]any{
+		"request-id":   reqId,
+		"request-path": reqPath,
+	})
+}
+
+// AddMetaDataToErr adds the provided metadata to the Error catalog error's metadata map.
+func AddMetaDataToErr(err error, meta map[string]any) error {
```
thisislawatts, line 98: **suggestion:** We should perhaps move this to a different file that is better matches the error domain.

j-luong: is this scope creep? There's a good argument to be made for a refactor so that the entire error handling logic is moved to a different file/package.

thisislawatts: I've not added it as a blocking comment but a suggestion whilst we have the code open.

thisislawatts (approved):

internal/presenters/components.go
```diff
@@ -84,10 +78,33 @@ func RenderError(err snyk_errors.Error) string {
 		title = title + fmt.Sprintf(" (%s)", err.ErrorCode)
 	}
 
+	newln := "\n"
+	if err.StatusCode > 0 {
```
PeterSchafer, line 82: Suggestion: Do we want to use `http.StatusOK` instead of 0?

PR #310, opened 2025-02-21
fix: add line break below description/detail

thisislawatts:
Adjust how white space is added around error object. 

* Missing no status code test: Adds a case for errors without a status code, preventing display issues and ensuring the message is still shown correctly. This is particularly important for certain error types where a status code might not be available.
* Improved layout: Adjusts the layout of error details, making the information more clearly presented.


internal/presenters/__snapshots__/components_test.snap
```diff
@@ -3,7 +3,7 @@
 
 [43m [0m[30;43mWARN[0m[43m [0m   [1mService temporarily throttled (SNYK-0001)[0m
          The request rate limit has been exceeded. Wait a few minutes, then try again.  
-                               
+
```
thisislawatts, line 6: **note:* Update removes whitespace from empty line.

PeterSchafer (approved):

PR #312, opened 2025-02-25
feat: set interactionID render context instead of passing it in an error

j-luong:
## What
This PR:
- reduces the number of links shown in an error catalog err to 1 link
- allows setting of render context so that we can pass additional render items to the presenter
- fixes a high vulnerability in the go-git dependency


sandor-trombitas (approved):

pkg/ui/userinterface.go
```diff
@@ -54,13 +56,26 @@ type consoleUi struct {
 	errorWriter        io.Writer
 	progressBarFactory func() ProgressBar
 	reader             *bufio.Reader
+	//nolint:containedctx // needed to preserve non-breaking change
+	context context.Context
```
PeterSchafer, line 60: Issue: @j-luong let's remove the state here please. A possible solution is an internal structure to put this in stead of the UI instance

PeterSchafer (approved):

PR #316, opened 2025-03-19
fix: fix backward compatibility issue in sarif driver name

PeterSchafer:
Instead of using runtime information in the Sarif driver name, derive the name from the Summary type. This shall fix a change in behaviour that caused issues in downstream processing of the Sarif data, where a certain value is expected.


dotkas (approved):
You don't want to test the template generation as well?

PeterSchafer:
> You don't want to test the template generation as well?

Very good point! The existing test for the template renderer was written in a way that it had to succeed 😓 I adapted it so that it failed before the fix and now passes.

PR #319, opened 2025-03-26
feat: use gaf to get sast settings

acke:
Changes in Go Application Framework:

- added a `defaultFuncGetSastSettings` callback function that executes when the snyk-ls asks for sast settings from the Go Application Framework Configuration.
- moved sast_settings.go from "/internal/api/contract" to `/pkg/common` to be used in snyk-ls.
- introduced a new constant for the new settings in Configuration: `SAST_SETTINGS` 
- removed the cachedResponse

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -5,7 +5,7 @@ package api
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/common"
 	"io"
 	"net/http"
 	"net/url"
@@ -22,7 +22,7 @@ type ApiClient interface {
 	GetFeatureFlag(flagname string, origId string) (bool, error)
 	GetUserMe() (string, error)
 	GetSelf() (contract.SelfResponse, error)
-	GetSastSettings(orgId string) (configuration.SastResponse, error)
+	GetSastSettings(orgId string) (common.SastResponse, error)
 }
 
 var _ ApiClient = (*snykApiClient)(nil)
@@ -197,9 +197,9 @@ func (a *snykApiClient) GetSelf() (contract.SelfResponse, error) {
 	return selfData, nil
 }
 
-func (a *snykApiClient) GetSastSettings(orgId string) (configuration.SastResponse, error) {
-	var response configuration.SastResponse
-	var defaultResult configuration.SastResponse
+func (a *snykApiClient) GetSastSettings(orgId string) (common.SastResponse, error) {
+	var response common.SastResponse
+	var defaultResult common.SastResponse
 
 	endpoint := a.url + "/v1/cli-config/settings/sast?org=" + url.QueryEscape(orgId)
 	res, err := a.client.Get(endpoint)
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -5,6 +5,7 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/snyk/go-application-framework/pkg/configuration"
 	"io"
 	"net/http"
 	"net/url"
@@ -21,7 +22,7 @@ type ApiClient interface {
 	GetFeatureFlag(flagname string, origId string) (bool, error)
 	GetUserMe() (string, error)
 	GetSelf() (contract.SelfResponse, error)
-	GetSastSettings(orgId string) (contract.SastResponse, error)
+	GetSastSettings(orgId string) (configuration.SastResponse, error)
 }
 
 var _ ApiClient = (*snykApiClient)(nil)
@@ -196,9 +197,9 @@ func (a *snykApiClient) GetSelf() (contract.SelfResponse, error) {
 	return selfData, nil
 }
 
-func (a *snykApiClient) GetSastSettings(orgId string) (contract.SastResponse, error) {
-	var response contract.SastResponse
-	var defaultResult contract.SastResponse
+func (a *snykApiClient) GetSastSettings(orgId string) (configuration.SastResponse, error) {
+	var response configuration.SastResponse
+	var defaultResult configuration.SastResponse
 
 	endpoint := a.url + "/v1/cli-config/settings/sast?org=" + url.QueryEscape(orgId)
 	res, err := a.client.Get(endpoint)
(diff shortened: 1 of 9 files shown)


pkg/app/app.go
```diff
@@ -44,6 +44,28 @@ func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Co
...
+			return existingValue, nil
+		}
+		client := engine.GetNetworkAccess().GetHttpClient()
+		url := config.GetString(configuration.API_URL)
+		apiClient := apiClientFactory(url, client)
+		orgId := config.GetString(configuration.ORGANIZATION)
+		if len(orgId) == 0 {
+			return existingValue, nil
+		}
+		response, err := apiClient.GetSastSettings(orgId)
```
ShawkyZ, line 59: can't we have caching here?

pkg/local_workflows/code_workflow.go
```diff
@@ -45,18 +45,9 @@ func GetCodeFlagSet() *pflag.FlagSet {
 // WORKFLOWID_CODE defines a new workflow identifier
 var WORKFLOWID_CODE workflow.Identifier = workflow.NewWorkflowIdentifier(codeWorkflowName)
 
-func getSastSettings(engine workflow.Engine) (*contract.SastResponse, error) {
+func getSastSettings(engine workflow.Engine) (*common.SastResponse, error) {
 	config := engine.GetConfiguration()
```
ShawkyZ, line 49: I'm not sure if this func is still needed. 
Shouldn't the code_workflow get the sastSettings from the config?

pkg/local_workflows/code_workflow_test.go
```diff
@@ -98,7 +99,7 @@ func Test_Code_entrypoint(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotNil(t, rs)
 	assert.Equal(t, expectedData, rs[0].GetPayload().(string))
-	assert.Equal(t, 1, sastSettingsCalled)
+	assert.Equal(t, 2, sastSettingsCalled)
```
ShawkyZ, line 102: why?

internal/api/api.go
```diff
@@ -196,9 +197,9 @@ func (a *snykApiClient) GetSelf() (contract.SelfResponse, error) {
 	return selfData, nil
 }
 
-func (a *snykApiClient) GetSastSettings(orgId string) (contract.SastResponse, error) {
-	var response contract.SastResponse
-	var defaultResult contract.SastResponse
+func (a *snykApiClient) GetSastSettings(orgId string) (common.SastResponse, error) {
+	var response common.SastResponse
+	var defaultResult common.SastResponse
```
ShawkyZ, line 202: @PeterSchafer fyi we need SastResponse struct to be outside internal pkg to use it in ls.

pkg/configuration/storage.go
```diff
@@ -3,14 +3,14 @@ package configuration
 import (
 	"context"
 	"encoding/json"
+	"github.com/snyk/go-application-framework/internal/utils"
```
ShawkyZ, line 6: please move the import back to be at the end

pkg/local_workflows/code_workflow.go
```diff
@@ -2,13 +2,13 @@ package localworkflows
 
 import (
 	"fmt"
+	"github.com/snyk/go-application-framework/pkg/common"
```
ShawkyZ, line 5: same here

pkg/local_workflows/code_workflow_test.go
acke: Because now its also called when then default func is setup.

pkg/app/app.go
```diff
@@ -44,6 +44,28 @@ func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Co
 	return callback
 }
 
+func defaultFuncGetSastSettings(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger, apiClientFactory func(url string, client *http.Client) api.ApiClient) configuration.DefaultValueFunction {
+	callback := func(existingValue interface{}) (interface{}, error) {
+		if existingValue != nil {
+			return existingValue, nil
+		}
+		client := engine.GetNetworkAccess().GetHttpClient()
```
ShawkyZ, line 52: shouldn't this use the apiClientFactory to get a client?

pkg/app/app.go
```diff
@@ -44,6 +44,28 @@ func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Co
 	return callback
 }
 
+func defaultFuncGetSastSettings(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger, apiClientFactory func(url string, client *http.Client) api.ApiClient) configuration.DefaultValueFunction {
```
PeterSchafer, line 47: Issue: Please don't add domain logic into `app`, which is more or less domain independent. There is already a function registered in the `code_workflow`, please let's use this one.

pkg/local_workflows/code_workflow.go
PeterSchafer: See https://github.com/snyk/go-application-framework/pull/319/files#r2013897675

pkg/app/app.go
acke: So we want:
1. Rewrite the engine.GetConfiguration() to have the sast settings instead of just a ConfigurationSastEnabled configuration.
2. Setup the callback in code_workflow that will query the API for sast settings whenever asked for.
3. NOT setup a callback for sast settings in app.

pkg/app/app.go
ShawkyZ: I thought about this and it might not make sense to have caching for this in GAF and leave it up to the consumer to cache if necessary.

pkg/app/app.go
PeterSchafer: I don't think so, what I wanted to say is, to keep product and domain specific logic close to each other. This means in this case that the default function for SAST specific configuration values should be registered [here](https://github.com/snyk/go-application-framework/blob/main/pkg/local_workflows/code_workflow.go#L126) and implemented somewhere there as well.

One benefit is, that these values only get available if an application registers the code workflow and are not wide spread through the code base.

internal/api/api.go
PeterSchafer: Understand! Can we find a more descriptive name than `common`?

PR #321, opened 2025-03-27
fix: using getSastSettingsConfig for sastsettings

acke:
Changes in Go Application Framework:

- Added a getSastSettingsConfig callback function that executes when the snyk-ls asks for sast settings from the Go Application Framework Configuration.
- Moved `sast_settings.go` from "`/internal/api/contract` to `local_workflows/code_workflow/sast_contract` to be used in snyk-ls.
- Introduced a new constant for the new sast settings called `ConfigurationSastSettings` in `/pkg/local_workflows/code_workflow/native_workflow.go`
- Removed the cachedResponse
- Added `RemoveHeaderField` function that removes a static header field value from requests.

### DIFF
--- a/internal/mocks/api.go
+++ b/internal/mocks/api.go
@@ -10,7 +10,7 @@ import (
 
 	gomock "github.com/golang/mock/gomock"
 	contract "github.com/snyk/go-application-framework/internal/api/contract"
-	"github.com/snyk/go-application-framework/pkg/local_workflows/code_workflow/sast_contract"
+	sast_contract "github.com/snyk/go-application-framework/pkg/local_workflows/code_workflow/sast_contract"
 )
 
 // MockApiClient is a mock of ApiClient interface.
--- a/internal/mocks/api.go
+++ b/internal/mocks/api.go
@@ -10,8 +10,7 @@ import (
 
 	gomock "github.com/golang/mock/gomock"
 	contract "github.com/snyk/go-application-framework/internal/api/contract"
-
-	"github.com/snyk/go-application-framework/pkg/common"
+	"github.com/snyk/go-application-framework/pkg/local_workflows/code_workflow/sast_contract"
 )
 
 // MockApiClient is a mock of ApiClient interface.
@@ -83,10 +82,10 @@ func (mr *MockApiClientMockRecorder) GetOrgIdFromSlug(slugName interface{}) *gom
 }
 
 // GetSastSettings mocks base method.
-func (m *MockApiClient) GetSastSettings(orgId string) (common.SastResponse, error) {
+func (m *MockApiClient) GetSastSettings(orgId string) (sast_contract.SastResponse, error) {
 	m.ctrl.T.Helper()
 	ret := m.ctrl.Call(m, "GetSastSettings", orgId)
-	ret0, _ := ret[0].(common.SastResponse)
+	ret0, _ := ret[0].(sast_contract.SastResponse)
 	ret1, _ := ret[1].(error)
 	return ret0, ret1
 }
--- a/internal/mocks/api.go
+++ b/internal/mocks/api.go
@@ -10,6 +10,8 @@ import (
 
 	gomock "github.com/golang/mock/gomock"
 	contract "github.com/snyk/go-application-framework/internal/api/contract"
+
+	"github.com/snyk/go-application-framework/pkg/common"
 )
 
 // MockApiClient is a mock of ApiClient interface.
@@ -81,10 +83,10 @@ func (mr *MockApiClientMockRecorder) GetOrgIdFromSlug(slugName interface{}) *gom
 }
 
 // GetSastSettings mocks base method.
-func (m *MockApiClient) GetSastSettings(orgId string) (contract.SastResponse, error) {
+func (m *MockApiClient) GetSastSettings(orgId string) (common.SastResponse, error) {
 	m.ctrl.T.Helper()
 	ret := m.ctrl.Call(m, "GetSastSettings", orgId)
-	ret0, _ := ret[0].(contract.SastResponse)
+	ret0, _ := ret[0].(common.SastResponse)
 	ret1, _ := ret[1].(error)
 	return ret0, ret1
 }
--- a/pkg/common/sast_settings.go
+++ b/pkg/common/sast_settings.go
@@ -1,4 +1,4 @@
-package contract
+package common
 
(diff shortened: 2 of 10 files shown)


internal/mocks/api.go
acke, line 1: Should I generate and commit this file to the pr?

pkg/networking/networking.go
```diff
@@ -213,6 +213,23 @@ func (n *networkImpl) addDefaultHeader(request *http.Request) {
...
 		}
 	}
+
+	ua := n.config.Get(configuration.USER_AGENT)
+	if ua != nil {
+		userAgent, ok := ua.(UserAgentInfo)
+		if !ok {
+			n.logger.Printf("a is not of type UserAgentInfo")
+		}
+		if userAgent.App == "snyk-ls" {
```
PeterSchafer, line 223: Issue: Please don't mix framework logic with application names and headers. Why can't this be configured from the application?

pkg/local_workflows/code_workflow.go
```diff
@@ -67,10 +58,27 @@ func getSastSettings(engine workflow.Engine) (*contract.SastResponse, error) {
...
+			return existingValue, nil
+		}
+
+		response, err := getSastSettings(engine)
+		if err != nil {
+			engine.GetLogger().Err(err).Msg("Failed to access settings.")
+			return false, err
+		}
+
+		return *response, nil
```
PeterSchafer, line 77: Question: any reason why you don't return a pointer?

pkg/common/sast_settings.go
```diff
@@ -1,4 +1,4 @@
-package contract
```
PeterSchafer, line 1: Suggestion: I think there are better names than common. Maybe instead of creating a new package in the public library API, move it into `code_workflow`

pkg/networking/networking.go
acke: It was suggested that we want to separate concerns by encapsulating request headers in GAF. We will then let GAF infer the correct headers using information from the UserAgent.

PeterSchafer: Maybe there are different interpretations of separation of concerns but if we look at the consequence of the change we are extending a framework to know about possible applications that are using it. While the idea of a framework, sdk or application development kit is to be application agnostic as much as possible to be usable in many context. 

And I know this is not yet consistently done, but this should mean we shouldn't be aiming for a better, clearer architecture. Do you see any possibility to inject the wanted behaviour into the network stack?

bastiandoetsch: This is hard to discuss in writing. It would be better, to have a sync on that on a call - IMHO, you're right, but you're also wrong ;).

PeterSchafer: Now I'm very curious 😃

pkg/local_workflows/code_workflow.go
acke: Ill will change the function to return a pointer to stay consistent with the implementations in the class.

pkg/common/sast_settings.go
acke: The reason I put sast_settings.go in a separate package instead of code_workflow was to avoid the cyclic dependency. 

Can we put sast_settings.go in a package with a better name than common? 
How about local_workflows/code_workflow/sast_contract

PeterSchafer: great idea!

internal/mocks/api.go
PeterSchafer: yes please

pkg/networking/networking.go
PeterSchafer: For next weeks chat: What I meant is that it should be possible to use [dynamicHeader](https://github.com/snyk/go-application-framework/blob/main/pkg/networking/networking.go#L43C2-L43C23) to realize the exact same behaviour, just by returning an empty array when the value should not be set.

ShawkyZ, line 223: This would panic if userAgent is not of UserAgent type. Since it will just log and continue

pkg/local_workflows/code_workflow.go
```diff
@@ -67,10 +59,27 @@ func getSastSettings(engine workflow.Engine) (*contract.SastResponse, error) {
 		return &tmp, err
 	}
 
-	engine.GetConfiguration().Set(key, &tmp)
+	engine.GetConfiguration().Set(org, &tmp)
```
bastiandoetsch, line 62: why would we set the org here?

pkg/configuration/constants.go
```diff
@@ -19,6 +19,7 @@ const (
 	AUTHENTICATION_TOKEN string = "snyk_token"
 	//nolint:gosec // not a token value, a configuration key
 	AUTHENTICATION_BEARER_TOKEN     string = "snyk_oauth_token"
+	USER_AGENT                      string = "snyk_user_agent"
```
PeterSchafer, line 22: Issue: Please [follow the namespacing recommendatios](https://snyksec.atlassian.net/wiki/x/QgDgjg?atlOrigin=eyJpIjoiMTE4NzNhMjc3NTk3NGFjYzk2MWNmYWUwOGQxMDY5ZjgiLCJwIjoiYyJ9) for internal keys, currently this would expose an environment variable for the CLI.

pkg/networking/networking.go
```diff
@@ -213,6 +221,19 @@ func (n *networkImpl) addDefaultHeader(request *http.Request) {
 			request.Header.Add(k, v[i])
 		}
 	}
+
+	ua := n.config.Get(configuration.USER_AGENT)
```
PeterSchafer, line 225: Question: In general I'm still struggling to understand from a semantic perspective why user agent is a configuration value. Why didn't we just introduce something like `setUserAgent()`  to the NetworkAccess? This would provide a clear interface.

PeterSchafer: Suggestion: To me this change seems unrelated to what the PR actually wants to achieve, so maybe separate things, so that you can get the task at hand resolved and merged and we can have architectural discussions and decisions outside and non-blocking.

PeterSchafer:
Question: Did you see that the tests are failing?

pkg/configuration/constants.go
acke: Removed useragent from configuration.

acke:
> Question: Did you see that the tests are failing?

Yes, fixed!

PR #322, opened 2025-04-04
fix: WIP allow direct fetching of feature flags

rrama:
Previously all feature flags had to be registered with the GAF configuration before they could be used.


PeterSchafer:
> I need this for https://github.com/snyk/snyk-ls/pull/820/files#diff-d74ed587c3dae58e0c1ddc2a0eec71bdb58552ef0b6852d664fbb849e868ec19R59 so I can directly fetch feature flags using GAF's API. I will be calling:
> 
> ```go
> config_utils.CurrentFeatureFlagChecker().GetFeatureFlag(config.CurrentConfig().Engine(), ffStr)
> ```
> 
> I needed to mock the fetching of feature flags for the test (https://github.com/snyk/snyk-ls/pull/820/files#diff-29de3939a8441248dd44aa2f2eb09f23a57909cf126d85840d0a1519580e25f3R37), which is why I did the horrible stuff with and interface and struct. If there is a better way to do it, then please let me know, it's my first time using mocks in Go.

Question: I think I'm not yet understanding the use case or the problem you are solving well enough. Could you please explain this a bit more?

rrama:
> Question: I think I'm not yet understanding the use case or the problem you are solving well enough. Could you please explain this a bit more?

Certainly, for some context, in the language server we have the LSP command [`snyk.getFeatureFlagStatus`](https://github.com/snyk/snyk-ls/#:~:text=Feature%20Flag%20Status%20Command), this allows a client to request the status of any feature flag by it's name. Currently the language server has its own code to fetch feature flags (and SastSettings) from Snyk's backend API. We are wanting to deprecate the language server having its own API client and instead make it use GAF's API client.

Now I am trying to rewrite the `snyk.getFeatureFlagStatus` command using GAF's API client. I didn't realise that `Configuration.GetBool(key string)` for feature flags used a key that is internal to GAF (e.g. `FF_CODE_CONSISTENT_IGNORES = "internal_snyk_code_ignores_enabled"`), which is mapped to the real feature flag name using:
```go
config_utils.AddFeatureFlagToConfig(engine, configuration.FF_CODE_CONSISTENT_IGNORES, "snykCodeConsistentIgnores")
```
For the `snyk.getFeatureFlagStatus` command to work, it needs to bypass this internal key name mapping to feature flag name code, otherwise we won't be able to fetch feature flags unless they are already explicitly known to GAF (plus we'd have to create a reverse map of feature flag names to the internal key name to then be able to then call `GetBool`).

Now whether this is the best solution for bypassing the internal config key names or not, I don't know. The code was a lot simpler before I added the ability to mock the response so that I could get the unit test for the `snyk.getFeatureFlagStatus` command to work. I have ended up spewing a lot of bloat code in that is horrible :'(

PeterSchafer:
Thank you @rrama ! That is very helpful.

pkg/local_workflows/config_utils/feature_flag.go
```diff
@@ -4,27 +4,62 @@ import (
...
+	return currentFFC
+}
+
+func SetCurrentFeatureFlagChecker(ffc FeatureFlagChecker) {
+	mutex.Lock()
+	defer mutex.Unlock()
+	currentFFC = ffc
+}
+
+func (ffc *featureFlagCheckerImpl) GetFeatureFlag(engine workflow.Engine, featureFlagName string) (bool, error) {
```
PeterSchafer, line 53: Suggestion: The singleton construct is not really used here, so I would recommend to just have a method `GetFeatureFlag()` without the whole singleton.

ShawkyZ: maybe instead of a singleton the fetureFlagChecker can be set on the engine lvl `engine.SetFeatureFlagChecker`. or smth similar ? That way tests and other clients can set the mock.
cc @PeterSchafer

PeterSchafer: sorry, it seems I missed any notification on this thread. is this still something we are looking to merge?

PeterSchafer:
Just to summarize my understanding.

**Goal**: We want to enable accessing feature flags whose names are not yet known at compile time.

1. After re-reading the following part of the comment above
> For the snyk.getFeatureFlagStatus command to work, it needs to bypass this internal key name mapping to feature flag name code, otherwise we won't be able to fetch feature flags unless they are already explicitly known to GAF (plus we'd have to create a reverse map of feature flag names to the internal key name to then be able to then call GetBool).

I think there is a misunderstanding. the configuration key and the feature flag name can be the same, there is no such thing as an internal key required. Only if you want to support setting a feature flag remotely and via other means like environment variables, a mapping is helpful. But this is not what you need.

2. For your use case, you could also just not use the configuration at all, since you are just interested in the value based on a name. You could replace the snyk-ls api client by the [gaf api client](https://github.com/snyk/go-application-framework/blob/main/internal/api/api.go#L23). The only change it would need would be to expose this api client.

I hope this makes sense @rrama. If I'm missing something here, please let's jump on a call to get to a resolution. 
And again sorry for the delay on this topic.

pkg/local_workflows/config_utils/feature_flag.go
rrama: @PeterSchafer Yes, still looking to merge, but there isn't a rush for it anymore, so can be cooldown. Would you prefer I set it on the function on the engine level as Shawky suggested?

pkg/local_workflows/config_utils/feature_flag.go
```diff
@@ -4,27 +4,62 @@ import (
...
+
+type FeatureFlagChecker interface {
+	GetFeatureFlag(engine workflow.Engine, featureFlagName string) (bool, error)
+}
+
+type featureFlagCheckerImpl struct {
+}
+
+var (
+	currentFFC FeatureFlagChecker
```
bastiandoetsch, line 34: I don't like that we introduce a global variable here. Why is it needed? Can't we just have an instance in the configuration that is created when the configuration is created?

pkg/local_workflows/config_utils/feature_flag.go
bastiandoetsch: Haha, I see, everyone already commented on that. I think it's a bit overengineered. Having an interface for mocking is fine, but IMHO a singleton is not needed. I'd implement the interface in Configuration as it's in the configuration domain, not the engine.

pkg/local_workflows/config_utils/feature_flag.go
```diff
@@ -4,27 +4,62 @@ import (
...
+
+func SetCurrentFeatureFlagChecker(ffc FeatureFlagChecker) {
+	mutex.Lock()
+	defer mutex.Unlock()
+	currentFFC = ffc
+}
+
+func (ffc *featureFlagCheckerImpl) GetFeatureFlag(engine workflow.Engine, featureFlagName string) (bool, error) {
+	config := engine.GetConfiguration()
+	httpClient := engine.GetNetworkAccess().GetHttpClient()
```
bastiandoetsch, line 55: I wonder whether it would be better to have config & httpclient injected, to decouple and avoid side-effects (same like it's done with apiClient).

pkg/local_workflows/config_utils/feature_flag.go
```diff
@@ -4,27 +4,62 @@ import (
 	"github.com/snyk/go-application-framework/internal/api"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/workflow"
+	"sync"
 )
 
+//go:generate $GOPATH/bin/mockgen -source=feature_flag.go -destination ../../mocks/feature_flag.go -package mocks -self_package github.com/snyk/go-application-framework/pkg/local_workflows/config_utils
+
```
bastiandoetsch, line 11: nice :)

rrama:
I am going to move this to a draft for now until the next cooldown, then I will look closer at the solutions suggested and then we can pick which works best.

PeterSchafer:
Since we might need the same functionality in another use case, I went ahead and opened #349  which should do what you need. Let me know if this makes sense.

PR #323, opened 2025-04-07
feat: create ignores workflow

ShawkyZ:
## How to test:
To test e2e use CLI branch https://github.com/snyk/cli/tree/spike/iaw_pitch
## Command:
`snyk ignore create --id=findingsId`

## Changes:
* Move analytics generated model from clients package to analytics pkg. Since auto-generated models can't live in the same package because they share many fields.
* Small refactor: move the call to fetch git repo url to a separate function to reuse it. Also introduce a func to get branch name from directory path.
* Policy API spec and model generation
* New function in configuration `GetStringWithError `to not suppress errors and deliver them back to the caller.
* Create Ignores implementaion that supports both interactive (CLI) and non-interactive (IDE) modes.
* Bash script to download the API spec.


pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	payload := createPayload(repoUrl, branchName, expire, ignoreType, reason, findingsId)
+	response, err := sendCreateIgnore(invocationCtx, payload, url)
+
+	if err != nil {
+		return nil, err
+	}
+
+	uiErr := userInterface.Output("\n✅ Your ignore request has been submitted for approval.")
```
bastiandoetsch, line 173: edge case: wouldn't it only be "for approval" if it's needing governance and approval? Can we see that somehow from org/group settings, if ignores must be approved or are auto-approved?

ShawkyZ: we have no knowledge about this. But we can actually know from the API response. If Review == pending then the flag is on.

pkg/utils/git/git.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/config"
+)
+
+func RepoUrlFromDir(inputDir string) (string, error) {
+	_, remoteConfig, err := RepoFromDir(inputDir)
+	if err != nil {
+		return "", err
+	}
+	repoRemoteUrl := remoteConfig.URLs[0]
```
bastiandoetsch, line 15: Should we return all and display a user selection if it's not clear? Should we maybe take origin as default if there's more than one remote repository?

Use case:
- contribution to repo without write rights
- original repo is "upstream"
- fork is "origin"
Most likely the ignore should be in origin, not upstream, as we are not commiter in upstream.

PeterSchafer: I think we anyways have some duplication [here](https://github.com/snyk/code-client-go/blob/30592824b40dc41d7a720b2f27b2581fdc3e7808/internal/util/repository.go#L25). IMHO a great candidate to deduplicate when both repos are merged.

pkg/configuration/configuration.go
```diff
@@ -395,6 +397,18 @@ func (ev *extendedViper) GetWithError(key string) (value interface{}, err error)
 	return value, err
 }
 
+// GetStringWithError returns a configuration value as string.
+func (ev *extendedViper) GetStringWithError(key string) (string, error) {
```
PeterSchafer, line 401: Suggestion: There is a proposal for a different solution [here](https://github.com/snyk/go-application-framework/blob/main/internal/utils/convert.go). Please let's align on one solution and stick to it.

pkg/instrumentation/get-target-id.go
```diff
@@ -185,24 +184,11 @@ func filesystemBaseId(path string) (*url.URL, error) {
 }
 
 func gitBaseId(path string) (*url.URL, error) {
-	repo, err := git.PlainOpenWithOptions(path, &git.PlainOpenOptions{
-		DetectDotGit: true,
-	})
-
+	repo, remoteConfig, err := git.RepoFromDir(path)
```
PeterSchafer, line 187: Praise: Nice refactoring!

internal/api/policy/2024-10-15/spec.yaml
```diff
@@ -0,0 +1,1378 @@
+# OpenAPI spec generated by vervet, DO NOT EDIT
+components:
```
ShawkyZ, line 2: just fyi: spec will probably be updated it again tmrw

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,323 @@
...
+
+	return output, err
+}
+
+func getIgnoreRequestDetailsStructure(expire *time.Time, userName string, ignoreType string) string {
+	expireDisplayText := "Does not expire"
+	if expire != nil {
+		expireDisplayText = expire.Format(time.DateOnly)
+	}
+	return fmt.Sprintf("  Requested on:  2024-08-10\n  Requested by:  %s\n  Expiration:    %s\n  Ignore type:   %s", userName, expireDisplayText, ignoreType)
```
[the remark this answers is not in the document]
ShawkyZ: oops

acke (approved):
LGTM

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	expirationDescription = "Expiration (YYYY-MM-DD)"
+
+	RemoteRepoUrlKey         = code_workflow.ConfigurationRemoteRepoUrlFlagname
+	remoteRepoUrlDescription = "Remote Repository URL"
+
+	InteractiveKey    = "interactive"
+	EnrichResponseKey = "enrich_response"
+
+	policyAPIVersion = "2024-10-15"
+	policyApiTimeout = 5 * time.Minute
```
j-luong, line 52: question: 5 minute timeout for an api request is quite high. If a request is expected to take this long, there should be optimisations that can be made elsewhere?

suggested: 30 seconds. 1 minute max

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	deleteFlagSet.String(IgnoreIdKey, "", ignoreIdDescription)
+	// If set to false, no response will be returned
+	deleteFlagSet.Bool(EnrichResponseKey, false, "")
+	deleteFlagSet.Bool(InteractiveKey, true, "")
+	_, err = engine.Register(WORKFLOWID_IGNORE_DELETE, workflow.ConfigurationOptionsFromFlagset(deleteFlagSet), ignoreDeleteWorkflowEntryPoint)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
j-luong, line 99: nitpick: I would follow YAGNI here and omit these until we need it

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func ignoreEditWorkflowEntryPoint(_ workflow.InvocationContext, _ []workflow.Data) (output []workflow.Data, err error) {
+	panic("not implemented")
+}
+
+func ignoreDeleteWorkflowEntryPoint(_ workflow.InvocationContext, _ []workflow.Data) (output []workflow.Data, err error) {
+	panic("not implemented")
+}
```
j-luong, line 108: nitpick: YAGNI here also

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	config := invocationCtx.GetConfiguration()
+	id := invocationCtx.GetWorkflowIdentifier()
+
+	interactive := config.GetBool(InteractiveKey)
+	addCreateIgnoreDefaultConfigurationValues(invocationCtx, interactive)
+
+	userName, err := getUser(invocationCtx)
+	if err != nil {
+		return nil, fmt.Errorf("user is not authenticated: %w", err)
+	}
```
j-luong, line 122: question: this seems to ignore the types of errors returned from `getUser()` and just return a not authenticated error?
suggest: implement Error Catalog errors instead of using `fmt.Error()`

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	noAuthErr := fmt.Errorf("user is not authenticated")
+	if len(data) > 0 {
+		user, ok := data[0].GetPayload().(string)
+		if !ok {
+			return "", noAuthErr
+		}
+		return user, nil
+	} else {
+		return "", noAuthErr
+	}
```
j-luong, line 287: question: is this necessary? the relevant error should bubble up from the invocation, this extra logic seems to just obfusicate the original error

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("user is not authenticated: %w", err)
+	}
+
+	findingsId, err := config.GetStringWithError(FindingsIdKey)
+	if err != nil {
+		return nil, err
+	}
+
+	ignoreType, err := config.GetStringWithError(IgnoreTypeKey)
```
j-luong, line 129: suggestion: same as above, these params can be validate at the start so we can exit early if needed

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	repoUrl, err := config.GetStringWithError(code_workflow.ConfigurationRemoteRepoUrlFlagname)
```
j-luong, line 134: thought: there's probably a better way to share flagnames across workflows here

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	expireDisplayText := "Does not expire"
+	if expire != nil {
+		expireDisplayText = expire.Format(time.DateOnly)
+	}
+	return fmt.Sprintf("  Requested on:  %s\n  Requested by:  %s\n  Expiration:    %s\n  Ignore type:   %s", requestedOn, userName, expireDisplayText, ignoreType)
+}
+
+func getExpireValue(config configuration.Configuration) (*time.Time, error) {
+	shouldParse := config.IsSet(ExpirationKey) || config.GetBool(InteractiveKey)
+	if !shouldParse {
```
j-luong, line 204: nitpick: calling it `shouldNotExpire` and not checking for negation (i.e. `if shouldNotExpire {...}`), is easier to reason with

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+		uiErr := userInterface.Output(fmt.Sprintf("👉🏼 Make sure the code containing the issue is committed, "+
+			"and pushed to a remote origin, so the approvers are able to analyze it.\n%s", getIgnoreRequestDetailsStructure(expire, userName, ignoreType)))
+		if uiErr != nil {
+			logger.Print(uiErr)
+		}
+	}
+
+	branchName, branchNameErr := git.BranchNameFromDir(config.GetString(configuration.INPUT_DIRECTORY))
+	if branchNameErr != nil {
+		logger.Print(branchNameErr)
```
j-luong, line 161: suggestion: `logger.Warn().Err(branchNameErr).Msg("some mesg")`

internal/api/analytics/2024-03-07/analytics.go
```diff
@@ -1,7 +1,7 @@
 // Package clients provides primitives to interact with the openapi HTTP API.
 //
 // Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.3.0 DO NOT EDIT.
-package clients
+package v20240307
```
j-luong, line 4: question: why did this need to change?

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -0,0 +1,333 @@
...
+	response, err := sendCreateIgnore(invocationCtx, payload)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if interactive {
+		uiErr := userInterface.Output("\n✅ Your ignore request has been submitted.")
+		if uiErr != nil {
+			logger.Print(uiErr)
```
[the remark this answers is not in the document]
ShawkyZ: We can't have one package for all auto generated clients/models. Auto generated models and clients share properties with the same name.

internal/api/analytics/2024-03-07/analytics.go
ShawkyZ: We can't have one package for all auto generated clients/models. Auto generated models and clients share properties with the same name.

pkg/local_workflows/ignore_workflow/ignore_workflow.go
ShawkyZ: Yup agree. 5 mins was an arbitrary timeout I put there. But 1 min should be more than enough

ShawkyZ: Imo it would be cool if the caller can give his own context.Context to a wf invocation. But that's a topic for another day.

pkg/local_workflows/ignore_workflow/ignore_workflow.go
ShawkyZ: I would rather choose panic for not implemented cases.

ShawkyZ: Thinking about it again. I can see why this would be too much. Since it will now panic when ignore edit cmd is executed.
I will change to return an err instead.
Thanks Jason

PeterSchafer: Issue: Please don't register not implemented workflows! The behaviour of the CLI would be to show the help if something is not registered, if you return an error, the CLI will show an error, this seems like a poor UX. Also who would call these not implemented workflows? I assume no one.

pkg/local_workflows/ignore_workflow/config.go
```diff
@@ -0,0 +1,101 @@
+package ignore_workflow
```
j-luong, line 1: question: do we validate the inputs? i.e. expiry date validation

j-luong (approved):

pkg/configuration/configuration.go
ShawkyZ: Synced with @CatalinSnyk. Changed his impl to GetBoolWithError.

PR #327, opened 2025-04-11
chore: redact the username when scrubbing logs

PeterSchafer:
This PR adds an additional redact rule to the scrubbing logger to redact the os username.

### DIFF
--- a/pkg/logging/scrubbingLogWriter.go
+++ b/pkg/logging/scrubbingLogWriter.go
@@ -19,6 +19,7 @@ package logging
 import (
 	"fmt"
 	"io"
+	"os/user"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/logging/scrubbingLogWriter.go
```diff
@@ -188,6 +189,13 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
 		groupToRedact: 2,
 		regex:         regexp.MustCompile(s),
 	}
+
+	user, err := user2.Current()
```
dotkas, line 193: `user2` 😞 Can't we do
```suggestion
	u, err := user.Current()
```

Or does this break something I haven't seen?

dotkas (approved):

PR #328, opened 2025-04-11
fix: [CLI-841] redacting extension endpoint of instrumentation payload

dotkas:
In https://github.com/snyk/cli/pull/5813 we are going to pass the legacy v1 analytics to the new analytics service, utilizing the `extension` attribute in the payload.

Since the extension payload can be used by all product lines, it would be good practice to ensure proper PII protection of that part of the payload.

Further, since the `v1` data is already passed through that sanitation logic, it makes sense to do the same to the data when passed to the analytics service.

I also re-wrote the test a tad to make each test independent of the other, as some test runners were mutating the assertion object. Also a tendency to flakiness as timestamps were being compared.

### DIFF
--- a/pkg/analytics/instrumentation_collector.go
+++ b/pkg/analytics/instrumentation_collector.go
@@ -1,8 +1,10 @@
 package analytics
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"os/user"
 	"time"
 
 	"github.com/snyk/error-catalog-golang-public/snyk_errors"
@@ -123,13 +125,14 @@ func (ic *instrumentationCollectorImpl) AddExtension(key string, value interface
 
 func GetV2InstrumentationObject(collector InstrumentationCollector) (*api.AnalyticsRequestBody, error) {
 	t, ok := collector.(*instrumentationCollectorImpl)
-	if ok {
-		return t.GetV2InstrumentationObject(), nil
+	if !ok {
+		return nil, fmt.Errorf("failed to convert collector")
 	}
-	return nil, fmt.Errorf("failed to convert collector")
+
+	return t.GetV2InstrumentationObject()
 }
 
-func (ic *instrumentationCollectorImpl) GetV2InstrumentationObject() *api.AnalyticsRequestBody {
+func (ic *instrumentationCollectorImpl) GetV2InstrumentationObject() (*api.AnalyticsRequestBody, error) {
 	a := ic.getV2Attributes()
 
 	d := api.AnalyticsData{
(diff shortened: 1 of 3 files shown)


pkg/analytics/instrumentation_collector.go
```diff
@@ -123,23 +125,53 @@ func (ic *instrumentationCollectorImpl) AddExtension(key string, value interface
...
-func (ic *instrumentationCollectorImpl) GetV2InstrumentationObject() *api.AnalyticsRequestBody {
+func (ic *instrumentationCollectorImpl) GetV2InstrumentationObject() (*api.AnalyticsRequestBody, error) {
 	a := ic.getV2Attributes()
 
 	d := api.AnalyticsData{
 		Type:       ic.instrumentationType,
 		Attributes: a,
 	}
 
+	// Since the `extension` attribute in the analytics payload is a value any
```
PeterSchafer, line 143: Question: I'm just thinking a bit about the error handling here. if the sanitization fails for whatever reason, we might want to still send the other data and just skip the extension data?

dotkas: Yes, good point. Will fix.

pkg/analytics/instrumentation_collector.go
```diff
@@ -121,22 +124,72 @@ func (ic *instrumentationCollectorImpl) AddExtension(key string, value interface
 	ic.extension[key] = value
 }
 
-func GetV2InstrumentationObject(collector InstrumentationCollector) (*api.AnalyticsRequestBody, error) {
+func GetV2InstrumentationObject(logger *zerolog.Logger, collector InstrumentationCollector) (*api.AnalyticsRequestBody, error) {
```
PeterSchafer, line 127: Question: I think this interface change might break the current integration in snyk-ls. You might see this when building the CLI with this gaf version.

dotkas: Got it. Maybe I need another approach.

pkg/analytics/instrumentation_collector.go
dotkas: Update on this for reviewers: We synced offline and decided we don't have a strong opinion on an approach between:

Fail hard
Fail semi-hard with type assertions on different errors higher in the call stack, but having to default to string comparisons
Fail soft but pass a logger object (thus causing a change in the SDK API requiring downstream changes to other libs)
Current decision: Make a new API endpoint WithLogger or something of that sort to avoid breaking changes, possibly update snyk-ls to use this if IDE agrees.
Fail soft, log nothing, lose the context in debug logs

pkg/local_workflows/report_analytics_workflow.go
```diff
@@ -224,7 +224,7 @@ func instrumentScanDoneEvent(invocationCtx workflow.InvocationContext, input wor
 	ic.SetTestSummary(toTestSummary(scanDoneEvent.Data.Attributes.UniqueIssueCount, scanDoneEvent.Data.Type))
 	ic.AddExtension("device_id", scanDoneEvent.Data.Attributes.DeviceId)
 
-	data, err := analytics.GetV2InstrumentationObject(ic)
+	data, err := analytics.GetV2InstrumentationObjectWithLogger(logger, ic)
```
dotkas, line 227: Not a super big fan of this approach myself, but open to suggestions. More thoughts [here](https://github.com/snyk/go-application-framework/pull/328/files#r2044143016).

pkg/analytics/instrumentation_collector_test.go
```diff
@@ -75,6 +35,10 @@ func Test_InstrumentationCollector(t *testing.T) {
 	})
 
 	t.Run("it sets the userAgent application data", func(t *testing.T) {
+		ic := setupBaseCollector(t)
+		expectedV2InstrumentationObject := buildExpectedBaseObject(t)
+
+		mockUserAgent := networking.UserAgentInfo{}
```
dotkas, line 41: These were added as previously, as the tests were dependent on each other. Now they're added explicitly.

pkg/analytics/instrumentation_collector_test.go
```diff
@@ -220,14 +205,112 @@ func Test_InstrumentationCollector(t *testing.T) {
...
+		expectedV2InstrumentationObject.Data.Attributes.Interaction.Extension = nil
+
+		actualV2InstrumentationObject, err := GetV2InstrumentationObjectWithLogger(&logger, ic)
+		assert.NoError(t, err)
+		expectedV2InstrumentationJson, err := json.Marshal(expectedV2InstrumentationObject)
+		assert.NoError(t, err)
+		actualV2InstrumentationJson, err := json.Marshal(actualV2InstrumentationObject)
+		assert.NoError(t, err)
+
+		assert.Equal(t, string(expectedV2InstrumentationJson), string(actualV2InstrumentationJson))
```
PeterSchafer, line 248: Suggestion: maybe let's use assert.JSONEq()

PeterSchafer (approved):

PR #330, opened 2025-04-13
fix: ignore determination should rely on Status property and ensure IAW FF enabled

ShawkyZ:
## Determining the Current Suppression State
Each finding in the SARIF includes a suppression array. To determine the current suppression entry, follow this logic:
1. Search for an accepted or empty suppression status entry:
   * If found, return this entry.
   * Result: accepted

2. If not found, search for an underReview entry:

   * If found, return this entry.
   * Result: underReview

3. If still not found, search for a rejected entry:
   * If found, return this entry.
   * Result: rejected

4. If no valid suppression entry is found, or if values are invalid:
Return nil.


internal/utils/sarif/sarif.go
```diff
@@ -86,6 +87,21 @@ func CreateCodeSummary(input *sarif.SarifDocument, projectPath string) *json_sch
...
+// GetIgnoreDetails returns current suppression and whether the issue is ignored or not
+// in case of an existing suppression Test API will always enrich the SARIF with one element.
+// Currently, surfaced states are either "underReview" or "accepted"
+// This logic might change in the future when the "rejected" state is surfaced
+func GetIgnoreDetails(suppressions []sarif.Suppression) (bool, *sarif.Suppression) {
+	if len(suppressions) == 0 {
+		return false, nil
+	}
+
+	currentSuppression := &suppressions[0]
```
PeterSchafer, line 99: Question: To determine the answer to the question [if a result is suppressed or not](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790911), shouldn't we be inspecting all entries in the array? 

Suggestion: Maybe it makes sense to have something like `getSuppressionStatus() *status` where nil indicates no suppression and any other value is the actual state.  WDT?

ShawkyZ: That makes sense, I discussed this today with the ignores team.
We want to ensure that the determination behavior would be future-proof to de-risk breaking changes in the future.

I'm converting this to a draft for now, I will continue working on it in the next days.

pkg/utils/sarif/sarif.go
```diff
@@ -86,6 +86,42 @@ func CreateCodeSummary(input *sarif.SarifDocument, projectPath string) *json_sch
 	return summary
 }
 
+// HasSuppressionInStatus returns true if the suppression with the provided status is found.
+func HasSuppressionInStatus(suppressions []sarif.Suppression, status sarif.SuppresionStatus) bool {
```
rrama, line 90: Function name and doc is misleading.
```go
HasSuppressionInStatus([]sarif.Suppression{
	{
		Status: sarif.Rejected,
	},
	{
		Status: sarif.UnderReview,
	},
}, sarif.Rejected)
```
returns `false`, which isn't what I would expect given the function doc.

Maybe call it `IsHighestSuppressionStatus`.

rrama (approved):
TY, LGTM.

internal/presenters/testdata/with-ignores.json
```diff
@@ -671,6 +671,7 @@
               {
                 "justification": "False positive",
                 "kind": "inSource",
+                "status": "accepted",
```
PeterSchafer, line 674: Question: was this change required. From my understanding I would have assumed that status undefined is accepted as well.

PeterSchafer: How does the current response look like? Does it contain the status already? if not, do we need to adapt the data structure to omid status when empty?

PeterSchafer:
Please use this change in the CLI as well, to see if it doesn't break anything. thank you!

internal/presenters/testdata/with-ignores.json
ShawkyZ: The testdata change wasn't necessary, yes, empty will be treated as accepted.

Current response doesn't have the Status field, yet. Currently it will default to empty until the API enriches it.

The ignore wf will be guarded behind IAW FF for now, https://github.com/snyk/go-application-framework/pull/337

ShawkyZ: this shouldn't break since empty is still being treated as accepted.

ShawkyZ: I will add additional cases to test both

ShawkyZ:
Waiting on CLI acceptance tests before merging 
https://github.com/snyk/cli/pull/5849

PR #331, opened 2025-04-14
feat: [DGP-165] Low-level Unified Test API client

bsalomon-snyk:
Adds Unified Test API low-level routines based on the closed-beta OpenAPI spec.  These will be built upon to form the updated test CLI. See README.md for update process: pulling the latest spec & regenerating the Go definitions.


scripts/pull-down-unified-test-api-spec.sh
```diff
@@ -0,0 +1,40 @@
...
+git pull
+
+# Return to project directory
+cd -
+
+# place OpenAPI build artifacts
+mkdir -p $TEST_API_GENERATED
+
+cp $API_SPEC_PATH/internal/api/closed-beta/versions/2024-10-15/spec.yaml $TEST_API_GENERATED/unified-test-api.yaml
+echo $GENERATE_DATE $API_SPEC_BRANCH $API_COMMIT > $TEST_API_GENERATED/generated.txt
```
cmars, line 40: Consider shortening the name, given the parent directories provide the context,

```suggestion
cp $API_SPEC_PATH/internal/api/closed-beta/versions/2024-10-15/spec.yaml $TEST_API_GENERATED/spec.yaml
echo $GENERATE_DATE $API_SPEC_BRANCH $API_COMMIT > $TEST_API_GENERATED/generated.txt
```

scripts/pull-down-unified-test-api-spec.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+TEST_API_DIR=$(realpath ./internal/api/test-api)
+
+# Data source
+API_SPEC_PATH=$TEST_API_DIR/unified-test-repo
+API_SPEC_BRANCH=${API_SPEC_BRANCH:-main}
+
+# Outputs
+TEST_API_GENERATED=$TEST_API_DIR/unified-test-spec
```
cmars, line 10: Using the API version date here would be a nice way to indicate at-a-glance which version of the test API we're targeting. I think the parent directory `testapi` gives enough context.

```suggestion
TEST_API_GENERATED=$TEST_API_DIR/2024-10-15
```

scripts/pull-down-unified-test-api-spec.sh
cmars, line 1: It looks like there's an existing script here with a similar name, `scripts/pull-down-test-api-spec.sh`.

I wonder if we still need it, given that we're no longer using Cue for local findings. Let's ask the CLI team if we can repurpose it.

cmars (commented):
Some thoughts on naming & directory structure, though I'll defer to the CLI team on how they would like things arranged.

internal/api/test-api/cfg.yaml
```diff
@@ -0,0 +1,6 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/oapi-codegen/oapi-codegen/HEAD/configuration-schema.json
+package: unifiedtestapi
```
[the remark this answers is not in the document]
bsalomon-snyk: Thanks.  Changed in next commit.

scripts/pull-down-unified-test-api-spec.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+TEST_API_DIR=$(realpath ./internal/api/test-api)
+
+# Data source
+API_SPEC_PATH=$TEST_API_DIR/unified-test-repo
```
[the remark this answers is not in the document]
bsalomon-snyk: I like this solution! Adding in next commit.

scripts/pull-down-unified-test-api-spec.sh
bsalomon-snyk: Replaced the existing, unused Cuelang script.  However, I left "internal/cueutils/*" because it's referenced by pkg/local_workflows/output_workflow_test.go, for a test that validates Sarif data.

scripts/local-client/main.go
cmars, line 1: Suggest renaming this from `local-client` to `testapi-example`, so that the meaning and purpose in this repo is clearer.

A short README explaining what it is & how to use it would be nice as well.

scripts/local-client/testdata/req-dep-graph-test-npm.json
cmars, line 1: Do we need to include this? Or could we use CLI's `snyk depgraph` output directly? I'd prefer that (see scripts/local-client/cli.sh in test-api-shim for an example of how to structure the request using `snyk depgraph --json` output).

cmars (commented):
LGTM with a couple of suggestions and history squashed into a single commit.

scripts/local-client/main.go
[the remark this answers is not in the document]
bsalomon-snyk: It's definitely a dev test/usage demonstration.  Another version of it is coming that tests/demos the high-level API, when that feature branch is ready.  Open to moving them but not sure where makes sense outside the repo (perhaps if there's a dev test repo somewhere).  I'll rename it for now.

cmars (approved):

scripts/pull-down-test-api-spec.sh
```diff
@@ -1,36 +1,37 @@
 #!/usr/bin/env bash
 
-API_SPEC_PATH=$(realpath ../dragonfly)
+TEST_API_DIR=$(realpath ./internal/api/testapi)
```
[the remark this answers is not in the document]
bsalomon-snyk: Hi @PeterSchafer . We talked to Lucian about this as well. The previous spec from the Dragonfly repo is obsolete, and it's no longer located where the script tries to pull it from. The new spec we're working on it should replace it.

Before the new spec is ready, if builds require a spec then the one in the GAF repo is possibly good enough. We could bring back the Dragonfly branch where it lives if an update is needed.

PR #332, opened 2025-04-15
chore: extend network trace level debug logs to include request and response bodies

PeterSchafer:
* This PR includes a fix where network logging was only available on debug level, trace level disabled it completely
* This PR extends the debug level logging of network requests to include request and response bodies in trace level logging. Response bodies continue to be logged in debug level logging when the response status code > 400.

### DIFF
--- a/pkg/networking/logging_test.go
+++ b/pkg/networking/logging_test.go
@@ -1,0 +1,285 @@
+package networking
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+)
+
+func Test_shouldNotLog(t *testing.T) {
+	testCases := []struct {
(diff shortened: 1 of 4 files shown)


pkg/networking/logging_test.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+func Test_LogRequest_happyPath_gzipped_request(t *testing.T) {
+	expectedBody := "hello world"
+	gzipBuffer := bytes.NewBuffer([]byte{})
+	gzipWriter := gzip.NewWriter(gzipBuffer)
+	n, err := gzipWriter.Write([]byte(expectedBody))
+	assert.NoError(t, err)
+	assert.Greater(t, n, 0)
+	err = gzipWriter.Close()
```
dotkas, line 136: Pretty nitpicky, but should you maybe defer the `Close()` in case something goes wrong in the writes?

dotkas (approved):

pkg/networking/logging_test.go
PeterSchafer: I can't really defer the close in this case, because I need it to flush the writer content into the buffer before making the request.

PR #334, opened 2025-04-22
chore: WIP improve Makefile and add PR checklist

rrama:
Plus regenerated files which had been missed in previous PRs.

For reference, here is the PR checklist for the LS: https://github.com/snyk/snyk-ls/blob/main/.github/pull_request_template.md


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+### Description
```
PeterSchafer, line 1: Praise: Nice, long overdue!

.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
...
+
+_Provide description of this PR and changes._
+
+### Checklist
+
+- [ ] Tests added and all succeed (`make test`)
+- [ ] Regenerated mocks, etc. (`make generate`)
+- [ ] Linted (`make lint`)
```
PeterSchafer, line 9: Suggestion: let's add a point to integrate this in the CLI to ensure early that the changes didn't break anything.

PeterSchafer: Since we had past occasions where gaf changes broke the CLI and its Acceptance tests, I would say `open a PR to test and after merging in gaf ensure to merge in the CLI` as well. otherwise we will see multiple changes coming to the CLI at once just because they were never integrated there before making it difficult to debug.

rrama: "open a PR to test", I'm not sure I follow how opening a PR would test it, unless you are suggesting that the person updates the go.mod in the CLI PR to point to their GAF PR's commit?

rrama: Hi @PeterSchafer, I've given the wording an attempt, let me know what you think.

PR #335, opened 2025-04-22
fix: separate title and message in progress bar [IDE-1067]

rrama:
Since the language server has to conform to the LSP, we are bound by how the progress bars are defined there, as such it makes sense to align the GAF progress bar with it.


rrama:
In the future I would like to split the ProgressBar interface into two types of progress bar, one representing infinite progress and other quantifiable progress. Something like:
```
type UserInterface interface {
	Output(output string) error
	OutputError(err error, opts ...Opts) error
-->     NewProgressBar(title string, message string) ProgressBar
-->     NewProgressBarInfinite(title string, message string) ProgressBarInfinite
	Input(prompt string) (string, error)
}
type ProgressBar interface {
	UpdateProgress(progress float64) error
	UpdateProgressWithMessage(progress float64, message string) error
	Clear() error
}

// ProgressBarInfinite is like ProgressBar, but without a reported percentage,
// so the spinner is infinite.
type ProgressBarInfinite interface {
	SetMessage(message string) error
	Clear() error
}
```
(It is reasonably trivial to have the underlying `consoleProgressBar` struct be being duck-typeable to bother interfaces)

But alas, that change ended up being too large for now. So I have parked it here: https://github.com/snyk/go-application-framework/tree/fix/progress-bar-v3

pkg/ui/progressbar.go
```diff
@@ -29,52 +29,58 @@ const (
...
 	UpdateProgress(progress float64) error
 
-	// SetTitle sets the title of the progress bar, which is displayed next to the bar.
-	// The title provides context or description for the operation that is being tracked.
-	SetTitle(title string)
+	// SetMessage sets the message of the progress bar, which is displayed next to the bar.
+	// The message is displayed alongside the title passed in by `NewProgressBar(title string)`.
+	// The message provides context or description for the operation that is being tracked.
+	// Set to "" to clear the message.
+	SetMessage(message string)
```
PeterSchafer, line 66: Question: @rrama did you integrate these changes in the CLI? I assume given that this PR changes interfaces, it'll require changes to existing extensions. Did you consider doing this without a breaking change?

rrama:
Very stale, so closing, we can re-do it when the Jira ticket is picked up again.

PR #340, opened 2025-04-28
feat: [DGP-166] begin high-level unified test API

bsalomon-snyk:
### Goal: provide a high-level API for unified tests, starting with abstracting away manual polling for test results.

This commit allows executing a test, polling for results, and reporting overall status. Findings still use the low-level API so that will come later.

  - Includes [DGP-168], the paginated findings PR, so reviewers can review the entire process from creating a test to fetching findings.
  - Moved test metadata (previously "FinalStatus") and findings into a single structure called Result.
  - result.Findings() is lazy-loaded so users can avoid the overhead of fetching them when unwanted.
  - Set a minimum polling interval of 1 second and a default of 2 seconds.
  - Updates the TestAPI spec with the addition of TestID in the GetJob response body. This means it doesn't have to be parsed from GetJob's
    redirect URL in order to call GET /tests/{test_id}.

unit tests:
  - Restructured with a focus on the testapi calls in the test, moving much of the mock backend server and asserts into supporting calls.

### Decisions of note:
  - Relies on the GAF for creating an http.Client{} which handles 50x and certain 40x by returning an error from the request
  (i.e., the test API expects an error in those cases rather than getting a Response with resp.StatusCode() of 50x/40x).
  This decision is to eliminate redundant checks where the Unified Test API was checking codes after the GAF client already did so.
  However, it requires that users use the GAF client (or roll their own error handler) and disable auto-follow of redirects.  This decision
  may need revisiting if/when GAF clients (say, users external to Snyk) want to build a client -- we'd want to enforce these rules, or wrap their
  client, etc.

[DGP-168]: https://snyksec.atlassian.net/browse/DGP-168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/api/testapi/testapi.go
```diff
@@ -0,0 +1,385 @@
+package testapi
```
PeterSchafer, line 1: Suggestion: We might want to think of moving this to `pkg` rather than `internal` to make the high level testapi usable by other modules.

internal/api/testapi/testapi.go
```diff
@@ -0,0 +1,385 @@
...
+
+import (
```
PeterSchafer, line 3: Suggestion: Please add some tests for the high level client. The tests might even replace the scripts obsolete at some point. The benefit is just that tests keep running while the scripts at some point might unknowingly diverge from the code.

internal/api/testapi/testapi.go
bsalomon-snyk: the latest commit moves it to /pkg/testapi

scripts/pull-down-test-api-spec.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-TEST_API_DIR=$(realpath ./internal/api/testapi)
+TEST_API_DIR=$(realpath ./pkg/testapi)
```
PeterSchafer, line 3: Suggestion: Since there will be more coming, how about `pkg/apiclients/testapi`?

pkg/testapi/testapi_test.go
```diff
@@ -0,0 +1,1173 @@
...
+	defer cleanup()
+
+	handle, err := testClient.StartTest(ctx, params)
+
+	// Assert
+	assert.NoError(t, err, "StartTest returned an unexpected error")
+	assert.NotNil(t, handle, "StartTest returned a nil handle")
+
+	// No result before Wait
+	initialResult, err := handle.Result()
```
PeterSchafer, line 101: Question: Would we expect a `handle.Wait()` before `handle.Result()`? Maybe we should do some artificial waiting in the handler to make this actually wait?

bsalomon-snyk: Yes, we should always have a Wait before getting the Result.  This test was checking only up to StartTest(), and then making sure Result() does not return an error if called too early.  But that could set a bad example if people copy the test so I'll add a proper Wait.  I don't think we want an actual sleep in here since it's all mocked.

pkg/apiclients/testapi/testapi_test.go
```diff
@@ -0,0 +1,1173 @@
...
+
+	// Act
+	clientConfig := testapi.Config{
+		APIVersion:   apiVersion,
+		PollInterval: 10 * time.Millisecond, // Use a short poll interval for faster tests
+	}
+	testClient, cleanup := startMockServerAndClientWithConfig(t, clientConfig, handler)
+	defer cleanup()
+
+	handle, err := testClient.StartTest(ctx, params)
```
PeterSchafer, line 79: Question: sorry if I could have been able to find this somewhere, but quick question, is it possible to inject an `http.Client`?

bsalomon-snyk: Yes, it got refactored into startMockServerAndClientWithConfig() to avoid setting it up over and over for the tests, but that's where the TestClient takes an http.Client like this:

`	testHTTPClient := &http.Client{
		Transport: baseClient.Transport,
		Timeout:   baseClient.Timeout,
		CheckRedirect: func(req *http.Request, via []*http.Request) error {
			return http.ErrUseLastResponse
		},
	}

	testClient, err := testapi.NewTestClient(server.URL, clientConfig, testapi.WithHTTPClient(testHTTPClient))
`

there's an example here, too, where the http.Client() adds a SNYK_TOKEN header to every request: https://github.com/bsalomon-snyk/dragonfly-scripts/blob/main/testapi-example-high-level/main.go#L180-L184

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+			errChan <- localErr
+			return
+		}
+
+		h.setResult(result)
+		errChan <- nil
+	})
+
+	// Wait until the polling goroutine (started by runOnce.Do) finishes.
+	<-h.doneChan
```
PeterSchafer, line 191: Question: Do we need both channels? If I'm not mistaken, `errChan` takes care of the waiting as well.

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+		return nil, fmt.Errorf("job %s completed (303) but the final Test ID path was incomplete in the response", jobID)
+	}
+}
+
+// Create a detailed error message for unexpected HTTP statuses during polling.
+func formatUnexpectedJobStatusError(resp *GetJobResponse, jobID uuid.UUID) error {
+	errorBodyBytes := resp.Body
+	errMsg := fmt.Sprintf("unexpected status code %d polling job %s", resp.StatusCode(), jobID)
+	// Attempt to get more detail from structured error response
+	if resp.ApplicationvndApiJSON400 != nil && len(resp.ApplicationvndApiJSON400.Errors) > 0 {
```
PeterSchafer, line 227: Suggestion: You might want to use [this helper](https://github.com/snyk/error-catalog-golang/blob/main/snyk_errors/json_api_error.go#L39) to get error information independent of the response code.

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+		errMsg = fmt.Sprintf("%s: %s", errMsg, resp.ApplicationvndApiJSON400.Errors[0].Detail)
+	} else if len(errorBodyBytes) > 0 {
+		// Fallback to raw body if no structured error
+		errMsg = fmt.Sprintf("%s, response: %s", errMsg, string(errorBodyBytes))
+	}
+	return fmt.Errorf("%s", errMsg)
```
PeterSchafer, line 233: nitpick: Possibly a good candidate for an Error Catalog error.

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+func (h *testHandle) pollJobToCompletion(ctx context.Context) (*uuid.UUID, error) {
+	ticker := time.NewTicker(h.client.config.PollInterval)
+	defer ticker.Stop()
+
+	getJobParams := &GetJobParams{Version: h.client.config.APIVersion}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context canceled while polling job %s: %w", h.jobID, ctx.Err())
```
PeterSchafer, line 246: Suggestion: Introduce an error catalog error `Operation Cancelled`

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+// Test state is captured in Result's State, Errors, and Warnings.
+func (h *testHandle) fetchResultStatus(ctx context.Context, testID uuid.UUID) (*Result, error) {
+	getTestParams := &GetTestParams{Version: h.client.config.APIVersion}
+	resp, err := h.client.lowLevelClient.GetTestWithResponse(ctx, h.orgID, testID, getTestParams)
+	if err != nil {
+		return nil, fmt.Errorf("get test %s request failed: %w", testID, err)
+	}
+
+	if resp.ApplicationvndApiJSON200 == nil {
+		return nil, fmt.Errorf("test %s had unexpected status %d or wrong response body format", testID, resp.StatusCode())
```
PeterSchafer, line 294: Suggestion: Maybe combine this with with other occurrences of unexpected Status and return an common `Unexpected Status` error catalog error.

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+		r.findingsError = fetchErr
+	})
+	return r.findings, r.findingsComplete, r.findingsError
+}
+
+// fetchFindingsInternal is an unexported method on Result that contains the actual
+// pagination and fetching logic.
+func (r *Result) fetchFindingsInternal(ctx context.Context) ([]FindingData, bool, error) {
+	if r.TestID == nil {
+		return nil, false, fmt.Errorf("cannot fetch findings, TestID is nil in internal method")
```
PeterSchafer, line 369: Suggestion: There seem to be some duplications in errors with slight variation. Please try to unify them to reduce cardinality.

pkg/apiclients/testapi/testapi.go
bsalomon-snyk: Good observation. errChan is local to Wait() and is for copying the error out of its worker thread since Do doesn't return a value.  doneChan is external to Wait() and signals the user in a non-blocking way that Wait() has completed (by selecting on testHandle.Done()).

I tried combining them but ran into concurrency trouble because the user's Select consumes the error that Wait wants to return.

My earlier approach was to keep a testHandle.waitError to send between Wait and its worker, protected by a mutex.  That didn't look great because it adds an error to testHandle that we only use once, as a return value from Wait(). 
 The advantage is if the user repeatedly calls Wait(), it can return the same error every time.  
 
 I'm adding an explicit message for followup calls to call Wait only once and use the first result.  Open to other suggestions!

pkg/apiclients/testapi/testapi.go
bsalomon-snyk: all set in next commit

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+	testID := *r.TestID
+	allFindings := []FindingData{}
+	var startingAfter *string
+
+	listParams := &ListFindingsParams{
+		Version: r.handle.client.config.APIVersion,
+		Limit:   ptr(int8(100)), // Use max limit
```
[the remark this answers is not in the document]
bsalomon-snyk: done in next commit. thanks

pkg/apiclients/testapi/testapi.go
bsalomon-snyk: Done in next commit

pkg/apiclients/testapi/testapi.go
cmars: Can we pick that up in a followup? We've identified the need for new error categories here, let's add tickets to get them into the error catalog, and improve them here once they're agreed on there.

The error catalog is lacking a lot of definition for these kinds of general non-HTTP-service errors, and there will probably be a need for discussion to handle them thoughtfully there.

I don't think we should be blocking this work on that.

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,530 @@
...
+)
+
+// StartTestParams defines parameters for the high-level StartTest function.
+type StartTestParams struct {
+	OrgID   string
+	Subject TestSubjectCreate
+}
+
+// High-level results and status of a completed test.
+type Result struct {
```
PeterSchafer, line 78: Question: Should we introduce a Result interface and return this instead of a struct? Looking at the functions to retrieve findings for example, it is more than just a DTO, so enabling mocking and DI seems like a good idea. WDT?

pkg/apiclients/testapi/testapi.go
```diff
@@ -0,0 +1,530 @@
...
+	}
+
+	if cfg.PollInterval <= 0 {
+		cfg.PollInterval = DefaultPollInterval
+	} else if cfg.PollInterval < MinPollInterval {
+		cfg.PollInterval = MinPollInterval
+	}
+
+	if cfg.APIVersion == "" {
+		cfg.APIVersion = "2024-10-15"
```
PeterSchafer, line 111: nitpick: please create a constant 😄

pkg/apiclients/testapi/testapi.go
PeterSchafer: Yes we can. Happy to split error handling from the PR as long as we follow up on it. It is an too important topic that gets easily forgotten.

@dotkas and I want to provide some guidelines to more consistent error message.

pkg/apiclients/testapi/testapi.go
bsalomon-snyk: updated!

pkg/apiclients/testapi/testapi.go
bsalomon-snyk: went ahead with this. It makes sense in the long run when we have more consumers. Thanks.

pkg/apiclients/testapi/testapi.go
bsalomon-snyk: settled on SNYK-0003 here and included the http status code, since "unable to process the request" seemed fitting.

PeterSchafer (approved):

PR #343, opened 2025-05-06
chore: update LFM with suppressions

CatalinSnyk:
What does this do?

- Reintroduces the old script for updating the LFM using our dragonfly branch (temporary, until we some more time to play around with the new test api)
- Updates the LFM, specifically the Suppression type to include the id and status fields
- Updates the transformations related to suppressions

TODOs:
- Some tests are failing, and will do so until the presenters are updated to account for the new changes

### DIFF
--- a/pkg/local_workflows/local_models/models.gen.go
+++ b/pkg/local_workflows/local_models/models.gen.go
@@ -29,6 +29,21 @@ const (
 	TypesBusinessCriticalityRiskFactorValueMedium TypesBusinessCriticalityRiskFactorValue = "medium"
 )
 
+// Defines values for TypesCodeSastFingerprintAssetV1Scheme.
+const (
+	Snykassetfindingv1 TypesCodeSastFingerprintAssetV1Scheme = "snyk/asset/finding/v1"
+)
+
+// Defines values for TypesCodeSastFingerprintProjectV1Scheme.
+const (
+	Snykorgprojectfindingv1 TypesCodeSastFingerprintProjectV1Scheme = "snyk/org/project/finding/v1"
+)
+
+// Defines values for TypesCodeSastFingerprintRepositoryV1Scheme.
+const (
+	Snykorgrepositoryfindingv1 TypesCodeSastFingerprintRepositoryV1Scheme = "snyk/org/repository/finding/v1"
+)
+
 // Defines values for TypesCodeSastFingerprintV0Scheme.
 const (
 	CodeSastV0 TypesCodeSastFingerprintV0Scheme = "code-sast-v0"
@@ -116,21 +131,6 @@ const (
 	Findings TypesFindingResourceType = "findings"
 )
 
-// Defines values for TypesFingerprintAssetV1Scheme.
-const (
-	Snykassetfindingv1 TypesFingerprintAssetV1Scheme = "snyk/asset/finding/v1"
-)
-
-// Defines values for TypesFingerprintProjectV1Scheme.
-const (
-	Snykorgprojectfindingv1 TypesFingerprintProjectV1Scheme = "snyk/org/project/finding/v1"
-)
-
-// Defines values for TypesFingerprintRepositoryV1Scheme.
-const (
-	Snykorgrepositoryfindingv1 TypesFingerprintRepositoryV1Scheme = "snyk/org/repository/finding/v1"
-)
-
 // Defines values for TypesGitSCMInputType.
 const (
 	GitScm TypesGitSCMInputType = "git-scm"
@@ -179,11 +179,11 @@ const (
 	True  TypesSuggestedPackageUpgradeUpgradeConflicts = true
 )
 
-// Defines values for TypesSuppressionKind.
+// Defines values for TypesSuppressionStatus.
 const (
-	TypesSuppressionKindIgnored               TypesSuppressionKind = "ignored"
-	TypesSuppressionKindOther                 TypesSuppressionKind = "other"
-	TypesSuppressionKindPendingIgnoreApproval TypesSuppressionKind = "pending_ignore_approval"
+	Accepted    TypesSuppressionStatus = "accepted"
+	Rejected    TypesSuppressionStatus = "rejected"
+	UnderReview TypesSuppressionStatus = "underReview"
 )
 
 // Defines values for TypesTestContextSdlcStage.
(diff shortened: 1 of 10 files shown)


pkg/local_workflows/local_models/transform.go
```diff
@@ -132,7 +140,7 @@ func mapSuppressions(res sarif.Result) *TypesSuppression {
 			},
 		},
 		Justification: &suppression.Justification,
-		Kind:          "ignored",
+		Status:        TypesSuppressionStatus(status),
```
PeterSchafer, line 143: Question: was kind removed from the data structure?

pkg/local_workflows/testdata/sarif-snyk-goof-ignores.json
```diff
@@ -2983,10 +2987,12 @@
           },
           "suppressions": [
             {
+              "guid":"bffa1c3c-7d67-4ede-b6d0-7d5208ad041d",
+              "state": "accepted",
               "justification": "",
               "properties": {
                 "category": "not-vulnerable",
-                "expiration": null,
+                "expiration": "never",
```
PeterSchafer, line 2995: Question: as this is test data, I assume this is how the API response looks like. are we confident that the expiration will not be null from the API?

pkg/local_workflows/local_models/models.gen.go
```diff
@@ -1179,12 +1179,13 @@ type TypesSuggestedPackageUpgradeUpgradeConflicts bool
 type TypesSuppression struct {
 	// Details Suppression meta data
 	Details       *TypesSuppressionDetails `json:"details,omitempty"`
+	Id            *openapi_types.UUID      `json:"id,omitempty"`
```
PeterSchafer, line 1182: Suggestion: stick to string types and handle such identifiers transparently. there is no benefit in assuming a certain format of the string like a uuid but an inherit maintenance effort by using it.

pkg/local_workflows/local_models/transform.go
```diff
@@ -109,19 +111,25 @@ func mapRules(sarifRule sarif.Rule) TypesRules {
...
-	expiration := ""
+	expiration := "never"
 	ignored_email := ""
 	if suppression.Properties.Expiration != nil {
 		expiration = *suppression.Properties.Expiration
 	}
 	if suppression.Properties.IgnoredBy.Email != nil {
 		ignored_email = *suppression.Properties.IgnoredBy.Email
 	}
+	var idPtr *uuid.UUID
```
PeterSchafer, line 126: see https://github.com/snyk/go-application-framework/pull/343/files#r2112343828

internal/presenters/templates/finding.component.tmpl
```diff
@@ -7,12 +18,18 @@
...
    Ignored on: {{ formatDatetime .Attributes.Suppression.Details.IgnoredOn "2006-01-02T15:04:05.999999999Z07:00" "January 02, 2006" }}
    Ignored by: {{ .Attributes.Suppression.Details.IgnoredBy.Name }}
    Reason:     {{ .Attributes.Suppression.Justification }}
    {{- end}}
+   {{- end}}
 {{end}}
-{{define "severity"}}{{if .Attributes.Suppression }}{{else}} ✗{{end}} [{{.Attributes.Rating.Severity.Value}}]{{end}}
+
+{{- define "severity"}}
+  {{- if or (not .Attributes.Suppression) (and (.Attributes.Suppression) (eq .Attributes.Suppression.Status "rejected"))}} ✗ {{else}} ! {{end -}}
```
PeterSchafer, line 33: see https://github.com/snyk/go-application-framework/pull/343/files#r2112369606

pkg/local_workflows/local_models/transform.go
j-luong: Yes, but more like corrected rather than removed.

The "kind" implemented in LFM refers to the SARIF's "state"/"status" and so this change was made. The SARIF's "kind" was/is hardcoded in the `local_findings.sarif.tmpl` as it was never used in LFM.

pkg/local_workflows/testdata/sarif-snyk-goof-ignores.json
j-luong: The expiration shouldn't be null from the api, it's an optional field and so can be omitted. 

suggestion: It might make sense to reflect this the same way in the sarif to represent a never expiring ignore, that way we can keep the date string format consistent for the `expiration` field.

pkg/local_workflows/local_models/models.gen.go
CatalinSnyk: I was following the other usages for identifiers definitions in the schemas, and since most of them were using the `@format('uuid')` decorator I also used it. I'll do another schema update for this

pkg/local_workflows/testdata/sarif-snyk-goof-ignores.json
CatalinSnyk: Made the expiration a pointer in the LFM struct, so it's better represented; and I just adapted the human readable presenter to handle this case in d1c5166, while the SARIF presenter will just not render the expiration field if the ignore never expires.

internal/presenters/templates/local_finding.tmpl
```diff
@@ -1,12 +1,19 @@
 {{- define "details" }}
     {{- $sortedFindings := .Findings | sortFindingBy "Attributes.Rating.Severity.Value" .Summary.Counts.CountKeyOrderAsc.Severity }}
-    {{- $openFindings := $sortedFindings | filterFinding (notHasField "Attributes.Suppression.Kind") }}
-    {{- $ignoredFindings := $sortedFindings | filterFinding (hasField "Attributes.Suppression.Kind") }}
+    {{- $openFindings := filterFindingsOr $sortedFindings (notHasField "Attributes.Suppression.Status") (fieldEquals "Attributes.Suppression.Status" "rejected") }}
```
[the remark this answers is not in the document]
CatalinSnyk: I would have love to make this work, but unfortunately I could not the template to work by passin in the callback `isOpenFinding`. If you refrence a function name inside the template, the function imediately get's called, so in my case isOpenFinding would get called with 0 arguments.

But considering `filterFindingsOr` and `filterFinding` (is this function also missing an `s` from it's name?) are only used for these filtering parts, would it not be better to just make 3 functions `filterOpenFindings`, `filterPendingFindings` and `filterIgnoredFindings` and just avoid extra complexity?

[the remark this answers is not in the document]
CatalinSnyk: Added in 3157c64, I think `filterFindingsOr` could be removed now since it's not used anymore, unless we keep it for the future.

PR #344, opened 2025-05-06
feat: update local findings human readable renderer to support IAW

j-luong:
This PR adds support for IAW in the human readable renderer


internal/presenters/funcs.go
```diff
@@ -141,6 +153,38 @@ func filterFinding(cmpFunc func(any) bool, findings []local_models.FindingResour
...
+// maintaining the original order.
+func filterFindingsOr(findings []local_models.FindingResource, cmpFuncs ...func(any) bool) (filteredFindings []local_models.FindingResource) {
+	filteredFindings = make([]local_models.FindingResource, 0)
+	addedFindingsHashes := make(map[string]bool)
+
+	for _, finding := range findings {
+		for _, cmpFunc := range cmpFuncs {
+			if cmpFunc(finding) {
+				// would be nice to use the finding ID, but this is not being sent currently
+				// e.g. `filteredFindingsMap[finding.Id.String()] = finding`
```
ShawkyZ, line 166: Hashing the finding feels like an expensive op. Specially in the context of monorepos.
Why is the findingsId missing in this case? Is this related to the field not existing in the dragonfly spec?

internal/presenters/templates/finding.component.tmpl
```diff
@@ -7,12 +12,18 @@
    Info: {{.Attributes.Message.Text}}
 
    {{- if .Attributes.Suppression}}
+   {{- if or (eq .Attributes.Suppression.Status "accepted") (eq .Attributes.Suppression.Status "underReview")}}
```
ShawkyZ, line 15: will this cover empty status as well ? `status == ""` is an accepted ignore

pkg/local_workflows/local_models/transform.go
```diff
@@ -115,16 +115,16 @@ func mapSuppressions(res sarif.Result) *TypesSuppression {
 	if suppression == nil {
 		return nil
 	}
-	expiration := ""
+	expiration := "never"
```
ShawkyZ, line 118: I'm not familiar with the local findings model, but just fyi, with the current WF impl, empty expiration is treated as never expires.

internal/presenters/templates/finding.component.tmpl
j-luong: Yes, the templates are based off the findings model, which ensures that there's always a [status](https://github.com/snyk/go-application-framework/pull/344/files#diff-48f318a603014a4e8398fbc15f04f9e41320bbb4d7dfcecdf675ce20521c1fedR114)

internal/presenters/funcs.go
j-luong: good point, AFAIK the findings ID is not being sent by code-client-go, we can compose some fields instead of hashing to ensure uniqueness.

pkg/local_workflows/local_models/transform.go
j-luong: Let's follow up with UX to determine the best wording here for consistency across IDE, CLI, & app UI

CatalinSnyk (approved):

PR #346, opened 2025-05-09
feat: Introduce retry middleware

PeterSchafer:
This PR introduces a retry middleware and adds it to the default network stack. Per default retries are disabled and the middleware will act as a pass through, but for `preview` versions and if the feature is enabled, network requests failing with a certain status code will automatically be retried.

This can be enabled for example in the CLI by setting the environment variable `INTERNAL_NETWORK_REQUEST_MAX_ATTEMPTS=<N>`, where <N> is the maximum number of request attempts, 1 being the minimum.

The Status Codes that will be retried can be found [here](https://github.com/snyk/go-application-framework/pull/346/files#diff-295d20b71bb7f7fdc47d3503d7fb17595142bb2c48dd5801cc4ad0517afb1a55R26)

There are some changes/improvement that are left for future iterations, like optimizations around memory consumption, observability and rollout.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 
 require (
 	github.com/cenkalti/backoff/v4 v4.3.0
+	github.com/cenkalti/backoff/v5 v5.0.2
 	github.com/go-git/go-git/v5 v5.14.0
 	github.com/gofrs/flock v0.12.1
 	github.com/mattn/go-isatty v0.0.20
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -1,0 +1,169 @@
+package middleware
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/cenkalti/backoff/v5"
+	"github.com/rs/zerolog"
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+)
+
+const defaultRetryCount uint = 1 // Per default retries (=1) are disabled and need to be enabled via the configuration
+const defaultRetryAfterSeconds = 5
+const maxRetryAfter = 10 * time.Minute
+const ConfigurationKeyRetryAttempts = "internal_network_request_max_attempts"
+const configurationKeyRetryAfter = "internal_network_request_retry_after_seconds"
+const retryCountHeaderKey = "Snyk-Request-Attempt-Count"
+
+// This lookup table defines the response status codes that should be retried
+var statusCodesToRetryLUT = map[int]bool{
+	http.StatusTooManyRequests:     true,
+	http.StatusTooEarly:            true,
+	http.StatusRequestTimeout:      true,
+	http.StatusInternalServerError: true,
+	http.StatusBadGateway:          true,
+	http.StatusServiceUnavailable:  true,
+	http.StatusGatewayTimeout:      true,
+}
+
+var errRetryNecessary = errors.New("retry with backoff")
+var errRetryAfterHeaderError = errors.New("retry-after is too much in the future")
+
+type RetryMiddleware struct {
+	nextRoundtripper http.RoundTripper
+	config           configuration.Configuration
+	logger           *zerolog.Logger
+}
+
+func NewRetryMiddleware(config configuration.Configuration, logger *zerolog.Logger, roundTripper http.RoundTripper) *RetryMiddleware {
+	return &RetryMiddleware{
+		nextRoundtripper: roundTripper,
+		config:           config,
+		logger:           logger,
+	}
+}
+
+func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
+	var finalResponse *http.Response
+	var finalError error
+	var localBodyBuffer []byte
+	var maxAttempts = defaultRetryCount
+	var retryAfterSeconds = defaultRetryAfterSeconds
+	var actualAttempts = 0
+
+	if tmp := (uint)(rm.config.GetInt(ConfigurationKeyRetryAttempts)); tmp > 0 {
+		maxAttempts = tmp
+	}
+
+	if tmp := rm.config.GetInt(configurationKeyRetryAfter); tmp > 0 {
+		retryAfterSeconds = tmp
+	}
+
+	// if a body is available, create a local copy to be able to use it multiple times
+	if req.Body != nil && maxAttempts > 1 {
+		// possible optimization for large request bodies to not buffer in memory but in filesystem
+		var localBufferError error
+		localBodyBuffer, localBufferError = io.ReadAll(req.Body)
+		closeError := req.Body.Close()
+
+		if localBufferError != nil {
+			return nil, localBufferError
+		}
+		if closeError != nil {
+			return nil, closeError
+		}
+
+		req.Body = io.NopCloser(bytes.NewBuffer(localBodyBuffer))
+	}
+
+	op := func() (*http.Response, error) {
+		actualAttempts++
+
+		// create a local copy of the request
+		localRequest := *req
+		if len(localBodyBuffer) > 0 {
+			localRequest.Body = io.NopCloser(bytes.NewBuffer(localBodyBuffer))
+		}
+
+		// try to send request
+		response, err := rm.nextRoundtripper.RoundTrip(&localRequest)
+
+		// keep track of actual retry attempts for monitoring/logging
+		if response != nil && response.Header != nil && actualAttempts > 1 {
+			response.Header.Set(retryCountHeaderKey, fmt.Sprintf("%d", actualAttempts))
+		}
+
+		// errors from the next round tripper cannot not be retried
+		if err != nil {
+			return response, backoff.Permanent(err)
+		}
+
+		// depending on the response determine if we should retry
+		if retryError := shouldRetry(response); retryError != nil {
+			rm.logger.Debug().Msgf("Retrying request, reason: %v", retryError)
+			return response, retryError
+		}
+
+		return response, nil
+	}
+
+	backoffMethod := backoff.NewExponentialBackOff()
+	backoffMethod.InitialInterval = time.Duration(retryAfterSeconds) * time.Second
+	finalResponse, finalError = backoff.Retry(req.Context(), op, backoff.WithBackOff(backoffMethod), backoff.WithMaxTries(maxAttempts))
+
+	// if retries fail to resolve the issue, we need to unset the locally used error type to not return it from the RoundTripper
+	if errors.Is(finalError, errRetryNecessary) {
+		finalError = nil
+	}
+
+	return finalResponse, finalError
+}
+
+func shouldRetry(response *http.Response) error {
+	if statusCodesToRetryLUT[response.StatusCode] {
+		fixRetryDelay := time.Duration(0)
+
+		// try to read retry-after header if available
+		if headerRetryAfterValue := response.Header.Get("Retry-After"); len(headerRetryAfterValue) > 0 {
+			fixRetryDelay = parseRetryAfterHeader(headerRetryAfterValue)
+		}
+
+		// if the fix retry delay is too big, we rather fail permanently than blocking too long
+		if fixRetryDelay > maxRetryAfter {
+			return backoff.Permanent(errRetryAfterHeaderError)
+		}
+
+		// if a retry after is defined, this is the time to wait for
+		if fixRetryDelay > 0 {
+			return &backoff.RetryAfterError{Duration: fixRetryDelay}
+		}
+
+		// if no retry after is defined, the backoff strategy determines the time to wait for
+		return errRetryNecessary
+	}
+
+	return nil
+}
+
+func parseRetryAfterHeader(headerRetryAfterValue string) time.Duration {
+	// Retry-After: 1230
+	if tmp, err := strconv.ParseInt(headerRetryAfterValue, 10, 64); err == nil {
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -25,6 +25,7 @@ require (
 
 require (
 	github.com/cenkalti/backoff/v4 v4.3.0
+	github.com/cenkalti/backoff/v5 v5.0.2
```
bastiandoetsch, line 28: why do we have both v4 and v5?

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+const defaultRetryCount uint = 3
+const configurationKeyRetryCount = "INTERNAL_MAX_NET_RETRIES"
+const defaultRetryAfterSeconds = 5
+
+type RetryMiddleware struct {
+	next   http.RoundTripper
+	config configuration.Configuration
+}
+
+func NewRetryMiddleware(roundTriper http.RoundTripper, config configuration.Configuration) *RetryMiddleware {
```
bastiandoetsch, line 25: nitpick: I'd put configuration first (as it makes sense to have a certain order like ctx, config, rest)

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+)
+
+const defaultRetryCount uint = 3
+const configurationKeyRetryCount = "INTERNAL_MAX_NET_RETRIES"
+const defaultRetryAfterSeconds = 5
+
+type RetryMiddleware struct {
+	next   http.RoundTripper
```
bastiandoetsch, line 21: nitpick: I'd rather not use `next`, but instead something like roundTripper. Next may be usual in many http filter chains, I still find it unintuitive.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
+	var finalResponse *http.Response = nil
+	var finalError error = nil
+	var localBodyBuffer []byte
+	var maxRetryCount = defaultRetryCount
+
+	// if a body is available, create a local copy to be able to use it multiple times
+	if req.Body != nil {
```
bastiandoetsch, line 39: suggestion: as this is a library and will be heavily used, I would suggest to have limits on body size. Imagine having a POST request with a 2 GB repo source code thingy (by accident or not). I see two options: either limit retries to a max request size or swap to filesystem after a threshold.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+		// possible optimization for large request bodies to not buffer in memory but in filesystem
```
bastiandoetsch, line 40: Haha, just saw this now. Doh. Same thing as above.

pkg/networking/middleware/retry_middleware.go
dotkas: Also a small typo `roundTriper => roundTripper` 😀

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return finalResponse, finalError
+}
+
+func shouldRetry(response *http.Response) error {
+	var retryError error
+
+	if response.StatusCode >= http.StatusInternalServerError ||
+		response.StatusCode == http.StatusTooManyRequests ||
+		response.StatusCode == http.StatusMovedPermanently {
```
dotkas, line 86: If this ends up growing in either number of codes or more complex logic, maybe a case-switch with some fallthrough would be easier to read.

dotkas: Also if `http.StatusMovedPermanently` we should probably not retry, or?

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if rm.config.IsSet(configurationKeyRetryCount) {
+		maxRetryCount = (uint)(rm.config.GetInt(configurationKeyRetryCount))
+	}
+
+	op := func() (*http.Response, error) {
+		// create a local copy of the request
+		localRequest := *req
+		if len(localBodyBuffer) > 0 {
+			localRequest.Body = io.NopCloser(bytes.NewBuffer(localBodyBuffer))
```
bastiandoetsch, line 57: For big bodies: maybe an approach like this (Gemini created code, so use with caution) would work without loading everything into memory:
```
package main

import (
	"fmt"
	"io"
	"log"
	"mime/multipart"
	"net/http"
	"os"
	"path/filepath"
)

func main() {
	filePath := "your-large-file.dat" // Path to your large file
	targetURL := "http://example.com/upload-multipart"
	fileFieldName := "myFile" // The field name for the file in the form

	// 1. Create a pipe
	pipeReader, pipeWriter := io.Pipe()

	// 2. Create a multipart writer
	multipartWriter := multipart.NewWriter(pipeWriter)

	// 3. Goroutine to write multipart data
	go func() {
		defer pipeWriter.Close() // Close writer when done or on error
		defer multipartWriter.Close() // Close multipart writer before pipeWriter

		// Add other form fields if needed
		if err := multipartWriter.WriteField("description", "This is a large file"); err != nil {
			log.Printf("Error writing field: %v", err)
			pipeWriter.CloseWithError(err) // Signal error to reader
			return
		}

		// Create a form file part
		partWriter, err := multipartWriter.CreateFormFile(fileFieldName, filepath.Base(filePath))
		if err != nil {
			log.Printf("Error creating form file: %v", err)
			pipeWriter.CloseWithError(err)
			return
		}

		// Open the large file
		file, err := os.Open(filePath)
		if err != nil {
			log.Printf("Error opening file: %v", err)
			pipeWriter.CloseWithError(err)
			return
		}
		defer file.Close()

		// Stream the file content to the part writer
		if _, err := io.Copy(partWriter, file); err != nil {
			log.Printf("Error copying file to part writer: %v", err)
			pipeWriter.CloseWithError(err)
			return
		}
	}()

	// 4. Create the HTTP request with the pipe reader as the body
	req, err := http.NewRequest("POST", targetURL, pipeReader)
	if err != nil {
		log.Fatalf("Error creating request: %v", err)
	}

	// 5. Set the Content-Type header from the multipart writer
	req.Header.Set("Content-Type", multipartWriter.FormDataContentType())

	// Content-Length is not set here, allowing for chunked transfer encoding.
	// If the server strictly requires Content-Length, you'd need to buffer the
	// multipart content first to calculate its size, which defeats the purpose of
	// pure streaming for very large files. Alternatively, some servers might
	// support a "final chunk" mechanism or require pre-signed URLs for large uploads.

	// 6. Send the request
	client := &http.Client{}
	resp, err := client.Do(req)
	if err != nil {
		log.Fatalf("Error sending request: %v", err)
	}
	defer resp.Body.Close()

	// 7. Check the response
	fmt.Println("Response Status:", resp.Status)
	// You can read resp.Body here if needed
	bodyBytes, _ := io.ReadAll(resp.Body)
	fmt.Println("Response Body:", string(bodyBytes))
}
```

bastiandoetsch: Probably it's slightly off-topic, as this should maybe be handled on a higher application level.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		// try to send request
+		response, err := rm.next.RoundTrip(&localRequest)
+		if err != nil {
+			return nil, err
+		}
+
+		if retryAfterError := shouldRetry(response); retryAfterError != nil {
```
bastiandoetsch, line 66: nitpick: `retryAfterError` is confusing here, as you return from errors above. Probably it should just be `retry` as we analyse response status codes.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+			return nil, retryAfterError
+		}
+
+		// depending on the response determine if we should retry
```
bastiandoetsch, line 70: question: is this comment in the right place?

pkg/networking/middleware/retry_middleware.go
bastiandoetsch: I think we need a map here that is used for whitelisting retries and that contains all retryable error code. Got a generated suggestion [.](https://docs.google.com/spreadsheets/d/1PXMtVy3NYH9v-eM6trfJlXyL6PMQ8g9EXgdv9YKGbYY/edit?gid=1308434658#gid=1308434658)

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+	var retryError error
+
+	if response.StatusCode >= http.StatusInternalServerError ||
+		response.StatusCode == http.StatusTooManyRequests ||
+		response.StatusCode == http.StatusMovedPermanently {
+
+		retryAfterSeconds := 0
+
+		// try to read retry-after header if available
+		if headerRetryAfterValue := response.Header.Get("Retry-After"); len(headerRetryAfterValue) > 0 {
```
bastiandoetsch, line 91: Very nice!

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+			tmp, err := strconv.ParseInt(headerRetryAfterValue, 10, 64)
+			if err == nil {
+				retryAfterSeconds = int(tmp)
+			} else {
+				retryAfterSeconds = defaultRetryAfterSeconds
```
bastiandoetsch, line 96: suggestion: make this configurable, like the retry count.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,108 @@
...
+		response.StatusCode == http.StatusTooManyRequests ||
+		response.StatusCode == http.StatusMovedPermanently {
+
+		retryAfterSeconds := 0
+
+		// try to read retry-after header if available
+		if headerRetryAfterValue := response.Header.Get("Retry-After"); len(headerRetryAfterValue) > 0 {
+			tmp, err := strconv.ParseInt(headerRetryAfterValue, 10, 64)
+			if err == nil {
+				retryAfterSeconds = int(tmp)
```
bastiandoetsch, line 94: nitpick: I'd name that `retryDelaySecs`

pkg/networking/middleware/retry_middleware_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/snyk/go-application-framework/pkg/configuration"
+)
+
+type failRoundtripper struct {
+	actualCount                  uint
```
bastiandoetsch, line 17: Wondering: is the backoff library using the given function in a goroutine or is it synchronous? If it's async, probably these should be types from `atomic` package or protected by a mutex to avoid race tests noping out.

bastiandoetsch (approved):

bastiandoetsch (approved):
left some comments, nothing blocking, though.

bastiandoetsch (approved):
left some comments, nothing blocking, though.

go.mod
PeterSchafer: v4 is used somewhere else in a test script. I would want to use the latest version and maybe in a separate PR remove/change the test script

pkg/networking/middleware/retry_middleware.go
PeterSchafer: Done & Done!

pkg/networking/middleware/retry_middleware.go
PeterSchafer: nope, moved it!

dotkas:
Suggestion: Would some debug logging help or clutter when we retried? Or are the existing logs already good enough to show this?

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,132 @@
...
+		finalError = nil
+	}
+
+	return finalResponse, finalError
+}
+
+func shouldRetry(response *http.Response, defaultRetryDelay int) error {
+	var retryError error
+
+	if statusCodeLookupTable[response.StatusCode] {
```
bastiandoetsch, line 111: nitpick: maybe just call it `statusCodeRetryWhitelist`?

pkg/networking/middleware/retry_middleware.go
```diff
@@ -0,0 +1,169 @@
...
+		return response, nil
+	}
+
+	backoffMethod := backoff.NewExponentialBackOff()
+	backoffMethod.InitialInterval = time.Duration(retryAfterSeconds) * time.Second
+	finalResponse, finalError = backoff.Retry(req.Context(), op, backoff.WithBackOff(backoffMethod), backoff.WithMaxTries(maxAttempts))
+
+	// if retries fail to resolve the issue, we need to unset the locally used error type to not return it from the RoundTripper
+	if errors.Is(finalError, errRetryNecessary) {
+		finalError = nil
```
dotkas, line 123: Will this error still get logged somewhere, if the `op` function returns an error that is not a `shouldRetry` error? Curious if they will get lost in a potential debugging scenario.

dotkas (approved):

pkg/networking/middleware/retry_middleware.go
PeterSchafer: Done!

PR #347, opened 2025-05-13
fix: [CLI-888] allowing populated response in RoundTrip error cases

dotkas:
We're implementing a change in how we handle network requests from the Legacy CLI in order to avoid writing the same logic for both the GAF network stack and the proxied requests from the Legacy CLI.

This requires us to keep the response object from the RoundTripper even on errors, which is a bit unidiomatic, but the best compromise I could find.

Currently, if our `RoundTrip()` logic fails on an error, we do not return the raw response, but only the error (which is probably correct), but in order for us to handle the errors from the legacy CLI correctly, we need to preserve the raw HTTP error and handle it later in the stack.

Relevant details from the dependent [CLI PR](https://github.com/snyk/cli/pull/5888):
> The relevant downstream logic is found in the `goproxy` [code](https://github.com/elazarl/goproxy/blob/088f758167d2ea5cf2a8e14aab45f825ef47399e/https.go#L331-L332) that expects a non-empty response body in order to not re-send a request that was intercepted by a custom handler.

Meaning that we need the intact response object in cases of errors, in order to adequately decorate the error responses from the Legacy CLI as it bubbles up the chain.

### DIFF
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -40,11 +40,6 @@ func (rm ResponseMiddleware) RoundTrip(req *http.Request) (*http.Response, error
 
 	err = rm.errHandler(err, res.Request.Context())
 
-	// RoundTrip should return one or the other.
-	if err != nil {
-		res = nil
-	}
-
 	return res, err
 }
 
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -31,7 +31,6 @@ func NewReponseMiddleware(roundTriper http.RoundTripper, config configuration.Co
 
 func (rm ResponseMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	res, err := rm.next.RoundTrip(req)
-
 	if err != nil {
 		return res, err
 	}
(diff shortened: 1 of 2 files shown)


bastiandoetsch (approved):

pkg/networking/middleware/response_test.go
```diff
@@ -75,9 +75,8 @@ func Test_ResponseMiddleware(t *testing.T) {
 			snykErr := snyk_errors.Error{}
 			url := fmt.Sprintf("%s/%d", server.URL, code)
 			req := buildRequest(url)
-			res, err := rt.RoundTrip(req)
+			_, err := rt.RoundTrip(req)
```
PeterSchafer, line 78: Question: is it right to ignore the response here or should we express the intention that the response is not nil

PeterSchafer (approved):

pkg/networking/middleware/response.go
```diff
@@ -40,11 +40,6 @@ func (rm ResponseMiddleware) RoundTrip(req *http.Request) (*http.Response, error
 
 	err = rm.errHandler(err, res.Request.Context())
 
-	// RoundTrip should return one or the other.
```
PeterSchafer, line 43: Question: Are we sure that the response body is still consumable?

dotkas: At least my tests in CLI worked 🤷🏻‍♂️ Let's talk async.

dotkas: @PeterSchafer check out my latest commit. I didn't add the restored body, as I think we missed it's already happening: https://github.com/snyk/go-application-framework/blob/65a591adf20f92968937a652e43f42958647a308/pkg/networking/middleware/response.go#L83

PeterSchafer (approved):
Final Review

PR #348, opened 2025-05-15
feat(PAT): update authenticators to support PAT

j-luong:


go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/oapi-codegen/oapi-codegen/v2 v2.4.1
 	github.com/oapi-codegen/runtime v1.1.1
+	github.com/oklog/ulid v1.3.1
```
PeterSchafer, line 34: Question: Is this dependency important for the functionality? Can't we just treat the token as a string?

internal/api/personal_access_tokens/2024-03-19/pat.go
```diff
@@ -0,0 +1,187 @@
+package v20240319
```
PeterSchafer, line 1: Question: Is this generated from the OpenAPI spec? It doesn't look like it. Did we consider generating it?

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	if err != nil {
+		return "", fmt.Errorf("failed to decode PAT metadata response: %w", err)
+	}
+
+	if patMetadataBody.Data.Attributes.Hostname == "" {
+		return "", fmt.Errorf("invalid empty hostname")
+	}
+
+	// ensure a match on any of eu.snyk.io, au.snyk.io, us.snyk.io, snyk.io
+	validHostRegex := `^(eu\.snyk\.io|au\.snyk\.io|us\.snyk\.io|snyk\.io)$`
```
PeterSchafer, line 137: Suggestion: Please re-use the existing host validation logic [here](https://github.com/snyk/go-application-framework/blob/main/pkg/auth/oauth2authenticator.go#L486). It is based on a [configured pattern](https://github.com/snyk/go-application-framework/blob/main/pkg/app/app.go#L214) which has been reviewed for correctness.

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	if isValid, err := isValidAuthHost(patMetadataBody.Data.Attributes.Hostname, validHostRegex); err != nil || !isValid {
+		return "", fmt.Errorf("invalid hostname: %s", patMetadataBody.Data.Attributes.Hostname)
```
PeterSchafer, line 139: Question: How about error catalog errors?

internal/api/personal_access_tokens/2024-03-19/pat.go
```diff
@@ -0,0 +1,187 @@
...
+		"personal_access_token": TypePAT,
+	}
+)
+
+// Type represents the different types of PAT.
+type Type int
+
+// ParseType parses a string representing a type and returns a Type if valid.
+// It returns an error if the string representation is not a valid type.
+func ParseType(str string) (t Type, err error) {
```
bastiandoetsch, line 31: are the named return parameters wanted? They can easily lead to unexpected behaviour. See https://www.reddit.com/r/golang/comments/1ajuhlr/weird_behavior_with_named_return_variables/

pkg/app/app.go
```diff
@@ -88,6 +88,14 @@ func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logge
 
 		if len(urlFromOauthToken) > 0 && len(urlFromOauthToken[0]) > 0 {
 			urlString = urlFromOauthToken[0]
+		} else if auth.IsAuthTypePAT(config.GetString(configuration.AUTHENTICATION_TOKEN)) { // derive url from PAT
+			tmp, patEndpointErr := auth.DeriveEndpointFromPAT(config.GetString(configuration.AUTHENTICATION_TOKEN), config, engine.GetNetworkAccess().GetUnauthorizedHttpClient(), config.GetString(configuration.ALTERNATIVE_PAT_SERVICE_URL))
```
bastiandoetsch, line 92: if this is causing an API call, it should probably be cached.

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	}
+	return false
+}
+
+// DeriveEndpointFromPAT makes an authenticated request to the Snyk PAT API and returns the correct Snyk endpoint from the given PAT
+var DeriveEndpointFromPAT = func(token string, config configuration.Configuration, client *http.Client, patServiceUrl string) (string, error) {
+	// TODO: we're implementing a local caching solution here but this will need a proper caching solution
+	// exit early if pat is already cached
+	cached_pat_key := GetCachedPatKeyname(token)
+	cached_pat := config.GetString(cached_pat_key)
```
bastiandoetsch, line 84: might be just naming, but where exactly is the endpoint returned in this case?

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	if len(cached_pat) > 0 {
+		return cached_pat, nil
```
bastiandoetsch, line 86: might be just naming, but where exactly is the endpoint returned in this case?

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	// ensure a match on any of eu.snyk.io, au.snyk.io, us.snyk.io, snyk.io
+	validHostRegex := `^(eu\.snyk\.io|au\.snyk\.io|us\.snyk\.io|snyk\.io)$`
+	if isValid, err := isValidAuthHost(patMetadataBody.Data.Attributes.Hostname, validHostRegex); err != nil || !isValid {
+		return "", fmt.Errorf("invalid hostname: %s", patMetadataBody.Data.Attributes.Hostname)
+	}
+
+	endpoint := fmt.Sprintf("https://api.%s", patMetadataBody.Data.Attributes.Hostname)
+
+	// set the token/endpoint in a cache to avoid infinite loops
+	config.Set(cached_pat_key, endpoint)
```
bastiandoetsch, line 145: yeah, just seems to be a naming issue with the key and the variable.

bastiandoetsch: Hm... this is called from a default function within a configuration.Get(). Do we have potential mutex/locking issues on the configuration here?

pkg/local_workflows/auth_workflow.go
```diff
@@ -71,34 +68,42 @@ func authEntryPoint(invocationCtx workflow.InvocationContext, _ []workflow.Data)
...
+	if len(token) > 0 && auth.IsAuthTypeToken(token) {
+		return auth.AUTH_TYPE_TOKEN
 	}
 
+	// currently the auth workflow defaults to traditional token auth when an IDE environment is detected
+	// this will need to change if IDEs want to invoke this workflow for oauth/PAT
 	integration := config.GetString(configuration.INTEGRATION_NAME)
 	if pgk_utils.IsSnykIde(integration) {
-		return authTypeToken
+		return auth.AUTH_TYPE_TOKEN
```
bastiandoetsch, line 86: this should just be removed

pkg/local_workflows/auth_workflow.go
```diff
@@ -71,34 +68,42 @@ func authEntryPoint(invocationCtx workflow.InvocationContext, _ []workflow.Data)
 		auth.WithLogger(logger),
 	)
 
-	err = entryPointDI(config, logger, engine, authenticator)
+	err = entryPointDI(invocationCtx, logger, engine, authenticator)
 	return nil, err
 }
 
 func autoDetectAuthType(config configuration.Configuration) string {
```
bastiandoetsch, line 75: why is this not detecting oauth?

pkg/local_workflows/auth_workflow_test.go
```diff
@@ -65,43 +74,133 @@ func Test_auth_token(t *testing.T) {
...
+
+		auth.DeriveEndpointFromPAT = func(token string, conf configuration.Configuration, client *http.Client, patServiceUrl string) (string, error) {
+			return mockedPatEndpoint, nil
+		}
+
+		err := entryPointDI(mockInvocationContext, &logger, engine, authenticator)
+		assert.NoError(t, err)
+
+		assert.Empty(t, config.GetString(auth.CONFIG_KEY_OAUTH_TOKEN))
+		assert.Empty(t, config.GetString(configuration.AUTHENTICATION_TOKEN))
```
bastiandoetsch, line 143: Wouldn't we expect authentication to set the PAT into AUTHENTICATION_TOKEN?

internal/api/personal_access_tokens/2024-03-19/pat.go
ShawkyZ: Example for generating client + models from spec: https://github.com/snyk/go-application-framework/blob/b8bef86bf4e6afc3128f887e60b757401c67255c/internal/api/policy/2024-10-15/gen.go#L3
you could follow the same pattern as in internal/api/policy 

fyi: to download the spec, you will need the complete compiled version: 
in the service it's usually located in `internal/api/rest/versions/{VERSION}/spec.yaml`

pkg/app/app.go
ShawkyZ: Question: Why are we not doing this in the defaultFunc of token configuration ? meaning
in the defaultFunc of the Token configuration, we can check there if it's PAT, if it's, check and update API endpoint

pkg/local_workflows/auth_workflow.go
PeterSchafer: I think it does, all the way at the end

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	// TODO: we're implementing a local caching solution here but this will need a proper caching solution
+	// exit early if pat is already cached
+	cached_pat_key := GetCachedPatKeyname(token)
+	cached_pat := config.GetString(cached_pat_key)
+	if len(cached_pat) > 0 {
+		return cached_pat, nil
+	}
+
+	if len(patServiceUrl) == 0 {
+		patServiceUrl = constants.SNYK_DEFAULT_API_URL
```
ShawkyZ, line 90: Question: why not use the endpoint that is in config  (e.g. API_URL) ?

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+		return "", fmt.Errorf("failed to parse Snyk API URL: %w", err)
+	}
+
+	apiUrl.Path = path.Join(apiUrl.Path, patMetadataEndpoint)
+	queryValues := apiUrl.Query()
+	queryValues.Add("version", patAPIVersion)
+	apiUrl.RawQuery = queryValues.Encode()
+	urlString := apiUrl.String()
+
+	req, err := http.NewRequest(http.MethodGet, urlString, nil)
```
ShawkyZ, line 103: This can be replaced with the automatically generated API client from the spec, if possible.

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +59,94 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create request: %w", err)
+	}
+
+	// add auth headers
+	req.Header.Set("Authorization", fmt.Sprintf("token %s", token))
```
ShawkyZ, line 109: why not just use AuthenticatedClient instead? The Roundtripper can handle this ?

pkg/app/app.go
PeterSchafer: > Why are we not doing this in the defaultFunc of token configuration
1. Default functions are called on access, not on set.
2. I would say because it mixes up concerns and spreads logic across different parts of the code base. Currently the intention is to keep the logic around what API URL is used close together.
3. Caching is applied and mitigates the API call overhead after the first call.

pkg/auth/tokenauthenticator.go
PeterSchafer: I think I read somewhere that this logic should also be able to use multiple fallback URLs.

pkg/app/app.go
ShawkyZ, line 92: Question: Why are we not doing this in the defaultFunc of token configuration ? meaning
in the defaultFunc of the Token configuration, we can check there if it's PAT, if it is, check and update API endpoint

pkg/auth/tokenauthenticator.go
PeterSchafer: Yeah, I think it is a naming issue 😆

pkg/auth/tokenauthenticator.go
PeterSchafer: We do the same for OAuth, we use the unauthenticated client at first.

pkg/local_workflows/auth_workflow.go
PeterSchafer: I think it is there for older IDEs, that do not explicitly set the auth type.

pkg/app/app.go
PeterSchafer: > Why are we not doing this in the defaultFunc of token configuration

1. Default functions are called on access, not on set.
2. I would say because it mixes up concerns and spreads logic across different parts of the code base. Currently the intention is to keep the logic around what API URL is used close together.
3. Caching is applied and mitigates the API call overhead after the first call.

internal/api/personal_access_tokens/2024-03-19/pat.go
j-luong: thanks, updated to generate from spec

pkg/auth/tokenauthenticator.go
j-luong: unfortunately I think the [configured pattern](https://github.com/snyk/go-application-framework/blob/main/pkg/app/app.go#L214) is initialised after we try to initialise the default api url.

I could set the regex string as a constant in GAF, so that the pattern is centralised at least?

internal/api/personal_access_tokens/2024-03-19/pat.go
j-luong: [updated](https://github.com/snyk/go-application-framework/pull/348#discussion_r2100701018) to generate the models via the spec

pkg/auth/tokenauthenticator.go
j-luong: I'll clarify whether this is a static url or not

pkg/auth/tokenauthenticator.go
j-luong: > Do we have potential mutex/locking issues on the configuration here?
We shouldn't, we have `snyc.RWMutex` implemented in our configuration

pkg/local_workflows/auth_workflow_test.go
j-luong: The PAT flow should resemble the legacy token flow, and in this case the AUTHENTICATION_TOKEN gets cleared to avoid it being used during the auth flow.

j-luong: To keep the PAT and legacy flows the same:
- we clear the AUTHENTICATION_TOKEN to avoid using it by mistake
- on success, we persist the new token to file, which has a different key to `AUTHENTICATION_TOKEN` (`snyk_token` vs `api`):
```
logger.Print("Set pat credentials in config")
config.PersistInStorage(auth.CONFIG_KEY_TOKEN)
config.Set(auth.CONFIG_KEY_TOKEN, pat)
```

On subsequent runs, this token in file gets set as the `AUTHENTICATION_TOKEN`.

The test checks that `CONFIG_KEY_TOKEN` is set as expected.

pkg/local_workflows/auth_workflow_test.go
```diff
@@ -65,43 +74,133 @@ func Test_auth_token(t *testing.T) {
...
+		engine.EXPECT().GetConfiguration().Return(engineConfig).Times(1)
+
+		expectedErr := errors.New("mocked DeriveEndpointFromPAT error")
+		auth.DeriveEndpointFromPAT = func(token string, conf configuration.Configuration, client *http.Client, patServiceUrl string) (string, error) {
+			return "", expectedErr
+		}
+
+		err := entryPointDI(mockInvocationContext, &logger, engine, authenticator)
+		assert.Error(t, err)
+		assert.Equal(t, expectedErr, err)
```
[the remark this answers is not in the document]
j-luong: The old token would still be persisted in the file as we only update the config once we successfully authenticate/derive the endpoint

I updated the test to include this check

pkg/auth/tokenauthenticator.go
j-luong: > I think I read somewhere that this logic should also be able to use multiple fallback URLs.

this should be added now

> Question: why not use the endpoint that is in config (e.g. API_URL) ?

this logic runs when trying to set the API_URL, so we'd end up in a loop if we use it

pkg/app/app.go
```diff
@@ -88,8 +94,28 @@ func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logge
...
+			if !config.GetBool(is_pat_valid_key) {
+				logger.Warn().Msg("the configured token is not valid; errors expected")
+			} else {
+				regions := config.GetStringSlice(configuration.SNYK_REGION_URLS)
+				supportedPatRegions := auth.FilterSupportedPatRegions(regions, config.GetStringSlice(auth.CONFIG_UNSUPPORTED_PAT_REGIONS))
+				for _, region := range auth.ShuffleStrings(supportedPatRegions) {
+					tmp, patEndpointErr := auth.DeriveEndpointFromPAT(config.GetString(configuration.AUTHENTICATION_TOKEN), config, engine.GetNetworkAccess().GetUnauthorizedHttpClient(), region)
+					if patEndpointErr != nil {
+						logger.Warn().Err(err).Str(configuration.API_URL, urlString).Msg("failed to get api url from pat")
+						config.Set(is_pat_valid_key, false)
```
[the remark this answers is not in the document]
ShawkyZ: We shouldn't always hit the default instance though, right?
Maybe only if current API != default API

Maybe?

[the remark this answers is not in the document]
PeterSchafer: Shall we try to clarify this synchronously?

pkg/app/app.go
```diff
@@ -77,7 +78,12 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 	return callback
 }
 
-func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logger) configuration.DefaultValueFunction {
+func defaultFuncApiUrl(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger) configuration.DefaultValueFunction {
+	// TODO: implement proper caching
+	// cache check needed in case an invalid PAT is already configured in env var or config file
+	is_pat_valid_key := fmt.Sprintf("%s_%s", auth.CACHED_PAT_IS_VALID_KEY_PREFIX, config.GetString(configuration.AUTHENTICATION_TOKEN))
+	// initial state assumes it's valid
+	config.Set(is_pat_valid_key, true)
```
PeterSchafer, line 86: Suggestion: after having the config caching in GAF, let's remove this caching workaround.

pkg/local_workflows/auth_workflow.go
```diff
@@ -111,7 +120,36 @@ func entryPointDI(config configuration.Configuration, logger *zerolog.Logger, en
...
+		instrumentation.AddExtension(authTypeParameter, auth.AUTH_TYPE_PAT)
+
+		pat := config.GetString(ConfigurationNewAuthenticationToken)
+
+		logger.Printf("Unset oauth key %q from config", auth.CONFIG_KEY_OAUTH_TOKEN)
+		config.Unset(auth.CONFIG_KEY_OAUTH_TOKEN)
+
+		config.Set(configuration.AUTHENTICATION_TOKEN, "") // clear token to avoid using it during authentication
+
+		regions := config.GetStringSlice(configuration.SNYK_REGION_URLS)
```
PeterSchafer, line 133: Suggestion: This logic is a duplication of the default function changes, let's try to apply DRY here please.

pkg/local_workflows/config_workflow.go
```diff
@@ -23,12 +23,12 @@ const (
 	defaultRegion         = "snyk-us-01"
 )
 
-var regions = []SnykRegion{
-	{alias: "snyk-us-01", url: "https://api.snyk.io"},
-	{alias: "snyk-us-02", url: "https://api.us.snyk.io"},
-	{alias: "snyk-au-01", url: "https://api.au.snyk.io"},
-	{alias: "snyk-eu-01", url: "https://api.eu.snyk.io"},
-	{alias: "snyk-gov-01", url: "https://api.snykgov.io"},
+var Regions = []SnykRegion{
```
PeterSchafer, line 26: Praise: Nice!

pkg/auth/authHost.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return r.MatchString(authHost), nil
+}
+
+// FilterSupportedPatRegions iterates a list of region URLs and filters out regions not supported by PAT.
+// It returns a filtered slice of URL strings
+func FilterSupportedPatRegions(regions []string, unsupportedRegions []string) []string {
```
PeterSchafer, line 49: Suggestion: Please write some Unit tests. I assume that this logic can be simplified but without unit tests it is difficult to fully follow what the intention is.

pkg/auth/authHost.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		formattedRegion := fmt.Sprintf("%s://%s", scheme, host)
+		filteredRegions = append(filteredRegions, formattedRegion)
+	}
+	return filteredRegions
+}
+
+// ShuffleStrings takes []string, shuffles the elements and returns a new []string with the shuffled elements
+func ShuffleStrings(originalSlice []string) []string {
```
PeterSchafer, line 74: Suggestion: Please write some Unit tests

PeterSchafer: If possible please link to the name of this algorithm

andrewrobinsonhodges-snyk (approved):
Looks good overall.

I don't think we still need the logic around shuffling and filtering endpoints any more, so we should probably remove this to keep the flow as simple as possible.

A few minor comments on naming as well.

pkg/auth/authHost.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		formattedRegion := fmt.Sprintf("%s://%s", scheme, host)
+		filteredRegions = append(filteredRegions, formattedRegion)
+	}
+	return filteredRegions
+}
+
+// ShuffleStrings takes []string, applies the Fisher-Yates shuffle algorithm and returns a new []string with the shuffled elements
+func ShuffleStrings(originalSlice []string) []string {
```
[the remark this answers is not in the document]
j-luong: no I don't think so, will remove

pkg/auth/authHost.go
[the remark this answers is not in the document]
j-luong: I think so, we provide the capability to set the target endpoint, and as it's a requirement to not use certain endpoints, we'll need this filter

pkg/auth/tokenauthenticator.go
```diff
@@ -38,3 +66,116 @@ func (t *tokenAuthenticator) AddAuthenticationHeader(request *http.Request) erro
...
+		err      error
+		errs     error
+		endpoint string
+	)
+
+	supportedPatRegions := FilterSupportedPatRegions(regions, UnsupportedPatRegions)
+	shuffledRegions := ShuffleStrings(supportedPatRegions)
+
+	// include the default region as a fallback
+	if !slices.Contains(shuffledRegions, constants.SNYK_DEFAULT_API_URL) {
```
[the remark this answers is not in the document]
j-luong: This would be needed to ensure that we always have the default region as a fallback. This covers the case when a user manually sets a region, we'll add the default region in this line to ensure we use it in case the user set region fails

pkg/configuration/constants.go
```diff
@@ -47,6 +47,8 @@ const (
 	PREVIEW_FEATURES_ENABLED       string = "internal_preview_features_enabled"   // boolean indicates if preview features shall be enabled
 	UNKNOWN_ARGS                   string = "internal_unknown_arguments"          // arguments unknown to the current application but maybe relevant for delegated application calls
 	IN_MEMORY_THRESHOLD_BYTES      string = "internal_in_memory_threshold_bytes"  // threshold to determine where to store workflow.Data
+	SNYK_REGION_URLS               string = "internal_snyk_region_urls"           // comma delimited string of snyk region URLs e.g. `https://api.snyk.io,https://api.eu.snyk.io'
```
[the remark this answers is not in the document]
j-luong: it's currently required for testing, it is set as an internal env var; which are env vars not intended for general use and so are undocumented.

pkg/local_workflows/auth_workflow_test.go
j-luong: Currently the `configuration.AUTHENTICATION_TOKEN` is for the legacy CLI token; we decided to use the same configuration value for PATs as they can be used interchangeably in most cases.

It's probably scope creep updating the naming, as it may have unintended side effects which need to be investigated.

As mentioned, a lot of the awkward naming is a result of how it is in the legacy CLI and the need to be backwards compatible.

pkg/app/app.go
```diff
@@ -88,6 +88,19 @@ func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logge
...
+				regions = []string{val}
+			}
+			apiUrl, endpointErr := auth.DeriveEndpointFromPat(config.GetString(configuration.AUTHENTICATION_TOKEN), config, engine.GetNetworkAccess().GetUnauthorizedHttpClient(), regions)
+			if endpointErr != nil {
+				logger.Warn().Err(endpointErr).Msg("failed to get api url from pat")
+			}
+			if len(apiUrl) > 0 {
+				urlString = apiUrl
+			}
 		} else if existingValue != nil { // configured value takes precedence
```
dotkas, line 104: thought: I think this comment is not true anymore. It is more the last resort if all else fails.

pkg/app/app.go
```diff
@@ -88,6 +88,19 @@ func defaultFuncApiUrl(config configuration.Configuration, logger *zerolog.Logge
 
 		if len(urlFromOauthToken) > 0 && len(urlFromOauthToken[0]) > 0 {
```
dotkas, line 89: suggestion: From a UX perspective I'm a bit unhappy with the logical flow here, as the warning on L86 (can't link it 🤷🏻 ) will always show up in logs of customers using PATs. The general fallback logic seems to not really favor early returns or breaks. Is it possible to refactor this a bit to avoid warning-spams?

Also it seems to just continue even if the PAT-part fails, causing an error further down the execution tree. I wonder if an earlier authentication error would be nicer and cause easier debugging for us in the future.

dotkas (approved):
Approving to unblock, LGTM, just some comments.

PR #350, opened 2025-05-20
chore: human readable prompt enhancements

CatalinSnyk:


pkg/local_workflows/local_models/transform_test.go
```diff
@@ -55,7 +55,7 @@ func Test_mapSuppressions(t *testing.T) {
 				Id: uuidPtr(validUUID1),
 				Details: &TypesSuppressionDetails{
 					Category:   "testCategory",
-					Expiration: "never",
+					Expiration: "does not expire",
```
PeterSchafer, line 58: Question: should this re-use the constant `DefaultSuppressionExpiry`?

pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -90,7 +103,15 @@ func ignoreCreateWorkflowEntryPoint(invocationCtx workflow.InvocationContext, _
 	}
 
 	interactive := config.GetBool(InteractiveKey)
-	addCreateIgnoreDefaultConfigurationValues(invocationCtx, interactive)
+	faint := lipgloss.NewStyle().Faint(true)
```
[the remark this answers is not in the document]
PeterSchafer: There are multiple such changes in the PR, @CatalinSnyk let's talk about them and brainstorm different solutions.

pkg/local_workflows/testdata/sarif-snyk-goof-ignores.json
```diff
@@ -2066,8 +2066,7 @@
               "severity": "high",
               "originalSeverity": "low"
             }
-          },
-          "suppressions": null
+          }
```
PeterSchafer, line 2069: Question: Why was this removed? I think I'm missing some context.

PeterSchafer (changes requested):
I stopped reviewing the PR for now until we resolve the mentioned issue.

pkg/local_workflows/testdata/sarif-snyk-goof-ignores.json
PeterSchafer: Did you double check that this is not a the way the API returns the data, just saying, it could well be that the data from the API looks like this and by removing the `null` now, we might miss a problem in the future.

PeterSchafer: Great, thank you for sharing. Let's undo the change for now please, as this is an unrelated change.

pkg/local_workflows/testdata/sarif-snyk-goof-ignores.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/schemas/sarif-schema-2.1.0.json",
+  "$schema": "https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/schemas/sarif-schema-2.1.0.json",
```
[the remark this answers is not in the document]
CatalinSnyk: Both schemas define the suppression type as array, so the warning was there even before the schema change.

pkg/local_workflows/ignore_workflow/config.go
```diff
@@ -83,71 +76,106 @@ func defaultFuncWithValidator(existingValue interface{}, userInterface ui.UserIn
...
+		return "", err
+	}
+
+	input, err := userInterface.Input(prompt)
+	if err != nil {
+		return "", err
 	}
 
-	input, err := userInterface.Input(inputDescription)
+	err = userInterface.Output("") // new line between prompts
```
PeterSchafer, line 98: Suggestion: This seems a bit wrong, trying to keep the mindset of a view unaware implementation. This clearly relates to a line based output. Why can't the spaces be part of the formatting of the prompt?

PeterSchafer (approved):
Nice Work! Pre-Approving assuming the two remaining comments are straight forward to apply.

PeterSchafer:
Maybe wait with merging this and let us finish the review of https://github.com/snyk/go-application-framework/pull/343 first. Thank you!

pkg/local_workflows/ignore_workflow/config.go
```diff
@@ -83,71 +69,125 @@ func defaultFuncWithValidator(existingValue interface{}, userInterface ui.UserIn
...
+	}
+
 	return input, nil
 }
 
 func isValidIgnoreType(input interface{}) error {
 	ignoreType, ok := input.(string)
 	if !ok {
-		return fmt.Errorf("invalid ignore type, expected string")
+		return cli.NewValidationFailureError("Ignore type must be a string. Ensure the provided value is a string.")
```
j-luong, line 107: question: why do we accept `interface{}` and then type check? We should accept a string param instead.
`func isValidIgnoreType(input string)`

CatalinSnyk: Change this to be of type string, since all the flags defined for the workflow are strings, and now we avoid multiple type castings.

PR #351, opened 2025-05-21
fix: WIP make auth cancellable [IDE-962]

rrama:
Previously you had to wait for the auth request to timeout. This change allows the caller to decide to cancel the current request on demand.

The linter complained the function was too long, so I had to split it.
Which at the same time I also prevented the function opening multiple browser windows if the listening HTTP server fails for an unexpected reason.

### DIFF
--- a/pkg/auth/authenticator.go
+++ b/pkg/auth/authenticator.go
@@ -1,6 +1,7 @@
 package auth
 
 import (
+	"context"
 	"net/http"
 
 	"github.com/snyk/go-application-framework/internal/api"
@@ -11,7 +12,8 @@ import (
 
 type Authenticator interface {
 	// Authenticate authenticates the user and returns an error if the authentication failed.
-	Authenticate() error
+	// Takes a context which can be used to interrupt the authentication.
+	Authenticate(ctx context.Context) error
 	// AddAuthenticationHeader adds the authentication header to the request.
 	AddAuthenticationHeader(request *http.Request) error
 	// IsSupported returns true if the authenticator is ready for use.
--- a/pkg/auth/oauth2authenticator.go
+++ b/pkg/auth/oauth2authenticator.go
@@ -262,20 +262,19 @@ func (o *oAuth2Authenticator) persistToken(token *oauth2.Token) error {
 	return nil
 }
 
-func (o *oAuth2Authenticator) Authenticate() error {
+func (o *oAuth2Authenticator) Authenticate(ctx context.Context) error {
 	var err error
 
 	if o.grantType == ClientCredentialsGrant {
-		err = o.authenticateWithClientCredentialsGrant()
+		err = o.authenticateWithClientCredentialsGrant(ctx)
 	} else {
-		err = o.authenticateWithAuthorizationCode()
+		err = o.authenticateWithAuthorizationCode(ctx)
 	}
 
 	return err
 }
 
-func (o *oAuth2Authenticator) authenticateWithClientCredentialsGrant() error {
-	ctx := context.Background()
+func (o *oAuth2Authenticator) authenticateWithClientCredentialsGrant(ctx context.Context) error {
 	config := getOAuthConfigurationClientCredentials(o.oauthConfig)
 
 	// Use the custom HTTP client when requesting a token.
@@ -293,7 +292,7 @@ func (o *oAuth2Authenticator) authenticateWithClientCredentialsGrant() error {
 	return err
 }
 
-func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
+func (o *oAuth2Authenticator) authenticateWithAuthorizationCode(ctx context.Context) error {
 	var responseCode string
 	var responseState string
 	var responseError string
@@ -308,7 +307,6 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
 	}
 	state := string(stateBytes)
 	codeChallenge := getCodeChallenge(verifier)
-	ctx := context.Background()
 
 	if o.headless {
 		return errors.New("headless mode not supported")
@@ -359,6 +357,14 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
 		// launch browser
 		go o.openBrowserFunc(authCodeUrl)
 
+		// Handle context cancellation
+		canceled := false
+		go func() {
+			<-ctx.Done()
+			canceled = true
+			o.shutdownServerFunc(srv)
+		}()
+
 		timedOut := false
 		timer := time.AfterFunc(TIMEOUT_SECONDS, func() {
 			timedOut = true
@@ -366,6 +372,9 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
 		})
 		listenErr = srv.Serve(listener)
 		if errors.Is(listenErr, http.ErrServerClosed) { // if the server was shutdown normally, there is no need to iterate further
+			if canceled {
+				return nil // No need to error, the user canceled this auth request
+			}
 			if timedOut {
 				return errors.New("authentication failed (timeout)")
 			}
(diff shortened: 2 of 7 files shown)


pkg/auth/authenticator.go
```diff
@@ -11,7 +12,8 @@ import (
 
 type Authenticator interface {
 	// Authenticate authenticates the user and returns an error if the authentication failed.
-	Authenticate() error
+	// Takes a context which can be used to interrupt the authentication.
+	Authenticate(ctx context.Context) error
```
bastiandoetsch, line 16: Hm. For backwards compatibility, it might be good to have a new interface `CancellableAuthenticator`. As this is a library, exported functions and interfaces should ideally not be changed if not needed.

pkg/auth/oauth2authenticator.go
```diff
@@ -262,20 +262,19 @@ func (o *oAuth2Authenticator) persistToken(token *oauth2.Token) error {
 	return nil
 }
 
-func (o *oAuth2Authenticator) Authenticate() error {
+func (o *oAuth2Authenticator) Authenticate(ctx context.Context) error {
```
bastiandoetsch, line 265: Same here - this should most likely be wrapped, as several clients using the library will be affected by a signature change of an exported function.

pkg/auth/oauth2authenticator.go
```diff
@@ -323,21 +321,67 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
 		} else {
 			appUrl := o.config.GetString(configuration.WEB_APP_URL)
 			responseCode = html.EscapeString(r.URL.Query().Get("code"))
 			responseState = html.EscapeString(r.URL.Query().Get("state"))
 			responseInstance = html.EscapeString(r.URL.Query().Get("instance"))
+			o.logger.Debug().
+				Str("responseCode", responseCode).
+				Str("responseState", responseState).
+				Str("responseInstance", responseInstance).
+				Msg("OAuth2 response callback received, re-directing user to /authenticated")
```
bastiandoetsch, line 336: Not sure this should be here. this would be output in the CLI debug logs, too. A library IMHO should minimize logging.

pkg/auth/oauth2authenticator.go
```diff
@@ -359,50 +403,40 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
 		listenErr = srv.Serve(listener)
 		if errors.Is(listenErr, http.ErrServerClosed) { // if the server was shutdown normally, there is no need to iterate further
+			if canceled {
+				o.logger.Info().Msg("OAuth2 request canceled.")
+				return nil, false // No need to error, the user canceled this auth request
+			}
 			if timedOut {
-				return errors.New("authentication failed (timeout)")
+				o.logger.Warn().Msg("OAuth2 request timed out.")
+				return errors.New("authentication failed (timeout)"), false
```
bastiandoetsch, line 428: how is that set? I don't see an assignment to timeOut anywhere?

pkg/auth/oauth2authenticator.go
```diff
@@ -359,50 +403,40 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
+				canceled = true
+			}
 			o.shutdownServerFunc(srv)
-		})
+		}()
+
 		listenErr = srv.Serve(listener)
 		if errors.Is(listenErr, http.ErrServerClosed) { // if the server was shutdown normally, there is no need to iterate further
+			if canceled {
+				o.logger.Info().Msg("OAuth2 request canceled.")
```
bastiandoetsch, line 423: could also be a timeout - the context could have been canceled or the timeout triggered.

pkg/auth/oauth2authenticator.go
```diff
@@ -359,50 +403,40 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
-			timedOut = true
+		canceled := false
+		go func() {
+			<-ctx.Done()
+			switch err := ctx.Err(); {
+			case errors.Is(err, context.DeadlineExceeded):
+				timedOut = true
+			case errors.Is(err, context.Canceled):
+				canceled = true
+			}
```
bastiandoetsch, line 416: should this maybe just be executed after srv.Serve and not in a go func? I see a potential race for timeout/canceled.

pkg/auth/oauth2authenticator.go
```diff
@@ -417,6 +451,11 @@ func writeCallbackErrorResponse(w http.ResponseWriter, q url.Values, responseErr
 		Description: html.EscapeString(q.Get("error_description")),
 	}
 
+	o.logger.Error().
+		Str("responseError", responseError).
+		Str("responseErrorDescription", data.Description).
+		Msg("Received an error in the OAuth2 callback request.")
+
```
bastiandoetsch, line 458: Not sure, we should log here.

pkg/auth/oauth2authenticator.go
rrama: Done.

pkg/auth/oauth2authenticator.go
```diff
@@ -359,47 +408,35 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
+			case errors.Is(err, context.DeadlineExceeded):
+				timedOut.Store(true)
+			case errors.Is(err, context.Canceled):
+				canceled.Store(true)
+			}
+			// We impose that o.shutdownServerFunc is/calls a server function that is safe to be called before srv.Server
+			// and will result in srv.Server exiting immediately when called.
 			o.shutdownServerFunc(srv)
-		})
+		}()
```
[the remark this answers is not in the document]
rrama: I think the bot is confused, it maybe hasn't understood the diff, as the additions (ignoring the incorrect deletions) add back the same code I wrote.

pkg/auth/oauth2authenticator.go
```diff
@@ -359,47 +408,35 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
+		var timedOut atomic.Bool
+		var canceled atomic.Bool
+		go func() {
+			<-ctx.Done()
+			switch err := ctx.Err(); {
+			case errors.Is(err, context.DeadlineExceeded):
+				timedOut.Store(true)
+			case errors.Is(err, context.Canceled):
+				canceled.Store(true)
+			}
```
bastiandoetsch, line 421: why isn't that done in the error handling below? That way we would have error handling in one place.

pkg/auth/oauth2authenticator_test.go
```diff
@@ -475,3 +477,56 @@ func Test_Authenticate_AuthorizationCode(t *testing.T) {
...
+func Test_CancellableAuthenticate_AuthorizationCodeGrant_ContextCanceled(t *testing.T) {
+	t.Parallel()
+
+	// Setup
+
+	config := configuration.NewInMemory()
+	config.Set(configuration.FF_OAUTH_AUTH_FLOW_ENABLED, true)
+
+	authCtx, cancelAuthCtx := context.WithCancel(context.Background())
+	defer cancelAuthCtx() // For safety to prevent resource leaks
```
bastiandoetsch, line 490: why not use t.Cleanup?

rrama: Done.

pkg/auth/oauth2authenticator_test.go
```diff
@@ -475,3 +477,56 @@ func Test_Authenticate_AuthorizationCode(t *testing.T) {
...
+	var authCtxCanceled atomic.Bool
+
+	mockMux := http.NewServeMux()
+	mockMux.HandleFunc("/oauth2/authorize", func(w http.ResponseWriter, r *http.Request) {
+		cancelAuthCtx()
+		authCtxCanceled.Store(true)
```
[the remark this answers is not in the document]
bastiandoetsch: Hm. Not sure that I understand the code then. I was referring to line 531.

pkg/auth/oauth2authenticator.go
```diff
@@ -359,47 +407,27 @@ func (o *oAuth2Authenticator) authenticateWithAuthorizationCode() error {
...
-		})
+		}()
+
 		listenErr = srv.Serve(listener)
 		if errors.Is(listenErr, http.ErrServerClosed) { // if the server was shutdown normally, there is no need to iterate further
-			if timedOut {
-				return errors.New("authentication failed (timeout)")
+			switch err := ctx.Err(); {
+			case errors.Is(err, context.Canceled):
+				return false, nil // No need to error, the user canceled this auth request
```
PeterSchafer, line 422: Question: Shouldn't we return an error here instead of swallowing it in the library? If the application decides to swallow it, it can but from a framework perspective we shouldn't take the decision.

pkg/auth/oauth2authenticator_test.go
```diff
@@ -111,7 +112,7 @@ func Test_getToken(t *testing.T) {
 	expectedTokenString, err := json.Marshal(expectedToken)
 	assert.NoError(t, err)
 
-	config := configuration.NewInMemory()
+	config := configuration.NewWithOpts()
```
PeterSchafer, line 115: Praise: Thank you!

PR #352, opened 2025-05-27
feat: add configuration caching

PeterSchafer:
This PR introduces optional configuration caching to avoid or reduce repeated execution of default functions, which can have a high costs in form of API requests etc.

The caching will be per default disabled to remain the current behaviour and needs to be enabled and configured in the applications. Enabling caching has to happen initially when creating a config with `WithCachingEnabled()` but can be adapted at runtime since the configuration values are also available as configuration values themselves.

Within this PR I added some small helpers like `toBool()` and `toDuration()`


pkg/configuration/configuration.go
```diff
@@ -741,3 +787,66 @@ func (ev *extendedViper) GetFiles() []string {
...
+	enabled := true
+	duration := time.Duration(0)
+	if disabledRaw != nil {
+		disabled, boolError := toBool(disabledRaw)
+		enabled = boolError == nil && !disabled
+	}
+
+	if durationRaw != nil {
+		duration, err = toDuration(durationRaw)
+		enabled = err == nil
```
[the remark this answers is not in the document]
PeterSchafer: Decided to return the error, even if it seems quite harsh. But on the other hand it reflects a fundamental configuration issue that could lead to subtle issues later on.

pkg/configuration/configuration.go
```diff
@@ -248,6 +265,9 @@ func readConfigFilesIntoViper(files []string, config *extendedViper) {
 
 // Clone creates a copy of the current configuration.
 func (ev *extendedViper) Clone() Configuration {
+	cacheDuration := ev.GetDuration(CONFIG_CACHE_DURATION)
+	cacheCleanupInterval := ev.GetDuration(CONFIG_CACHE_CLEANUP_INTERVAL)
```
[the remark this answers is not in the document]
PeterSchafer: Added a comment to explain why this is required.

pkg/configuration/configuration.go
```diff
@@ -11,7 +11,9 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
+	"github.com/patrickmn/go-cache"
```
dotkas, line 16: Do we really need an external library for this?

pkg/configuration/configuration.go
```diff
@@ -43,6 +45,8 @@ type Configuration interface {
 	GetStringSlice(key string) []string
 	GetBool(key string) bool
 	GetBoolWithError(key string) (bool, error)
+	GetDuration(key string) time.Duration
```
bastiandoetsch, line 48: nitpicking: I'd create a new interface to remain backwards compatible. Something like `CacheableConfiguration`. I'd also rename the interface func to `GetCacheDuration`. `GetDuration` feels to generic to me.

pkg/configuration/configuration.go
```diff
@@ -298,6 +326,9 @@ func (ev *extendedViper) Set(key string, value interface{}) {
 	localStorage := ev.storage
 	isPersisted := ev.persistedKeys[key]
 	ev.viper.Set(key, value)
+	if ev.defaultCache != nil {
+		ev.defaultCache.Delete(key)
```
bastiandoetsch, line 330: I'm wondering if this is correct. Maybe it would be better to update the cache with the new value?

pkg/configuration/configuration.go
PeterSchafer: This function is a generic getter and has nothing specific todo with Cache. It is like all the other getters, it gives you out of the box support for accessing duration type values.

pkg/configuration/configuration.go
```diff
@@ -741,3 +795,77 @@ func (ev *extendedViper) GetFiles() []string {
...
+
+func (ev *extendedViper) setCache(c *cache.Cache) {
+	ev.mutex.Lock()
+	defer ev.mutex.Unlock()
+	ev.defaultCache = c
+}
+
+// return true if enabled and a caching duration
+func (ev *extendedViper) getCacheSettings() (bool, time.Duration, error) {
+	if ev.defaultCache == nil {
```
bastiandoetsch, line 807: would this need a read lock?

pkg/configuration/configuration.go
PeterSchafer: > Do we really need an external library for this?
It's a question of appetite I guess. We can for sure implement this ourselves but then the effort is much higher.

> Do we really need an external library for this?
I used our internal directory to look up the different modules already in use and cross-referenced it with a deep research into stability and reliability and `go-cache` was the outcome. I personally don't have a bias, I didn't know of imcache until you mentioned it. From the docs, it seems possible to replace now or even later, as it is somehow an implementation detail.

pkg/configuration/configuration.go
```diff
@@ -84,6 +90,7 @@ type extendedViper struct {
 	mutex               sync.RWMutex
 	automaticEnvEnabled bool
 	configFiles         []string
+	defaultCache        *cache.Cache
```
j-luong, line 93: question: wouldn't it be better to define our own cache interface here? This seems quite tightly coupled with the caching package we're using.

pkg/configuration/constants.go
```diff
@@ -36,6 +36,9 @@ const (
 	RAW_CMD_ARGS                   string = "internal_raw_cmd_args"
 	WEB_APP_URL                    string = "internal_snyk_app"
 	MAX_THREADS                    string = "internal_max_thread_count"
+	CONFIG_CACHE_DURATION          string = "internal_config_cache_ttl"
+	CONFIG_CACHE_CLEANUP_INTERVAL  string = "internal_config_cache_cleanup_interval"
```
j-luong, line 40: question: I wonder why a cleanup interval needs to be specified? Can we not rely solely on the TTL? Is this an implementation detail of the caching package?

IMO it makes sense to define a cache size and evict items based on LRU for example.

pkg/configuration/configuration.go
```diff
@@ -165,6 +172,16 @@ func WithFiles(files ...string) Opts {
 	}
 }
 
+func WithCachingEnabled(cacheDuration time.Duration, cleanupInterval time.Duration) Opts {
```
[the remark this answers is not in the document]
PeterSchafer: Good point! This is actually already possible and I think we might want to use this in the CLI, as it runs a command and finishes. I will add docs to explain how to cache for ♾️

pkg/configuration/constants.go
j-luong: I was thinking that things like LRU and cleanup intervals are methods to manage the size of the cache, and so defining a limitation on the cache, like max memory size might make more sense.

pkg/configuration/configuration.go
```diff
@@ -248,6 +265,11 @@ func readConfigFilesIntoViper(files []string, config *extendedViper) {
 
 // Clone creates a copy of the current configuration.
 func (ev *extendedViper) Clone() Configuration {
+	// these lookups need to happen outside the critical section to avoid dealocks
+	cacheDuration := ev.GetDuration(CONFIG_CACHE_DURATION)
+	cacheCleanupInterval := ev.GetDuration(CONFIG_CACHE_CLEANUP_INTERVAL)
+
+	// start critical section
```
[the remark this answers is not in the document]
PeterSchafer: I extended the comment a bit to be more explicit, I hope it helps!

dotkas (approved):

PR #353, opened 2025-05-28
feat: DGP-212 - pass in a client logger

bsalomon-snyk:
NewTestClient() accepts a logger for use by the high- and low-level test API clients.  Defaults to zerolog.Logger{} if not provided.

The logger is currently used in one case where the TestClient wants to log that an expected Job endpoint given by the server has returned a 404, but it wants to keep polling in case the URL becomes available soon after.

Note: another PR is coming that passes NewTestClient() an options list, which will include a "WithLogger()".


pkg/apiclients/testapi/testapi.go
```diff
@@ -150,9 +153,17 @@ func NewTestClient(serverBaseUrl string, cfg Config, opts ...ClientOption) (Test
 		cfg.APIVersion = DefaultAPIVersion
 	}
 
+	var clLogger zerolog.Logger
```
PeterSchafer, line 156: Question: Why not keep pointers here?

pkg/apiclients/testapi/testapi_test.go
```diff
@@ -711,6 +713,99 @@ func assertTestOutcomePass(t *testing.T, result testapi.TestResult, expectedTest
...
+
+		// Assert log output
+		logOutput := logBuffer.String()
+		assert.Contains(t, logOutput, "Job polling returned 404 Not Found")
+		assert.Contains(t, logOutput, fmt.Sprintf(`"jobID":"%s"`, expectedJobIDInLog.String()))
+		assert.Contains(t, logOutput, fmt.Sprintf(`"orgID":"%s"`, orgID.String()))
+		assert.Contains(t, logOutput, `"level":"warn"`)
+	})
+
+	t.Run("default logger is used when none provided", func(t *testing.T) {
```
PeterSchafer, line 788: Question: Isn't this case covered by the tests that are already here?

pkg/apiclients/testapi/testapi_test.go
```diff
@@ -711,6 +713,99 @@ func assertTestOutcomePass(t *testing.T, result testapi.TestResult, expectedTest
...
+		// pollJobToCompletion will encounter the 404 from our mock, log the warning, and continue polling.
+		// The Wait() call will likely terminate due to context deadline.
+		waitCtx, cancelWait := context.WithTimeout(ctx, 2*time.Second)
+		defer cancelWait()
+
+		_ = handle.Wait(waitCtx) //nolint:errcheck // error expected if context times out
+
+		// Assert log output
+		logOutput := logBuffer.String()
+		assert.Contains(t, logOutput, "Job polling returned 404 Not Found")
```
PeterSchafer, line 782: Suggestion: testing the exact content of a debug log can increase the maintenance effort when debug messages are changed. How about reducing the strictness to just check that "404 Not Found" if testing this at all 😅

PeterSchafer:
> Note: another PR is coming that passes NewTestClient() an options list, which will include a "WithLogger()".

Question: Why don't you just already do this change in this PR? There are ClientOptions already in NewTestClient().

pkg/apiclients/testapi/testapi_test.go
bsalomon-snyk: Very true.  It was on the idea of having an explicit test for it that's not dependent on anything else changing, but basically every test will cover it so I'll take it out.

pkg/apiclients/testapi/testapi_test.go
bsalomon-snyk: ok, but i did want to keep something in there and this is the only log written currently. It could assert that any string is written but that seemed vague for the specific error the test generates.

bsalomon-snyk: we chatted elsewhere, but basically the options were split between high-level/low-level and merging them will be a big enough change to merit a new PR.

PeterSchafer (approved):

PR #354, opened 2025-05-30
chore(testapi): update oapi def

hideandsnyk:
This change updates the openapi test client definition for interacting with the test api shim.


pkg/apiclients/testapi/2024-10-15/spec.yaml
```diff
@@ -801,33 +782,32 @@ components:
         Test subject representing a source tree located in a Git repository that
         has a Snyk SCM integration.
       properties:
-        asset_id:
-          description: Asset associated with the tested content.
+        commit_id:
```
PeterSchafer, line 785: Question: Is the commit id really a uuid? seems like a mistake to me, shouldn't it be just a string?

pkg/apiclients/testapi/oapi.gen.go
```diff
@@ -1,6 +1,6 @@
 // Package testapi provides primitives to interact with the openapi HTTP API.
 //
-// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.4.1 DO NOT EDIT.
+// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version (devel) DO NOT EDIT.
```
PeterSchafer, line 3: Question: Is it intentional that the oapi-codegen version is `devel`? not sure if this has an effect though

go.mod
```diff
@@ -108,6 +108,7 @@ require (
 	github.com/xanzy/ssh-agent v0.3.3 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
+	go.uber.org/mock v0.5.2 // indirect
```
PeterSchafer, line 111: Question: Could it be that you are using a different set of tools in your local toolchain?

go.sum
```diff
@@ -294,6 +294,8 @@ github.com/xeipuuv/gojsonschema v1.2.0 h1:LhYJRs+L4fBtjZUfuSZIKGeVu0QRy8e5Xi7D17
 github.com/xeipuuv/gojsonschema v1.2.0/go.mod h1:anYRn/JVcOK2ZgGU+IjEV4nwlhoK5sQluxsYJ78Id3Y=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
+go.uber.org/mock v0.5.2 h1:LbtPTcP8A5k9WPXj54PPPbjcI4Y6lhyOZXn+VS7wNko=
```
bsalomon-snyk, line 297: "github.com/golang/mock" is already included in the GAF, so it's good if we can reuse that instead of adding more packages.  If you set your local package to the golang one then your toolchain will use it instead of adding Uber's.

go.mod
hideandsnyk: Great call thank you.

PR #359, opened 2025-06-06
feat: DGP-391 - expose local policy through the Test API Client

bsalomon-snyk:
### What does this do?

- Allows Test API client users to set a local policy that is already available in the backend API spec.  This includes
  - RiskScoreThreshold
  - SeverityThreshold
  - SuppressPendingIgnores

- validates risk score threshold in range 0-1000
- updates unit tests to set policy and validate risk score threshold range

Jira: [DGP-391](https://snyksec.atlassian.net/browse/DGP-391)

### DIFF
--- a/pkg/apiclients/testapi/testapi.go
+++ b/pkg/apiclients/testapi/testapi.go
@@ -148,8 +148,9 @@ var (
 
 // StartTestParams defines parameters for the high-level StartTest function.
 type StartTestParams struct {
-	OrgID   string
-	Subject TestSubjectCreate
+	OrgID       string
+	Subject     TestSubjectCreate
+	LocalPolicy *LocalPolicy
 }
 
 // testResult is the concrete implementation of the TestResult interface for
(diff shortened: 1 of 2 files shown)


pkg/apiclients/testapi/testapi_test.go
```diff
@@ -52,13 +52,29 @@ func Test_StartTest_Success(t *testing.T) {
 
 	// Create a DepGraph to test
 	testSubject := newDepGraphTestSubject(t)
-	params := testapi.StartTestParams{OrgID: orgID.String(), Subject: testSubject}
+
+	// Define LocalPolicy
+	riskScoreThreshold := uint16(750)
```
PeterSchafer, line 57: Question: Is the narrowed typing here helpful? It seems a bit unnecessary limiting and would require many changes in the future to extend. Why not stick to `int` here? Since the data transfer uses json, there is no saving in storage using either number type.

pkg/apiclients/testapi/testapi.go
```diff
@@ -231,6 +232,13 @@ func NewTestClient(serverBaseUrl string, options ...ConfigOption) (TestClient, e
 // Create the initial test and return a handle to poll it
 func (c *client) StartTest(ctx context.Context, params StartTestParams) (TestHandle, error) {
 	// Validate params
+	if params.LocalPolicy != nil && params.LocalPolicy.RiskScoreThreshold != nil {
+		threshold := *params.LocalPolicy.RiskScoreThreshold
+		if threshold > 1000 { // uint16 cannot be < 0
```
PeterSchafer, line 237: Question: Is this a check we should do in the client application? This might make future changes more difficult. I would expect such checks in the backend with appropriate errors reported back.

PeterSchafer (approved):
Pre-Approving as the comments are minor concerns.

pkg/apiclients/testapi/testapi_test.go
bsalomon-snyk: Valid concern.  I'll bring this up with the devs defining the backend spec and continue the discussion.

PR #371, opened 2025-06-18
chore: add tip when no repo url is available

PeterSchafer:


ShawkyZ (approved):

fsargent (approved):
Well this is infinitely better and simpler than my PR.

PR #380, opened 2025-07-15
fix(logging): limit size of request bodies in trace level logging

PeterSchafer:


CatalinSnyk (approved):

pkg/networking/logging.go
```diff
@@ -108,7 +116,7 @@ func LogRequest(r *http.Request, logger *zerolog.Logger) {
 		return
 	}
 
-	logBody(logger, defaultNetworkLogLevel, logPrefixRequest, getRequestBody(r), r.Header)
+	logBody(logger, defaultNetworkLogLevel, logPrefixRequest, getRequestBody(r), r.Header, 80)
```
CatalinSnyk, line 119: Just a super small nit: I would say we might wanna put the 80 magic number in a constant.

PR #381, opened 2025-07-15
chore: remove script pointing to deprecated Dragonfly repo to avoid any confusion

bsalomon-snyk:
Summary: delete unused script

Note to reviewers:
This was re-added in June 2025 to help transition away from the deprecated repo, and is no longer needed.


PeterSchafer:
thank you @bsalomon-snyk for trying to cleanup. I think it is a bit too early to remove this script. 
We had to re-add it because we had to extend the v0 version of the local findings model. As long as this data structure is still there, I think we should keep the script. I will create a ticket to cleanup the old data structure and include the script there as well.

PR #395, opened 2025-08-01
feat: add PAT auto-region configuration

j-luong:


pkg/local_workflows/auth_workflow_test.go
```diff
@@ -120,13 +115,10 @@ func Test_pat(t *testing.T) {
 	authenticator := mocks.NewMockAuthenticator(mockCtl)
 
 	t.Run("happy", func(t *testing.T) {
+		pat := `{"h":"api.snyk.io"}`
```
j-luong, line 118: todo: use config instead of mocking it

pkg/app/app.go
```diff
@@ -89,6 +90,15 @@ func defaultFuncApiUrl(_ configuration.Configuration, logger *zerolog.Logger) co
 
 		if len(urlFromOauthToken) > 0 && len(urlFromOauthToken[0]) > 0 {
 			urlString = urlFromOauthToken[0]
+		} else if auth.IsAuthTypePAT(config.GetString(configuration.AUTHENTICATION_TOKEN)) {
```
PeterSchafer, line 93: Suggestion: maybe avoid the duplication of `config.GetString(configuration.AUTHENTICATION_TOKEN)`

pkg/app/app.go
```diff
@@ -89,6 +90,15 @@ func defaultFuncApiUrl(_ configuration.Configuration, logger *zerolog.Logger) co
...
+			claims, claimsErr := auth.ExtractClaimsFromPAT(config.GetString(configuration.AUTHENTICATION_TOKEN))
```
PeterSchafer, line 94: Suggestion: to reduce the complexity of this function here, maybe already already return the url from the pat token, instead of returning the claim etc. basically encapsulate all this pat specific logic in a function. maybe even just reframe `ExtractClaimsFromPAT`

pkg/auth/tokenauthenticator.go
```diff
@@ -66,3 +79,26 @@ func IsAuthTypePAT(token string) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode payload: %w", err)
+	}
+
+	var c Claims
+	if err = json.Unmarshal(payload, &c); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal payload: %w", err)
+	}
+
+	return &Claims{
```
PeterSchafer, line 101: Question: maybe an obsolete question due to [this proposal](https://github.com/snyk/go-application-framework/pull/395/files#r2254143585) but why can't we just return `c`?

pkg/auth/tokenauthenticator_test.go
```diff
@@ -17,3 +19,73 @@ func TestIsAuthTypePAT(t *testing.T) {
...
+
+		claims, err := extractClaimsFromPAT(pat)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "failed to unmarshal payload")
+		assert.Nil(t, claims)
+	})
+}
+
+func TestGetApiUrlFromPAT(t *testing.T) {
+	t.Run("Valid PAT", func(t *testing.T) {
```
j-luong, line 77: todo: unhappy test needed

pkg/auth/tokenauthenticator_test.go
```diff
@@ -17,3 +19,73 @@ func TestIsAuthTypePAT(t *testing.T) {
...
+		pat := createMockPAT(t, `{"h":"api.snyk.io"}`)
+		apiUrl, err := GetApiUrlFromPAT(pat)
+		assert.NoError(t, err)
+		assert.Equal(t, "api.snyk.io", apiUrl)
+	})
+}
+
+func createMockPAT(t *testing.T, payload string) string {
```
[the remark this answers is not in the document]
j-luong: Functions in _test.go files are only compiled and accessible during tests, not in production code or other packages. To use it outside, it must be moved to a non-test file.

pkg/local_workflows/auth_workflow_test.go
```diff
@@ -103,12 +103,6 @@ func Test_auth_token(t *testing.T) {
 }
 
 func Test_pat(t *testing.T) {
-	const (
```
PeterSchafer, line 106: Question: why was this and the other lines here removed? might be for a good reason, it is just not clear to me.

PR #396, opened 2025-08-05
chore: iaw uses org setting insted of feature flag

CatalinSnyk:
Previously, the IAW create workflow was gated behind a feature flag check. For EA, customers should be able to opt in/out of using the ignore approval, so a new org setting was added for them to toggle on/off. 

This PR just swaps out the feature flag check with an fetching the org settings and checking whether IAW is enabled for the current org before proceeding with the ignore create workflow.

Ref: [CLI-1006](https://snyksec.atlassian.net/browse/CLI-1006)

[CLI-1006]: https://snyksec.atlassian.net/browse/CLI-1006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -249,7 +249,7 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
 }
 
 func (a *snykApiClient) GetOrgSettings(orgId string) (*contract.OrgSettingsResponse, error) {
-	endpoint := a.url + fmt.Sprintf("/v1/org/%s/settings", url.QueryEscape(orgId))
+	endpoint := fmt.Sprintf("%s/v1/org/%s/settings", a.url, url.QueryEscape(orgId))
 	res, err := a.client.Get(endpoint)
 	if err != nil {
 		return nil, fmt.Errorf("unable to retrieve org settings: %w", err)
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -25,6 +25,7 @@ type ApiClient interface {
 	GetUserMe() (string, error)
 	GetSelf() (contract.SelfResponse, error)
 	GetSastSettings(orgId string) (*sast_contract.SastResponse, error)
+	GetOrgSettings(orgId string) (*contract.OrgSettingsResponse, error)
 }
 
 var _ ApiClient = (*snykApiClient)(nil)
@@ -242,6 +243,28 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
 	var response sast_contract.SastResponse
 	if err = json.Unmarshal(body, &response); err != nil {
 		return nil, fmt.Errorf("unable to retrieve settings (status: %d): %w", res.StatusCode, err)
+	}
+
+	return &response, err
+}
+
+func (a *snykApiClient) GetOrgSettings(orgId string) (*contract.OrgSettingsResponse, error) {
+	endpoint := a.url + fmt.Sprintf("/v1/org/%s/settings", url.QueryEscape(orgId))
+	res, err := a.client.Get(endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve org settings: %w", err)
+	}
+	//goland:noinspection GoUnhandledErrorResult
+	defer res.Body.Close()
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve org settings: %w", err)
+	}
+
+	var response contract.OrgSettingsResponse
+	if err = json.Unmarshal(body, &response); err != nil {
+		return nil, fmt.Errorf("unable to retrieve org settings (status: %d): %w", res.StatusCode, err)
 	}
 
 	return &response, err
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -249,22 +249,15 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
 }
 
 func (a *snykApiClient) GetOrgSettings(orgId string) (*contract.OrgSettingsResponse, error) {
-	endpoint := fmt.Sprintf("%s/v1/org/%s/settings", a.url, url.QueryEscape(orgId))
-	res, err := a.client.Get(endpoint)
-	if err != nil {
-		return nil, fmt.Errorf("unable to retrieve org settings: %w", err)
-	}
-	//goland:noinspection GoUnhandledErrorResult
-	defer res.Body.Close()
-
-	body, err := io.ReadAll(res.Body)
+	endpoint := fmt.Sprintf("/v1/org/%s/settings", url.QueryEscape(orgId))
+	body, err := clientGet(a, endpoint, nil)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -60,6 +60,8 @@ const (
 
 	interactiveEnsureVersionControlMessage    = "👉🏼 Ensure the code containing the issue is committed and pushed to remote origin, so approvers can review it."
 	interactiveIgnoreRequestSubmissionMessage = "✅ Your ignore request has been submitted."
+
+	configIgnoreApprovalEnabled = "internal_iaw_enabled"
```
ShawkyZ, line 64: shouldn't this be exported to call it from the IDE as well ?

internal/api/api.go
```diff
@@ -247,6 +248,28 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
 	return &response, err
 }
 
+func (a *snykApiClient) GetOrgSettings(orgId string) (*contract.OrgSettingsResponse, error) {
+	endpoint := a.url + fmt.Sprintf("/v1/org/%s/settings", url.QueryEscape(orgId))
```
bastiandoetsch, line 252: a.url could actually be pulled into the fmt.Sprintf, too :)

internal/api/contract/OrgSettingsResponse.go
```diff
@@ -0,0 +1,11 @@
+package contract
+
+type OrgIgnoreSettings struct {
+	ReasonRequired          bool `json:"reasonRequired,omitempty"`
+	AutoApproveIgnores      bool `json:"autoApproveIgnores,omitempty"`
+	ApprovalWorkflowEnabled bool `json:"approvalWorkflowEnabled,omitempty"`
+}
+
+type OrgSettingsResponse struct {
+	Ignores OrgIgnoreSettings `json:"ignores"`
```
bastiandoetsch, line 10: Why only Ignores? Are there not more fields? If we add them now, we don't have to touch the implementation again, when we need a different org setting :).

pkg/local_workflows/ignore_workflow/config.go
```diff
@@ -43,6 +44,28 @@ func addCreateIgnoreDefaultConfigurationValues(invocationCtx workflow.Invocation
 	})
 }
 
+func getOrgIgnoreApprovalEnabled(engine workflow.Engine) configuration.DefaultValueFunction {
+	return func(existingValue interface{}) (interface{}, error) {
+		if existingValue != nil {
```
bastiandoetsch, line 49: this should probably be backed by a cache - after xy time (e.g. 30min), it should not use the previous value. This is important for long-running CLIs, e.g. for Language Server.

PeterSchafer: The config uses a cache and how long things are cached is based on the a setting for the configuration. Caching is transparent for the default function and it just needs to handle the business logic itself.

internal/api/contract/OrgSettingsResponse.go
CatalinSnyk: Added the only other [field](https://github.com/snyk/registry/blob/0a8ef04ea1bc57bc0a23781d9ed2de501eed73cf/src/web/routes/api/v1/external/org/settings/types.ts#L15-L18) there was.

A question for the audience from my side now: 
- Should I be marking [the fields as pointers](https://github.com/snyk/registry/blob/0a8ef04ea1bc57bc0a23781d9ed2de501eed73cf/src/web/routes/api/v1/external/org/settings/types.ts#L15-L27) since they are marked as optional, even if all of them are boolean values? I was thinking that if a field is not specified then it should be handled as false, but I'm wondering if that's a good assumption.

internal/api/api.go
```diff
@@ -247,6 +248,28 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
...
+		return nil, fmt.Errorf("unable to retrieve org settings: %w", err)
+	}
+
+	var response contract.OrgSettingsResponse
+	if err = json.Unmarshal(body, &response); err != nil {
+		return nil, fmt.Errorf("unable to retrieve org settings (status: %d): %w", res.StatusCode, err)
+	}
+
+	return &response, err
+}
```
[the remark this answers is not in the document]
j-luong: suggestion: I think it's suggesting to use [clientGet()](https://github.com/snyk/go-application-framework/pull/396/files#diff-5b9b550fff634f1fd5596893d63ceeb6afa6a355427db919a644a21f097b7834R292), which wraps `a.client.Get()` with status code checks.

internal/api/api.go
```diff
@@ -247,6 +248,28 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
 	return &response, err
 }
 
+func (a *snykApiClient) GetOrgSettings(orgId string) (*contract.OrgSettingsResponse, error) {
+	endpoint := fmt.Sprintf("%s/v1/org/%s/settings", a.url, url.QueryEscape(orgId))
+	res, err := a.client.Get(endpoint)
```
j-luong, line 253: suggestion: use [clientGet()](https://github.com/snyk/go-application-framework/pull/396/files#diff-5b9b550fff634f1fd5596893d63ceeb6afa6a355427db919a644a21f097b7834R292) instead

internal/api/contract/OrgSettingsResponse.go
j-luong: > Should I be marking [the fields as pointers](https://github.com/snyk/registry/blob/0a8ef04ea1bc57bc0a23781d9ed2de501eed73cf/src/web/routes/api/v1/external/org/settings/types.ts#L15-L27) since they are marked as optional, even if all of them are boolean values? I was thinking that if a field is not specified then it should be handled as false, but I'm wondering if that's a good assumption.

I think so, pointers along with the `omitempty` tag suffix seem to be the accepted pattern in the codebase.

internal/api/api.go
CatalinSnyk: Added it, only think I don't especially like is the fact that clientGet is for versioned API's and the one we are using is `v1`, but I'm guessing it should be fine?

internal/api/api.go
```diff
@@ -247,6 +248,29 @@ func (a *snykApiClient) GetSastSettings(orgId string) (*sast_contract.SastRespon
...
+		return nil, fmt.Errorf("unable to retrieve org settings: %w", err)
+	}
+
+	var response contract.OrgSettingsResponse
+	if err = json.Unmarshal(body, &response); err != nil {
+		return nil, fmt.Errorf("unable to retrieve org settings (status: %d): %w", res.StatusCode, err)
+	}
+
+	return &response, err
+}
```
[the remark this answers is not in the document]
CatalinSnyk: Should be handled at the networking level by our middleware https://github.com/snyk/go-application-framework/blob/67306d427077e3faf06d2e6b2d79585fc185b279/pkg/networking/middleware/response.go#L88

PR #397, opened 2025-08-06
fix: stop logging large content [IDE-1351]

rrama:
We have an issue where if a large binary file download was interrupted the amount that was downloaded will be logged, which due to scrubbing would use a lot of RAM.

This PR contains <s>3</s> 2 techniques that will prevent us from logging / scrubbing overly large content. They would all work on their own, so we only actually need to implement 1 of these. I will let people have their say on which one or both.


pkg/networking/logging.go
```diff
@@ -14,7 +15,21 @@ import (
 const defaultNetworkLogLevel = zerolog.DebugLevel
 const extendedNetworkLogLevel = zerolog.TraceLevel
 const maxNumberOfRequestBodyCharacters = 60
-const maxNumberOfResponseBodyCharacters = -1 // log complete response body
+const maxNumberOfResponseBodyCharacters = 10 * 1024 // 10KiB, to ensure we don't log an entire CLI download response that is interrupted.
```
j-luong, line 18: suggestion: I think limiting the response body characters is most reasonable here

PeterSchafer, line 18: Praise: I think setting `maxNumberOfResponseBodyCharacters`  is the best minimal solution.

rrama:
Basti has said:
> I'd use a hybrid - no logging for binary mime types. I don't want to see 10k of binary symbols in the log.

So we are 2 vs 1 atm, with minimal fix currently winning. I'm going to claim I couldn't possibly vote on picking between my own two <s>children</s> code snippets.

ShawkyZ (approved):

PR #401, opened 2025-08-12
feat: [DGP-730] jitter TestAPI polling from 0.5 to 1.5 of the pollInterval

bsalomon-snyk:
### What this does

Jitters TestAPI's test job polling from 0.5 to 1.5 of the pollInterval to avoid falling into a sync pattern with other clients or threads


dotkas (approved):
LGTM - but you don't want a test for this? Don't know how hard it is to do. But could maybe be interesting to ensure upper and lower bounds are not breached, unless this is already tested.

pkg/apiclients/testapi/testapi.go
```diff
@@ -691,3 +694,13 @@ func (r *testResult) handleFindingsError(err error, partialFindings []FindingDat
 func ptr[T any](v T) *T {
 	return &v
 }
+
+// Jitter returns a random duration between 0.5 and 1.5 of the given duration.
+func Jitter(d time.Duration) time.Duration {
```
PeterSchafer, line 699: Suggestion: Let's add a simple Unit test for this please.

PeterSchafer (approved):

PR #405, opened 2025-08-15
fix: get default org when org in config is empty [IDE-1375]

acke:
Sometimes the Organization is overwritten by "" and saved in the configs organisation cache. This disables the default function from running and retrieving the actual default organisation.

This PR clears the cache if the Organization config is empty, thus allowing for the default func to be triggered.


andrewrobinsonhodges-snyk (approved):

PeterSchafer:
@acke if I understand your changes correctly, you are clearing the cache when the cached value is empty. did you consider doing this more centrally, for example not caching empty values at all? A more central solution would benefit more parts of the application.

PR #409, opened 2025-08-21
fix: respect the reason required org setting for ignore creation

CatalinSnyk:
Adds support for the `ReasonRequired` org setting in regards to ignores. The changes add a new config value for it and adapt the validator function in regards to the setting.

### DIFF
--- a/pkg/local_workflows/ignore_workflow/config.go
+++ b/pkg/local_workflows/ignore_workflow/config.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/snyk/error-catalog-golang-public/cli"
 	"github.com/snyk/go-application-framework/internal/api"
+	"github.com/snyk/go-application-framework/internal/api/contract"
 	policyApi "github.com/snyk/go-application-framework/internal/api/policy/2024-10-15"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/local_workflows/local_models"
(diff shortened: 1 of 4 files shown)


pkg/local_workflows/ignore_workflow/config.go
```diff
@@ -40,26 +41,55 @@ func addCreateIgnoreDefaultConfigurationValues(invocationCtx workflow.Invocation
 
 	config.AddDefaultValue(ReasonKey, func(_ configuration.Configuration, existingValue interface{}) (interface{}, error) {
 		isSet := config.IsSet(ReasonKey)
-		return defaultFuncWithValidator(existingValue, isSet, isValidReason)
+		reasonRequired := config.GetBool(ConfigIgnoreReasonRequired)
+		return defaultFuncWithValidator(existingValue, isSet, getReasonValidator(reasonRequired))
 	})
 }
 
+func getOrgIgnoreSettings(engine workflow.Engine) (*contract.OrgSettingsResponse, error) {
```
j-luong, line 49: suggestion: pass in the necessary parameters only, instead of the entire engine

pkg/local_workflows/ignore_workflow/config.go
```diff
@@ -70,6 +100,26 @@ func getOrgIgnoreApprovalEnabled(engine workflow.Engine) configuration.DefaultVa
 	}
 }
 
+func getOrgIgnoreReasonRequired(engine workflow.Engine) configuration.DefaultValueFunction {
```
j-luong, line 103: suggest: `getOrgIgnoreReasonRequired`, `getOrgIgnoreSettingsConfig`, `getOrgIgnoreApprovalEnabled` are all making use of `getOrgIgnoreSettings`. Is `getOrgIgnoreSettings` something that can be added to config via `config.AddDefaultValue`? That way we can take advangate of caching to save mulutiple API calls.

PR #418, opened 2025-08-28
feat(config): Introduce concept of configuration dependencies

PeterSchafer:
This PR introduces a new optional concept of configuration dependencies.

This means that for example value A depends on value B, if value B gets changed, the cache of value A gets invalidated. This is also true for deeper nested dependencies. This is an optional pattern to express these types of dependencies and ensure that caching can still be efficiently used. 

This PR also removes a workaround that was recently introduced to force clear a configuration cache in the OAuth2Authenticator. This workaround has more side effects than expected and modelling the dependencies based on the new pattern is actually the better way forward.

### DIFF
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -32,6 +32,28 @@ import (
 	"golang.org/x/oauth2/jws"
 )
 
+func createOAuthTokenWithAudience(t *testing.T, audience string) string {
+	t.Helper()
+	header := &jws.Header{}
+	claims := &jws.ClaimSet{
+		Aud: audience,
+	}
+	pk, err := rsa.GenerateKey(rand.Reader, 2048)
+	assert.NoError(t, err)
+
+	accessToken, err := jws.Encode(header, claims, pk)
+	assert.NoError(t, err)
+
+	token := oauth2.Token{
+		AccessToken: accessToken,
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/app/app.go
```diff
@@ -236,14 +255,20 @@ func initConfiguration(engine workflow.Engine, config configuration.Configuratio
...
-	config.AddDefaultValue(configuration.API_URL, defaultFuncApiUrl(config, logger))
 	config.AddDefaultValue(configuration.TEMP_DIR_PATH, defaultTempDirectory(engine, config, logger))
 
+	config.AddDefaultValue(configuration.API_URL, defaultFuncApiUrl(config, logger))
+
+	err = config.AddKeyDependency(configuration.WEB_APP_URL, configuration.API_URL)
+	if err != nil {
+		logger.Print("Failed to add dependency for WEB_APP_URL:", err)
+	}
+
```
dotkas, line 266: nitpick: Maybe we should move this below line `267`? 

That would keep it consistent with the "first add dependency, then dependencies" flow I've seen above... I think?

And I wouldn't normally nitpick over these, but I feel the `initConfiguration` is already super verbose - so things could easily be overlooked when we change things a year from now and remember nothing, to avoid leaving "dangling" logic.

dotkas (approved):

PR #419, opened 2025-08-29
fix: Fix improper regex that can lead to SSRF

j-luong:
This PR addresses improper regex checks that can lead to an SSRF vulnerability

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -102,7 +102,7 @@ func defaultFuncApiUrl(_ configuration.Configuration, logger *zerolog.Logger) co
 
 		// Same logic for PAT - if a PAT is provided, and it has a URL in the claims, use that instead
 		if auth.IsAuthTypePAT(authToken) {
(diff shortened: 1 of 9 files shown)


dotkas (approved):

pkg/app/app.go
```diff
@@ -102,7 +102,7 @@ func defaultFuncApiUrl(_ configuration.Configuration, logger *zerolog.Logger) co
 
 		// Same logic for PAT - if a PAT is provided, and it has a URL in the claims, use that instead
 		if auth.IsAuthTypePAT(authToken) {
-			apiUrl, claimsErr := auth.GetApiUrlFromPAT(authToken)
+			apiUrl, claimsErr := auth.GetApiUrlFromPAT(authToken, config.GetString(auth.CONFIG_KEY_ALLOWED_HOST_REGEXP))
```
PeterSchafer, line 105: Question: is this something we should do for the apiUrl from the Oauth token as well?

dotkas (changes requested):
I looked over it again, can we sync up on it?

PR #423, opened 2025-09-08
chore: use `go tool` [IDE-1377]

rrama:
### What this does

With the new tool approach in `go.mod`, `tools.go` was no longer needed and has been removed.
Use oapi-codegen v2.4.1, to be aligned with the version in go.mod.
Regenerated all files after the update.
The tool will automatically download when `go tool` is used if you do not already have the version downloaded.

### Why the change?

No more will you have to remember to `make clean` and `make install-tools`, etc. when the version of the tool changes. No more will people run with the wrong version of the tool, causing inconsistencies with mocks in the codebase because someone forgot to do this.
Now you can `go generate` a single file without having to set up your `PATH` correctly first. This means you can click the green play button in your IDE to run the generate command in the file.

### Notes for the reviewer

For more information on `go tool` see https://tip.golang.org/doc/go1.24#tools


Makefile
```diff
@@ -26,6 +26,7 @@ build:
 clean:
 	@echo "Cleaning up..."
 	@GOOS=$(GOOS) GOARCH=$(GOARCH) go clean -testcache
+	@rm -rf $(GO_BIN)
```
bastiandoetsch, line 29: i would leave that for migration. else we'll keep an orphaned directory.

bastiandoetsch (approved):

Makefile
rrama: What's the issue with removing the tools (`golangci-lint` and the now unused Go ones) on a `make clean`? We do it in CCG.
What directory would be orphaned?

go.mod
```diff
@@ -121,4 +121,9 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
+tool (
+	github.com/golang/mock/mockgen
```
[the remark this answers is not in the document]
rrama: The version was already in the go.mod, as we use it as a direct dependency. (line 8: `github.com/golang/mock v1.6.0`)

If we only used it as a tool and not a direct code dependency, then it adds as an indirect dependency, like what has happened to `github.com/oapi-codegen/oapi-codegen/v2 v2.4.1 // indirect`.

This does mean that you can't use a different version of the tool than you use in the code, but to me that is a feature and not a bug, as I don't think there was any reason for us to be using oapi-codegen v2.4.1 as a dependency in the code, but using v2.3.0 for the `oapi-codegen` binary we were using.

PR #424, opened 2025-09-08
chore: update golangci-lint to v2.3.0 [IDE-1377]

rrama:
Migrate stylecheck, gosimple, and varcheck into staticcheck (they are now combined).
Add unused linter with `exported-fields-are-used: false` to replace varcheck's `exported-fields: true`.
Enable staticcheck explicitly, it was previously enabled, but was not listed.
Remove deprecated `EXC####` exclusion rules.
Disable package comment rules (ST1000).
Disable naming convention rules (ST1003).
Move the timeout to command-line arguments (v2 discourages timeouts in the config).
Re-enable stuttering checks.
Fix or ignore any new lint errors.
Improve updating of golangci-lint, now will be automatically updated if needed before use.

### Notes for the reviewer

I would recommend comparing `.golangci.yaml` with whitespace ignored.
Migration guide: https://golangci-lint.run/docs/product/migration-guide/


acke (approved):

rrama:
Stale, closing. The only still relevant part is the Makefile changes, where I made it write the version of the linter to a marker file, so we can ensure it's always the correct version automatically when we bump it.

PR #432, opened 2025-09-25
chore: use static terms rather the regexp for config values

PeterSchafer:
Configuration values should be added as static terms rather then as regular expression, to avoid panics when compiling the strings, as they might be non compilable.


bastiandoetsch (approved):
Sounds good.

PR #435, opened 2025-09-29
feat: support new test service commit ID

j-luong:
This PR updates the code-client-go with updates to support test service commit ID metadata


PeterSchafer (approved):
Pre-approving, assuming that `WithCommitId()` will be added

PR #438, opened 2025-10-03
feat: [IGNR-1709] Add source in payload

harshdev93:
Description:

This PR adds source in the payload for creating an ignore request. The source can be IDE/CLI depending on where the request is from. The source is determined from RuntimeInfo.

Jira: https://snyksec.atlassian.net/browse/IGNR-1709

### DIFF
--- a/pkg/local_workflows/ignore_workflow/ignore_workflow.go
+++ b/pkg/local_workflows/ignore_workflow/ignore_workflow.go
@@ -232,7 +232,7 @@ func ignoreCreateWorkflowEntryPoint(invocationCtx workflow.InvocationContext, _ 
 		return nil, err
 	}
 
-	payload := createPayload(repoUrl, branchName, expire, ignoreType, reason, findingsId)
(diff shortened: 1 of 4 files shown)


pkg/local_workflows/ignore_workflow/ignore_workflow.go
```diff
@@ -300,11 +300,25 @@ func createIgnoreWorkflowData(id workflow.Identifier, config configuration.Confi
...
 	meta := policyApi.Meta{}
 	meta["repo_url"] = repoUrl
 	meta["branch_name"] = branchName
 
+	// Determine source based on RuntimeInfo
+	var source *policyApi.PolicyAttributesSource
+	runtimeInfo := invocationCtx.GetRuntimeInfo()
+	if runtimeInfo != nil {
+		runtimeName := runtimeInfo.GetName()
+		if runtimeName == "snyk-ls" {
```
rrama, line 313: Can we not just send up the `runtimeName` instead of having this hard-coded string switching here? As this will break if another GAF consumer is added. For instance if we allow the MCP server to send up ignore approval requests.

PR #559, opened 2026-02-26
test: add failing test for multiple JSON output

snyk-tim:
## Summary

This PR adds a failing test that demonstrates an issue where only the first `workflow.Data` item is output when multiple items with `application/json` content type are passed to the output workflow with `--json` flag.

**Question for the team:** Is this intended behavior, or should multiple JSON items be output?

## Problem

When a workflow returns multiple `workflow.Data` items (e.g., scanning multiple lock files), only the first JSON result is printed to `stdout`. All subsequent results are silently dropped.

## Expectation

When encountering multiple `workflow.Data` they should all be output with the correct renderer. In this case with `application/json` I would expect the output to be JSONL/NDJSON. One line per JSON separated by a `\n` character.

### DIFF
--- a/pkg/local_workflows/output_workflow_test.go
+++ b/pkg/local_workflows/output_workflow_test.go
@@ -464,6 +464,38 @@ func Test_Output_outputWorkflowEntryPoint(t *testing.T) {
 		assert.Equal(t, expectedOutput, setup.writer.String())
 	})
 
+	// This test demonstrates a bug: when multiple application/json items are passed with --json flag,
+	// only the first item is output. The PopWritersByMimetype function removes the writer after the
+	// first use, causing subsequent items to find no writer available.
+	t.Run("should output all application/json items when json flag is set", func(t *testing.T) {
+		setup := setupTest(t)
+		setup.config.Set(output_workflow.OUTPUT_CONFIG_KEY_JSON, true)
+		defer setup.config.Set(output_workflow.OUTPUT_CONFIG_KEY_JSON, nil)
+
+		workflowIdentifier := workflow.NewTypeIdentifier(WORKFLOWID_OUTPUT_WORKFLOW, "output")
+
+		jsonPayload1 := `{"project":"first-project","dependencies":["pkg-a","pkg-b"]}`
+		jsonPayload2 := `{"project":"second-project","dependencies":["pkg-c","pkg-d"]}`
+
+		data1 := workflow.NewData(workflowIdentifier, "application/json", []byte(jsonPayload1))
+		data1.SetContentLocation("first-project/lock.file")
+
+		data2 := workflow.NewData(workflowIdentifier, "application/json", []byte(jsonPayload2))
+		data2.SetContentLocation("second-project/lock.file")
+
+		output, err := outputWorkflowEntryPoint(setup.invocationContextMock, []workflow.Data{data1, data2}, setup.outputDestination)
+
+		assert.Nil(t, err)
+		assert.Equal(t, []workflow.Data{}, output)
+
+		// BUG: Currently only the first JSON is output because PopWritersByMimetype
+		// removes the writer after the first item is processed.
+		// Expected: both JSON payloads should be in the output
+		outputStr := setup.writer.String()
+		assert.Contains(t, outputStr, "first-project", "First JSON payload should be in output")
+		assert.Contains(t, outputStr, "second-project", "Second JSON payload should be in output (BUG: this fails)")
+	})
+
 	t.Run("should print valid sarif json output", func(t *testing.T) {
 		setup := setupTest(t)
 		expectedSarif := "testdata/sarif-snyk-goof-ignores.json"


pkg/local_workflows/output_workflow_test.go
```diff
@@ -464,6 +464,38 @@ func Test_Output_outputWorkflowEntryPoint(t *testing.T) {
...
+
+		assert.Nil(t, err)
+		assert.Equal(t, []workflow.Data{}, output)
+
+		// BUG: Currently only the first JSON is output because PopWritersByMimetype
+		// removes the writer after the first item is processed.
+		// Expected: both JSON payloads should be in the output
+		outputStr := setup.writer.String()
+		assert.Contains(t, outputStr, "first-project", "First JSON payload should be in output")
+		assert.Contains(t, outputStr, "second-project", "Second JSON payload should be in output (BUG: this fails)")
```
bastiandoetsch, line 496: [Critical] This is an ungated failing test — `go test ./pkg/local_workflows/` fails here today (`"...first-project..." does not contain "second-project"`), empirically confirmed. There's no `t.Skip`, build tag, or other gate, and CI runs this in required gates, so merging it alone turns `main` red for every subsequent pipeline.

The bug it documents is real and well-diagnosed: `HandleContentTypeOther` loops per item and `useWriterWithOther` → `PopWritersByMimetype` `delete`s the writer from the map after the first use (writer.go:58-62), so the second `application/json` item finds no writer and is dropped. Nice reproduction.

Resolution (one required before merge): (a) land the production fix in this same PR so the test goes green on commit (preferred — a failing test should accompany its fix), or (b) `t.Skip("<TICKET>: multiple application/json items dropped by PopWritersByMimetype — un-skip when fixed")` to keep CI green while preserving the repro.

— AI review

pkg/local_workflows/output_workflow_test.go
```diff
@@ -464,6 +464,38 @@ func Test_Output_outputWorkflowEntryPoint(t *testing.T) {
...
+		data1 := workflow.NewData(workflowIdentifier, "application/json", []byte(jsonPayload1))
+		data1.SetContentLocation("first-project/lock.file")
+
+		data2 := workflow.NewData(workflowIdentifier, "application/json", []byte(jsonPayload2))
+		data2.SetContentLocation("second-project/lock.file")
+
+		output, err := outputWorkflowEntryPoint(setup.invocationContextMock, []workflow.Data{data1, data2}, setup.outputDestination)
+
+		assert.Nil(t, err)
+		assert.Equal(t, []workflow.Data{}, output)
```
bastiandoetsch, line 489: [Should Fix] This assertion is a tautology that gives false confidence. `outputWorkflowEntryPoint` always returns the empty slice declared at output_workflow.go:39 and discards the `input`/leftover returned by the `Handle*` calls, so `output` is `[]workflow.Data{}` regardless of how many items were dropped — it passes in both the buggy and fixed states. It reads as if it verifies 'no data left unprocessed', but the dropped second item is in fact neither written nor returned; this assertion hides that. Drop it, or replace with one that pins a real contract (assert the second item is written, or surfaced as unconsumed — currently it's neither).

— AI review

bastiandoetsch:
The bug diagnosis here is genuinely good — a precise, deterministic reproduction of a real `PopWritersByMimetype` defect. The issues are about how it's committed, not the investigation. Beyond the two inline items:

[Suggestion] The `Contains("second-project")` check demonstrates the symptom but doesn't pin the *correct* multi-JSON output contract — a fix that concatenates two JSON objects into malformed output would still make it pass. Decide and assert the intended format (JSON array vs NDJSON vs separated objects) so the test drives a correct fix, not just any fix.
[Suggestion] The subtest name ('should output all application/json items...') describes the post-fix green state while the body asserts the failing state with a '(BUG: this fails)' message — self-contradicting once fixed. After the fix, the name is right and the BUG parenthetical/comments (lines 467-469, 491-493, 496) should become a neutral regression-guard description.
[Suggestion] Reference the tracking ticket in the skip reason/comment so there's a breadcrumb to the fix.

— AI review

pkg/local_workflows/output_workflow_test.go
```diff
@@ -464,6 +464,38 @@ func Test_Output_outputWorkflowEntryPoint(t *testing.T) {
 		assert.Equal(t, expectedOutput, setup.writer.String())
 	})
 
+	// This test demonstrates a bug: when multiple application/json items are passed with --json flag,
+	// only the first item is output. The PopWritersByMimetype function removes the writer after the
+	// first use, causing subsequent items to find no writer available.
+	t.Run("should output all application/json items when json flag is set", func(t *testing.T) {
```
basti-snyk, line 470: **Should-Fix:** This adds a knowingly-failing test with no accompanying fix. Verified by running it: `assert.Contains(outputStr, "second-project")` fails — writer output is `{"project":"first-project",...}` only. That fails the parent `Test_Output_outputWorkflowEntryPoint` and makes `go test ./pkg/local_workflows/...` exit non-zero. If this branch merges to `main` on its own it red-lines CI for every consumer of the package until the fix lands.

The bug it demonstrates is real and the root cause is correctly identified: `writerMapImpl.PopWritersByMimetype` (`output_workflow/writer.go`) does `delete(w.writers, mimeType)`, so the single JSON writer is consumed by the first item and the second finds none.

Before merge, either land the production fix in the same PR/stack (test turns green), or quarantine with `t.Skip("<ticket>: multiple application/json items drop all but the first")` referencing the bug, removed by the fix commit. If this PR is deliberately the RED step of a stack whose next commit carries the fix, this is acceptable — flagging so the decision is explicit.

— AI review

pkg/local_workflows/output_workflow_test.go
basti-snyk, line 489: **Suggestion (non-blocking):** `assert.Equal(t, []workflow.Data{}, output)` is vacuous here — `outputWorkflowEntryPoint` initializes `output := []workflow.Data{}` and returns it unchanged (handler results are assigned back into `input`, never `output`), so this passes for the same trivial reason in every existing test and cannot detect the multiple-JSON bug. The meaningful signal is the writer content asserted below.

— AI review

PR #564, opened 2026-03-05
fix: allow the context on env load to be canceled [IDE-1791]

rrama:
Plus other nice to have options can be passed.
Needed for https://github.com/snyk/snyk-ls/pull/1173

### DIFF
--- a/pkg/envvars/environment.go
+++ b/pkg/envvars/environment.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


bastiandoetsch (approved):

pkg/envvars/environment.go
```diff
@@ -20,14 +21,93 @@ const (
...
+	options := loadConfiguredEnvironmentOptions{
+		ctx:    context.Background(),
+		logger: utils.Ptr(zerolog.Nop()),
+	}
+	for _, opt := range opts {
+		opt(&options)
+	}
+	logger := options.logger.With().Str("method", "LoadConfiguredEnvironment").Logger()
+
+	mu.Lock()
```
basti-snyk, line 81: **Should fix — the mutex acquire is not context-aware, so cancellation isn't honoured while queued.**

Following up on the earlier thread where the timeout was deliberately left inside `getEnvFromShell` rather than wrapping the whole load: even with that decision, `mu.Lock()` here is a plain, non-cancelable acquire. A caller that passes a context specifically to bound/cancel the load will still block on this global lock while a *concurrent* call runs — and each call can hold the lock through two sequential 5s shell execs (~10s). During that wait the caller's `ctx` cancellation has no effect. So the "cancelable env load" guarantee holds for the *running* call (via `exec.CommandContext`) and at the phase checkpoints, but not for a *queued* caller.

If prompt return for a queued+canceled caller is in scope, acquire the lock in a ctx-aware way (channel semaphore / `TryLock` polled against `ctx.Done()`). If it's out of scope, worth a comment so the limitation is explicit.

— AI review

pkg/envvars/environment.go
```diff
@@ -20,14 +21,93 @@ const (
...
+	ctx               context.Context
+	customConfigFiles []string
+	logger            *zerolog.Logger
+}
+
+type loadConfiguredEnvironmentOption func(opts *loadConfiguredEnvironmentOptions)
+
+func WithContext(ctx context.Context) loadConfiguredEnvironmentOption {
+	return func(opts *loadConfiguredEnvironmentOptions) {
+		opts.ctx = ctx
```
basti-snyk, line 37: **Should fix — `WithContext(nil)` will cause a nil-pointer panic.**

Unlike `WithLogger` (which explicitly guards `if logger == nil { return }`), `WithContext` sets `opts.ctx = ctx` unconditionally. `WithContext(nil)` leaves `options.ctx == nil`, and the first `options.ctx.Err()` (in the deferred closure and the pre-load check) then panics with a nil dereference.

Mirror the `WithLogger` guard:

```go
func WithContext(ctx context.Context) loadConfiguredEnvironmentOption {
	return func(opts *loadConfiguredEnvironmentOptions) {
		if ctx == nil {
			return
		}
		opts.ctx = ctx
	}
}
```

— AI review

pkg/envvars/environment.go
```diff
@@ -20,14 +21,93 @@ const (
...
+
+	// Check the context hadn't been canceled while loading the Bash environment.
+	if ctxErr := options.ctx.Err(); ctxErr != nil {
+		return
+	}
 
 	// this is applied at the end always, as it does not overwrite existing variables
-	defer func() { _ = gotenv.Apply(strings.NewReader(bashOutput)) }() //nolint:errcheck // we can't do anything with the error
+	defer func() {
+		applyErr := gotenv.Apply(strings.NewReader(bashOutput))
```
basti-snyk, line 106: **Should fix — a "canceled" load can still mutate the process environment.**

This deferred `gotenv.Apply(bashOutput)` is registered *before* the later cancellation checkpoints that `return` (the specific-shell check and the config-file loop). So if `ctx` is canceled during those later phases, the function returns "canceled" but this defer still applies the Bash env on the way out — and any `UpdatePath` calls that already ran are left in place too. The deferred logger then reports "Loading configured environment was canceled" despite having partially mutated global state.

`gotenv.Apply` doesn't overwrite existing vars so real-world impact is limited, but the semantics ("canceled" sometimes means no changes, sometimes partial changes) are surprising. Either gate this apply on `options.ctx.Err() == nil`, or document that cancellation is best-effort and may leave partial mutations (and adjust the "applied at the end always" comment, which is now inaccurate).

— AI review

PR #582, opened 2026-04-08
chore: fix transitive dependencies being marked as directly upgradable

CatalinSnyk:
### Description

In certain cases the Remediation Summary would count transitive dependency upgrades as directly upgradable. This would results in incorrect upgrade advice (e.g. `Upgrade from x@1.2.3 to x@1.2.3` - since the actual upgrade would be inside for a nested dependency).

The changes should allign the remediation summary building with the legacy implementation that can be found [here](https://github.com/snyk/cli/blob/b144a8f10b03f298e23bae08c04e3fa4a787faca/src/lib/formatters/remediation-based-format-issues.ts#L27). In terms of tests, I added a few more cases to the remedation testing logic to get the coverage to 85%, but I also slightly updated the `testresults_cli.json` to the newer TestAPI structure in order to add it as a test case for the human readable output.

### Checklist

- [x] Linted (`make lint`)

### DIFF
--- a/internal/ufm_helpers/remediation.go
+++ b/internal/ufm_helpers/remediation.go
@@ -92,6 +92,7 @@ func GetRemediationSummary(issues []testapi.Issue) *RemediationSummary {
 		summary.Pins = append(summary.Pins, pins...)
 	}
 
+	// Filter out unresolved issues that are already covered by a pin.
 	summary.Unresolved = filterUnresolvedCoveredByPins(summary.Unresolved, summary.Pins)
 
 	sort.Slice(summary.Upgrades, func(i, j int) bool {
@@ -173,6 +174,7 @@ func processUpgradeAdvice(issue testapi.Issue, advice testapi.UpgradePackageAdvi
 		}
 
 		for _, depPath := range paths {
+			// If the upgrade path does not match the dependency path, skip it.
 			if len(depPath) < 2 || !pathsMatch(upgradePath.DependencyPath, depPath) {
 				continue
 			}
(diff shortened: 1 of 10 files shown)


internal/ufm_helpers/remediation.go
```diff
@@ -171,20 +174,34 @@ func processUpgradeAdvice(issue testapi.Issue, advice testapi.UpgradePackageAdvi
...
 		for _, depPath := range paths {
+			// If the upgrade path does not match the dependency path, skip it.
 			if len(depPath) < 2 || !pathsMatch(upgradePath.DependencyPath, depPath) {
 				continue
 			}
 
-			matchedPaths++
 			fromPkg := depPath[1]
 			toPkg := upgradePath.DependencyPath[1]
+
```
CatalinSnyk, line 184: This is the main fix to ensure transitive dependency upgrades are not reported as directly upgradable.

internal/ufm_helpers/remediation.go
```diff
@@ -92,6 +92,9 @@ func GetRemediationSummary(issues []testapi.Issue) *RemediationSummary {
 		summary.Pins = append(summary.Pins, pins...)
 	}
 
+	// Filter out unresolved issues that are already covered by a pin.
+	summary.Unresolved = filterUnresolvedCoveredByPins(summary.Unresolved, summary.Pins)
```
CatalinSnyk, line 96: Added this to match the [logic](https://github.com/snyk/cli/blob/b144a8f10b03f298e23bae08c04e3fa4a787faca/src/lib/formatters/remediation-based-format-issues.ts#L85-L91) in the legacy implementation. I didn't add another fixture file in this PR, but I left an action items for adding a Python project one in CLI-1349, and settled for just a unit test case for now.

internal/ufm_helpers/remediation.go
```diff
@@ -319,3 +340,34 @@ func compareVersions(v1, v2 string) int {
...
+	if len(pinnedIssueIDs) == 0 {
+		return unresolved
+	}
+
+	filtered := make([]testapi.Issue, 0, len(unresolved))
+	for _, issue := range unresolved {
+		if pinnedIssueIDs[issue.GetID()] {
+			continue
+		}
+		filtered = append(filtered, issue)
```
PeterSchafer, line 369: Suggestion: It seems that none of the test cases covers the case that something isn't filtered. Let's add this, maybe to one of the existing cases.

PeterSchafer (approved):

PR #583, opened 2026-04-09
fix: remove duplicate rules from SARIF output

danskmt:
### Description

Deduplicates SARIF rules by problem ID in UFM output to fix duplicate rule entries. When multiple findings share the same problem ID (e.g., two `generic-api-key` secrets found in different files), the SARIF output previously emitted one rule per finding, producing duplicate rule IDs in the `rules` array. Per the [SARIF v2.1.0 spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html), rule IDs must be unique within the `rules` array.

The fix adds a `deduplicateIssuesByProblemID` template function that filters the issues list to one entry per unique problem ID (first-wins) for the `rules` loop only. The `results` loop is unchanged — all findings still appear as individual results referencing their shared rule.

**Relevant ticket:** [CLI-1344](https://snyksec.atlassian.net/browse/CLI-1344)

### Checklist

- [x] Linted (`make lint`)

[CLI-1344]: https://snyksec.atlassian.net/browse/CLI-1344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/presenters/funcs.go
+++ b/internal/presenters/funcs.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"os"
 	"reflect"
 	"slices"
 	"strconv"
@@ -386,8 +387,38 @@ func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinf
 	defaultMap["getTargetId"] = func(path string) (string, error) {
 		return target.GetTargetId(path, target.AutoDetectedTargetId, target.WithConfiguredRepository(config))
 	}
+	defaultMap["deduplicateIssues"] = deduplicateIssues
 
 	return defaultMap
+}
+
+type issueDedupeKey string
+
+const (
+	DedupeByProblemID issueDedupeKey = "problemID"
+	DedupeByID        issueDedupeKey = "id"
+)
+
+// deduplicateIssues returns a subset of issues with unique keys, preserving order (first-wins).
+func deduplicateIssues(issues []testapi.Issue, key issueDedupeKey) []testapi.Issue {
+	seen := make(map[string]bool)
+	result := make([]testapi.Issue, 0, len(issues))
+	for _, issue := range issues {
+		var field string
+		switch key {
+		case DedupeByProblemID:
+			field = issue.GetProblemID()
+		case DedupeByID:
+			field = issue.GetID()
+		default:
+			fmt.Fprintf(os.Stderr, "warning: unsupported issueDedupeKey %q, skipping deduplication for issue\n", key)
+		}
+		if field != "" && !seen[field] {
+			seen[field] = true
+			result = append(result, issue)
+		}
+	}
+	return result
 }
 
 func convertTypeToIssueName(findingType testapi.FindingType) string {
--- a/internal/presenters/presenter_ufm_test.go
+++ b/internal/presenters/presenter_ufm_test.go
@@ -131,7 +131,11 @@ func sortByID(arr []interface{}) {
 	})
 }
 
-// sortByRuleID sorts an array of results by their "ruleId" field
+// sortByRuleID sorts results for stable comparison. Tiebreakers (in order):
+//  1. ruleId — primary key
+//  2. artifact URI — needed when multiple findings share the same rule but differ in file
+//  3. fingerprint identity — needed when findings share both rule and file (e.g. same secret
+//     rule detected at different locations in the same file)
 func sortByRuleID(arr []interface{}) {
 	sort.Slice(arr, func(i, j int) bool {
 		iMap, iOk := arr[i].(map[string]interface{})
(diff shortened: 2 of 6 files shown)


internal/presenters/presenter_ufm_test.go
```diff
@@ -141,10 +142,39 @@ func sortByRuleID(arr []interface{}) {
...
+	if !ok {
+		return ""
+	}
+	uri, ok := artLoc["uri"].(string)
+	if !ok {
+		return ""
+	}
+	return uri
+}
+
```
j-luong: question: why did we need to add this change? is there a scenario where we need the results in the same order?

internal/presenters/funcs.go
```diff
@@ -386,10 +386,27 @@ func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinf
...
 	return defaultMap
 }
 
+func deduplicateIssuesByProblemID(issues []testapi.Issue) []testapi.Issue {
+	seen := make(map[string]bool)
+	result := make([]testapi.Issue, 0, len(issues))
+	for _, issue := range issues {
+		id := issue.GetProblemID()
+		if id == "" {
+			id = issue.GetID()
```
j-luong: issue: using a different ID if `GetProblemID` returns empty string doesn't fit the purpose of this fn (`deduplicateIssuesByProblemID`).

If the purpose is specifically to dedupe by Problem ID, then we need to handle `problemID == ""` differently IMO.

internal/presenters/funcs.go
```diff
@@ -386,10 +386,27 @@ func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinf
 	defaultMap["getTargetId"] = func(path string) (string, error) {
 		return target.GetTargetId(path, target.AutoDetectedTargetId, target.WithConfiguredRepository(config))
 	}
+	defaultMap["deduplicateIssuesByProblemID"] = deduplicateIssuesByProblemID
 
 	return defaultMap
 }
 
+func deduplicateIssuesByProblemID(issues []testapi.Issue) []testapi.Issue {
```
j-luong: suggest: instead of implementing a specific dedupe (issues by problem ID), we can implement a more general `deduplicateIssuesByID` that takes an ID as a param

internal/presenters/funcs.go
j-luong: I think instead of `id = issue.GetID()`, we can log a warning and skip dedupe for this iteration

internal/presenters/testdata/ufm/secrets.testresult.json
```diff
@@ -104,7 +104,8 @@
     "_problemRefs": {
       "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee": ["generic-api-key", "CWE-798"],
       "aaaaaaab-bbbb-cccc-dddd-eeeeeeeeeeee": ["generic-api-key-2", "CWE-798"],
-      "399f629d-d638-441c-82a0-379e327dc941": ["generic-secret", "CWE-798"]
+      "399f629d-d638-441c-82a0-379e327dc941": ["generic-secret", "CWE-798"],
+      "cccccccc-bbbb-cccc-dddd-eeeeeeeeeeee": ["generic-api-key", "CWE-798"]
```
j-luong: question: is this correct? i.e. how it's supposed to work

danskmt: The `_problemRefs` map works like this: `findingID → [problemStoreKey1, problemStoreKey2, ...]`. The keys reference entries in the **problemStore**. When two findings reference the **same problem store key**, they get the same problem deserialized, and therefore the same problem ID.

The problem store has one entry for that rule, and both findings reference it. You can see the one in line 105: `aaaaaaab-bbbb-cccc-dddd-eeeeeeeeeeee` and line 108 `cccccccc-bbbb-cccc-dddd-eeeeeeeeeeee`:
- Finding aaaaaaaa... → references generic-api-key → gets problem ID "generic-api-key"
- Finding cccccccc... → references generic-api-key → gets problem ID "generic-api-key"
- **Both produce the same rule ID in SARIF**, which is the duplication issue we're fixing

internal/presenters/funcs.go
```diff
@@ -386,10 +386,38 @@ func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinf
...
+	seen := make(map[string]bool)
+	result := make([]testapi.Issue, 0, len(issues))
+	for _, issue := range issues {
+		var field string
+		switch key {
+		case DedupeByProblemID:
+			field = issue.GetProblemID()
+		case DedupeByID:
+			field = issue.GetID()
+		}
```
j-luong, line 415: suggest: it looks like we invoke this with a string param for the `key`: 
`{{- $uniqueRules := deduplicateIssues $issues "problemID" }}` 

lets add a default case here to handle unsupported `issueDedupeKey`...maybe default to `issue.GetID()`?

```
		switch key {
		case DedupeByProblemID:
			field = issue.GetProblemID()
		default:
			field = issue.GetID()
		}
```

danskmt: I guess defaulting to ID is not good for something that is supposed to be generic

internal/presenters/funcs.go
danskmt: We are now considering that ** GetProblemID** never returns empty string

internal/presenters/presenter_ufm_test.go
danskmt: without it, any test with multiple results sharing a ruleId but different URIs would be flaky

j-luong (approved):

PR #587, opened 2026-04-15
fix: support for rendering multiple locations to SARIF

j-luong:
### Description

This PR refactors the location building logic in sarif.go such that it now considers all possible locations for rendering instead of just the first one.

#### Old
```
{
  "ruleId": "generic-secret",
  "level": "error",
  "message": {
    "text": "This file contains a high severity Generic Secret Key vulnerability."
  },
  "locations": [
    {
      "physicalLocation": {
        "artifactLocation": {
          "uri": ".gitleaksignore"
        },
        "region": {
          "startLine": 53,
          "startColumn": 1,
          "endLine": 53,
          "endColumn": 41
        }
      }
    }
  ],
  "fingerprints": {
    "identity": "53d6c119-b408-585a-9194-a451f33721a0",
    "snyk/asset/finding/v1": "53d6c119-b408-585a-9194-a451f33721a0"
  }
},
```
#### New
```
{
  "ruleId": "generic-secret",
  "level": "error",
  "message": {
    "text": "This file contains a high severity Generic Secret Key vulnerability."
  },
  "locations": [
    {
      "physicalLocation": {
        "artifactLocation": {
          "uri": ".gitleaksignore"
        },
        "region": {
          "startLine": 53,
          "startColumn": 1,
          "endLine": 53,
          "endColumn": 41
        }
      }
    },
    {
      "physicalLocation": {
        "artifactLocation": {
          "uri": ".gitleaksignore"
        },
        "region": {
          "startLine": 54,
          "startColumn": 1,
          "endLine": 54,
          "endColumn": 41
        }
      }
    },
    {
      "physicalLocation": {
        "artifactLocation": {
          "uri": ".gitleaksignore"
        },
        "region": {
          "startLine": 55,
          "startColumn": 1,
          "endLine": 55,
          "endColumn": 41
        }
      }
    },
    many more...
  ],
  "fingerprints": {
    "identity": "53d6c119-b408-585a-9194-a451f33721a0",
    "snyk/asset/finding/v1": "53d6c119-b408-585a-9194-a451f33721a0"
  }
},
```

### Checklist

- [x] Linted (`make lint`)

### DIFF
--- a/internal/presenters/presenter_ufm_test.go
+++ b/internal/presenters/presenter_ufm_test.go
@@ -602,7 +602,7 @@ func Test_UfmPresenter_Sarif(t *testing.T) {
 			ignoreSuppressions: true,
 		},
 		{
-			name:               "secrets_duplicated_rules",
+			name:               "secrets_duplicated_rules_and_multiple_findings",
 			expectedSarifPath:  "testdata/ufm/secrets.duplicated-sarif-rules.sarif.json",
 			testResultPath:     "testdata/ufm/secrets.duplicated-sarif-rules.testresult.json",
 			ignoreSuppressions: true,
--- a/internal/presenters/templates/ufm.sarif.tmpl
+++ b/internal/presenters/templates/ufm.sarif.tmpl
@@ -63,11 +63,7 @@
 				{{- $targetFile := "" }}
 				{{- if $displayTargetFile }} {{- $targetFile = $displayTargetFile }} {{- end }}
 				{{- range $index, $issue := $issues }}
-				{{- $location := buildLocationFromIssue $issue $targetFile }}
-				{{- $physicalLoc := index $location "physicalLocation" }}
-				{{- $logicalLocs := index $location "logicalLocations" }}
-				{{- $region := index $physicalLoc "region" }}
-				{{- $artifactLocation := index $physicalLoc "artifactLocation" }}
+				{{- $locations := buildLocationsFromIssue $issue $targetFile }}
 				{{- $fix := buildFixFromIssue $issue }}
 				{{- $ignoreDetails := $issue.GetIgnoreDetails }}
 				{{- $problemID := $issue.GetProblemID }}
@@ -79,23 +75,26 @@
 						"text": {{ getQuotedString (formatIssueMessage $issue) }}
 					},
 					"locations": [
+						{{- range $locIndex, $location := $locations }}
+						{{- $logicalLocs := $location.LogicalLocations }}
+						{{- $physicalLoc := $location.PhysicalLocation }}
 						{{- if $location }}
 						{
 							{{- if $physicalLoc }}
 							"physicalLocation": {
 								"artifactLocation": {
-									"uri": {{ getQuotedString (index $artifactLocation "uri") }}
+									"uri": {{ getQuotedString $physicalLoc.ArtifactLocation.URI }}
 								},
 								"region": {
-									"startLine": {{ index $region "startLine" }}{{if index $region "startColumn"}},{{end}}
-									{{- if index $region "startColumn" }}
-									"startColumn": {{ index $region "startColumn" }}{{if index $region "endLine"}},{{end}}
+									"startLine": {{ $physicalLoc.Region.StartLine }}{{if $physicalLoc.Region.StartColumn}},{{end}}
+									{{- if $physicalLoc.Region.StartColumn }}
+									"startColumn": {{ $physicalLoc.Region.StartColumn }}{{if $physicalLoc.Region.EndLine}},{{end}}
 									{{- end }}
-									{{- if index $region "endLine" }}
-									"endLine": {{ index $region "endLine" }}{{if index $region "endColumn"}},{{end}}
+									{{- if $physicalLoc.Region.EndLine }}
+									"endLine": {{ $physicalLoc.Region.EndLine }}{{if $physicalLoc.Region.EndColumn}},{{end}}
 									{{- end }}
-									{{- if index $region "endColumn" }}
-									"endColumn": {{ index $region "endColumn" }}
+									{{- if $physicalLoc.Region.EndColumn }}
+									"endColumn": {{ $physicalLoc.Region.EndColumn }}
 									{{- end }}
 								}
 							}
@@ -105,12 +104,14 @@
 							"logicalLocations": [
 								{{- range $logLoc := $logicalLocs }}
 								{
-									"fullyQualifiedName": {{ getQuotedString (index $logLoc "fullyQualifiedName") }}
+									"fullyQualifiedName": {{ getQuotedString $logLoc.FullyQualifiedName }}
 								}
 								{{- end }}
 							]
 							{{- end }}
 						}
+						{{- end }}
+						{{if lt $locIndex (sub (len $locations) 1)}},{{end}}
 						{{- end }}
 					]
 					{{- if and (ne (printf "%s" $issue.GetFindingType) "sca") (ne (printf "%s" $issue.GetFindingType) "license") }},
@@ -121,6 +122,10 @@
 					{{- end }}
 					{{- if $fix }},
 					"fixes": [
+						{{- if gt (len $locations) 0 }}
+						{{- $firstLoc := index $locations 0 }}
+						{{- if $firstLoc.PhysicalLocation }}
+						{{- $firstPhysicalLoc := $firstLoc.PhysicalLocation }}
 						{
 							"description": {
 								"text": {{ getQuotedString (index $fix "description") }}
@@ -128,12 +133,12 @@
 							"artifactChanges": [
 								{
 									"artifactLocation": {
-										"uri": {{ getQuotedString (index $artifactLocation "uri") }}
+										"uri": {{ getQuotedString $firstPhysicalLoc.ArtifactLocation.URI }}
 									},
 									"replacements": [
 										{
 											"deletedRegion": {
-												"startLine": {{ index $region "startLine" }}
+												"startLine": {{ $firstPhysicalLoc.Region.StartLine }}
 											},
 											"insertedContent": {
 												"text": {{ getQuotedString (index $fix "packageVersion") }}
@@ -143,6 +148,8 @@
 								}
 							]
 						}
+						{{- end }}
+						{{- end }}
 					]
 					{{- end }}
 					{{- if $ignoreDetails }},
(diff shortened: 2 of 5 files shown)


j-luong:
> [Interface Breaking Change](https://github.com/snyk/go-application-framework/pull/587/changes#diff-93189bb9079749ed60046d0914f05bc216fcd7baf7bbefd3c40a5727853ddb96R274-R274) 🔴 [critical]
The template function buildLocationFromIssue was renamed to buildLocationsFromIssue. While ufm.sarif.tmpl was updated, the cross-file context shows that internal/presenters/presenter_local_finding.go also uses getSarifTemplateFuncMap() to render local_finding.sarif.tmpl. Since that template is not updated in this PR, rendering local findings to SARIF will now fail with an 'execution error: function "buildLocationFromIssue" not defined'.

`local_finding.sarif.tmpl` does not use `buildLocationFromIssue` so the refactor would not impact the rendering

internal/presenters/templates/ufm.sarif.tmpl
```diff
@@ -63,11 +63,9 @@
 				{{- $targetFile := "" }}
 				{{- if $displayTargetFile }} {{- $targetFile = $displayTargetFile }} {{- end }}
 				{{- range $index, $issue := $issues }}
-				{{- $location := buildLocationFromIssue $issue $targetFile }}
-				{{- $physicalLoc := index $location "physicalLocation" }}
-				{{- $logicalLocs := index $location "logicalLocations" }}
-				{{- $region := index $physicalLoc "region" }}
-				{{- $artifactLocation := index $physicalLoc "artifactLocation" }}
+				{{- $locations := buildLocationsFromIssue $issue $targetFile }}
+				{{- $firstLoc := index $locations 0 }}
```
CatalinSnyk: suggestion: since this is now only used for the fixes, maybe we relocate this assignments closer to that. As the PR bot also mentioned we might also wanna do some nil checks.

internal/presenters/presenter_ufm_test.go
```diff
@@ -607,6 +607,12 @@ func Test_UfmPresenter_Sarif(t *testing.T) {
 			testResultPath:     "testdata/ufm/secrets.duplicated-sarif-rules.testresult.json",
 			ignoreSuppressions: true,
 		},
+		{
```
CatalinSnyk: suggestion: Since the two test results files are the same, maybe we use just one of them 😄 ? The [secrets.duplicated-sarif-rules.sarif.json](https://github.com/snyk/go-application-framework/pull/587/changes#diff-a2ef8effdefef4181782116641e8a1b71852f5fc62391a51382796ba23db4e8b) was already updated so I would say we don't actually need the same test case essentially.

pkg/utils/sarif/sarif.go
```diff
@@ -493,82 +493,199 @@ func appendDescriptionSection(sb *strings.Builder, issue testapi.Issue) {
 	}
 }
 
-// BuildLocation constructs a SARIF location object from issue data
-//
-//nolint:gocyclo // needs the global state for package name and version
-func BuildLocation(issue testapi.Issue, targetFile string) map[string]interface{} {
-	// Extract first finding from issue
+type regionData struct {
```
CatalinSnyk, line 496: suggestion: I would say we would like to avoid the use of another set of data structures here. How about we do something like:
```go
var locations []map[string]string

func buildSarifLocations(...) []map[string]string

location = make(map[string]string]
location["startLine"] = ...
location["endling"] = ...
location["uri"] = ...
```

And then in the template we can just use a simple indexing for the map to render things. Should still get the same result, just without the extra structures, just a flat map to hold everything in place.  WDYT?

CatalinSnyk (commented):
Left a couple of drive-by suggestions.

pkg/utils/sarif/sarif.go
j-luong: as a counter point, type safety is good. `[]map[string]string` isn't great for reasoning with the data structure and increases the possibility of mistakes being introduced.

internal/presenters/templates/ufm.sarif.tmpl
```diff
@@ -143,6 +148,8 @@
 								}
 							]
 						}
+						{{- end }}
+						{{- end }}
```
CatalinSnyk, line 152: small nit: less indent on the second one 😄

pkg/utils/sarif/sarif.go
```diff
@@ -493,82 +493,199 @@ func appendDescriptionSection(sb *strings.Builder, issue testapi.Issue) {
...
-		loc := finding.Attributes.Locations[0]
-
-		// Try source location first
-		sourceLoc, err := loc.AsSourceLocation()
-		if err == nil && sourceLoc.FilePath != "" {
-			uri = sourceLoc.FilePath
-			region["startLine"] = sourceLoc.FromLine
-			if sourceLoc.FromColumn != nil {
-				region["startColumn"] = *sourceLoc.FromColumn
+		for _, sLoc := range sLocations {
```
CatalinSnyk, line 550: suggestion: I don't think this whole logic is needed? we can just add a check for the issue finding type and just return early again if it's SCA. And for the rest of the findings I don't think there's a needed for the deduplication logic, since it's going to be one finding/issue, with multiple locations.

pkg/utils/sarif/sarif.go
```diff
@@ -493,82 +493,204 @@ func appendDescriptionSection(sb *strings.Builder, issue testapi.Issue) {
...
+	if findingType == testapi.FindingTypeSca {
+		scaFindingLocation := sarifLocation{
+			LogicalLocations: []logicalLocation{buildLogicalLocation(packageName, packageVersion)},
+		}
+		if targetFile != "" {
+			pLocation := buildPhysicalLocation(targetFile, regionData{StartLine: 1})
+			scaFindingLocation.PhysicalLocation = &pLocation
+		}
+		return []sarifLocation{scaFindingLocation}
 	}
```
danskmt, line 541: would be possible to not have this "special case" here and just use the code that follows it? Also, here forces StartLine to be 1 always. Another question is: in what situations targetFile is empty? Do we have any tests that are like this? Or is this an untested block at the moment?

pkg/utils/sarif/sarif.go
```diff
@@ -493,82 +493,204 @@ func appendDescriptionSection(sb *strings.Builder, issue testapi.Issue) {
...
-			"artifactLocation": map[string]interface{}{
-				"uri": uri,
-			},
-			"region": region,
+	// if no locations were generated
+	if len(sLocations) == 0 {
+		if fallback := buildFallbackSarifLocation(targetFile, packageName, packageVersion); fallback != nil {
+			sLocations = append(sLocations, *fallback)
 		}
 	}
```
danskmt, line 602: the condition `len(sLocations) == 0` can be suppressed, and actually here we have two ifs that could be just one.

Note that the step before it adds values to **sLocations**, so this logic could be inside that previous `if hasLocations && targetFile == ""`.

Or even better, make a early return if that first condition is false, so avoid nesting things unnecessarily

pkg/utils/sarif/sarif.go
```diff
@@ -493,82 +493,204 @@ func appendDescriptionSection(sb *strings.Builder, issue testapi.Issue) {
...
+			pVersion = pkgLoc.Package.Version
+		}
+	}
+
+	if pName != "" || pVersion != "" {
+		return sarifLocation{LogicalLocations: []logicalLocation{buildLogicalLocation(pName, pVersion)}}
+	}
+
+	return buildSarifLocationFromSourceLoc(loc)
+}
```
danskmt, line 627: Can we get rid of these fallbacks? I mean, this flow is enabling 4 situations:
1) **Package Name** and **Package Version** both present (Good)
2) **Package Name** present BUT **Package Version** not present (hmmm?)
3) **Package Name** not present BUT **Package Version** present (????)
4) None is present, so we just add the location.

I believe we should consider maybe 1 and 4 only?

pkg/utils/sarif/sarif.go
j-luong: We utilise the issue finding type here to perform this check early, this saves us having to loop through the findings later on.

The start line default is a carryover from the original implementation:
```
// Default to line 1 for manifest files
uri := targetFile
region := map[string]interface{}{
	"startLine": 1,
}
```

The target file is passed via the `ufm.sarif.tmpl`, which gets it from the metadata of the test API and this could be empty.

The SCA tests in `presenter_ufm_test.go` provide an integration test to validate this logic

pkg/utils/sarif/sarif.go
j-luong: Could you give a code suggestion to help explain?

> the condition len(sLocations) == 0 can be suppressed, and actually here we have two ifs that could be just one.

I'm assuming you mean something like: `if len(sLocations) == 0 && buildFallbackSarifLocation(...) != nil {` which may call `buildFallbackSarifLocation` unnecessarily, hence the reason to nest.

Generally, the intention here is to:
1. Try and build `[]sarifLocation` from the locations in the finding attributes
2. Fallback to building the `[]sarifLocation` from other properties if 1 isn't possible

I don't think these two can be combined

pkg/utils/sarif/sarif.go
j-luong: probably not without unintended side effects. We try to build the location data from various different properties, which must be inspected.

pkg/utils/sarif/sarif.go
danskmt: hmmm yeah, that's true. 
Sorry I mixed two comments in one, and the second part is actually incorrect.
I think I got confused with the fallback stuff, so I was actually wrong. I thought it could be simplified somehow, but looking at it again it is ok the way it is, so sorry about that.

pkg/utils/sarif/sarif.go
danskmt: Hmm ok, so it is not something we need to review... Forcing the value like this is not straightforward to understand.
But thank you for the explanation!

pkg/utils/sarif/sarif.go
danskmt: It is fine to have but it just feels too many fallbacks that might be unrealistic (_or may not, as you've said_). If we could confirm that we always either have both, or doesn't have both at the same time, would be nice. But not a problem having it either.

danskmt (approved):

CatalinSnyk (approved):

PR #588, opened 2026-04-17
fix: improve API endpoint regex [IDE-1896]

rrama:
### Description

Ensure the endpoint is restricted to Snyk domains.

### Checklist

 - N/A
- [x] Linted (`make lint`)
  - http://github.com/snyk/cli/pull/6733

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -5,4 +5,4 @@ const SNYK_DEFAULT_API_VERSION = "2024-04-22"
 const SNYK_DEFAULT_IN_MEMORY_THRESHOLD_MB = 512 * 1024 * 1024
 const SNYK_DOCS_URL = "https://docs.snyk.io"
 const SNYK_DOCS_ERROR_CATALOG_PATH = "/scan-with-snyk/error-catalog"
-const SNYK_DEFAULT_ALLOWED_HOST_REGEXP = `^(https?://)?api(\.(.+))?\.(snyk|snykgov)\.io$`
+const SNYK_DEFAULT_ALLOWED_HOST_REGEXP = `^(https?:\/\/)?api(\.([a-z0-9._-]+))?\.(snyk|snykgov)\.io$`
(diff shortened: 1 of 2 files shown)


bastiandoetsch (dismissed):

internal/constants/constants.go
```diff
@@ -5,4 +5,4 @@ const SNYK_DEFAULT_API_VERSION = "2024-04-22"
 const SNYK_DEFAULT_IN_MEMORY_THRESHOLD_MB = 512 * 1024 * 1024
 const SNYK_DOCS_URL = "https://docs.snyk.io"
 const SNYK_DOCS_ERROR_CATALOG_PATH = "/scan-with-snyk/error-catalog"
-const SNYK_DEFAULT_ALLOWED_HOST_REGEXP = `^(https?://)?api(\.(.+))?\.(snyk|snykgov)\.io$`
+const SNYK_DEFAULT_ALLOWED_HOST_REGEXP = `^(https?:\/\/)?api(\.([a-z0-9._-]+))?\.(snyk|snykgov)\.io$`
```
bastiandoetsch, line 8: This would prevent snyk to use private cloud unicode URLs.

It would be better to update `IsValidHost` with `url.Parse` to extract the actual host. The function could be moved to its own package to prevent possible import cycles.

bastiandoetsch (commented):
Added a comment for improvement

internal/constants/constants.go
rrama: Maybe I don't want to allow people to spin up PCs with unicode characters and this is my way of discouraging their use. 😉

But in all seriousness, you are right that since this regex is only used in [a few places](https://github.com/search?q=org%3Asnyk%20CONFIG_KEY_ALLOWED_HOST_REGEXP&type=code), it may be better to scrap it entirely and have a function which parses the URL and checks the beginning subdomain is "api" and the actual domain is in a list of allowed domains (the list could be stored in a GAF configuration like this regex).

rrama:
Closing, since we decided there is a better way of doing it. (I updated the Jira ticket to say)

PR #592, opened 2026-04-21
feat: Render WebUI report URL for secrets --report [CLI-1429]

danskmt:
### Description

Adds WebUI project page link rendering to the UFM presenter templates when `--report` is enabled for secret scans. The `cli-extension-secrets` extension populates a `project-page-link` metadata key on the `TestResult`; this PR consumes it.

**SARIF template (`ufm.sarif.tmpl`)**
- Adds a conditional `properties.uploadResult` block containing `reportUrl`, emitted only when `metadata["project-page-link"]` is present.

**Human-readable template (`ufm.human.tmpl`)**
- Adds a "Report" footer block in `resultFooter` that displays "Your test results are available at: \<bold URL\>" when `metadata["project-page-link"]` is present.

**Test fixtures**
- Adds `secrets.with-report.testresult.json` — scrubbed from a real `snyk secrets test --report` run with deterministic IDs/timestamps and redacted org/user data.
- Adds matching `secrets.with-report.sarif.json` and `secrets.with-report.human.readable` expected files.

**Where should the reviewer start?**

- `internal/presenters/templates/ufm.human.tmpl` — the "Report" block in `resultFooter`
- `internal/presenters/templates/ufm.sarif.tmpl` — the `properties.uploadResult` block
- `internal/presenters/presenter_ufm_test.go` — `secrets_with_report` table entries

**Risk assessment:** Low — templates only emit the new block when `project-page-link` metadata is present; no change for non-report scans.

### Checklist

- [x] Linted (`make lint`)
  
  PR in the CLI: https://github.com/snyk/cli/pull/6740

### DIFF
--- a/internal/presenters/presenter_ufm_test.go
+++ b/internal/presenters/presenter_ufm_test.go
@@ -607,6 +607,12 @@ func Test_UfmPresenter_Sarif(t *testing.T) {
 			testResultPath:     "testdata/ufm/secrets.duplicated-sarif-rules.testresult.json",
 			ignoreSuppressions: true,
 		},
+		{
+			name:               "secrets_with_report",
+			expectedSarifPath:  "testdata/ufm/secrets.with-report.sarif.json",
+			testResultPath:     "testdata/ufm/secrets.with-report.testresult.json",
+			ignoreSuppressions: true,
+		},
 	}
 
 	for _, tc := range testCases {
@@ -1200,6 +1206,13 @@ func Test_UfmPresenter_HumanReadable(t *testing.T) {
 			includeIgnores:    true,
 			severityThreshold: "",
 		},
+		{
+			name:              "secrets_with_report",
+			expectedPath:      "testdata/ufm/secrets.with-report.human.readable",
+			testResultPath:    "testdata/ufm/secrets.with-report.testresult.json",
+			includeIgnores:    true,
+			severityThreshold: "",
+		},
 	}
 
 	for _, tc := range testCases {
@@ -1228,6 +1241,7 @@ func Test_UfmPresenter_HumanReadable(t *testing.T) {
 
 			actualBytes := writer.Bytes()
 			assert.NotEmpty(t, actualBytes)
+
 			assert.NotEmpty(t, expectedBytes)
 			assert.Equal(t, string(expectedBytes), string(actualBytes))
 		})
--- a/internal/presenters/templates/ufm.human.tmpl
+++ b/internal/presenters/templates/ufm.human.tmpl
@@ -292,6 +292,15 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
     {{- /* Render test summary */}}
     {{- box (renderToString "testSummary" $) }}
 
+    {{- /* Render report link when available */ -}}
+    {{- with $.GetMetadataValue "report-url" -}}
+        {{- "\n" -}}
+        {{- "Report" | bold -}}
+        {{- "\n" -}}
+        {{- "Your test results are available at: " -}}
+        {{- printf "%s" . | bold -}}
+    {{- end -}}
+
     {{- "\n" }}
     {{- with $.GetMetadataValue "target-directory" }}
 	    {{- $targetId := getTargetId . }}
--- a/internal/presenters/templates/ufm.human.tmpl
+++ b/internal/presenters/templates/ufm.human.tmpl
@@ -293,12 +293,12 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
     {{- box (renderToString "testSummary" $) }}
 
     {{- /* Render report link when available */ -}}
-    {{- with $.GetMetadataValue "report-url" -}}
-        {{- "\n" -}}
-        {{- "Report" | bold -}}
-        {{- "\n" -}}
-        {{- "Your test results are available at: " -}}
-        {{- printf "%s" . | bold -}}
+    {{- with $.GetMetadataValue "report-url" }}
+        {{- "\n\n" }}
+        {{- "Report" | bold }}
+        {{- "\n\n" }}
+        {{- "  Your test results are available at:\n" }}
+        {{- printf "  %s" ( . | bold ) }}
     {{- end -}}
 
     {{- "\n" }}
--- a/internal/presenters/templates/ufm.sarif.tmpl
+++ b/internal/presenters/templates/ufm.sarif.tmpl
@@ -222,6 +222,13 @@
 					{{- end }}
 				},
 				*/ -}}
+				{{- with index $metadata "report-url" }}
+				"properties": {
+					"uploadResult": {
+						"reportUrl": {{ getQuotedString (printf "%s" .) }}
+					}
+				},
+				{{- end }}
 				"automationDetails": {
 					{{- $projectName := or (index $metadata "project-name") "" }}
 					{{- if not $projectName }} {{- $projectName = printf "%d" $resultIndex }} {{- end }}
--- a/internal/presenters/testdata/ufm/secrets.with-report.testresult.json
+++ b/internal/presenters/testdata/ufm/secrets.with-report.testresult.json
@@ -1,0 +1,355 @@
+[
+  {
+      "testId": "11111111-1111-4111-a111-111111111111",
+      "testConfiguration": {
+          "local_policy": {
+              "fail_on_upgradable": false,
+              "severity_threshold": "low",
+              "suppress_pending_ignores": false
+          },
+          "project_environment": [],
+          "project_lifecycle": [],
+          "project_tags": [],
+          "publish_report": true,
+          "scan_config": {
+              "secrets": {}
+          },
+          "timeout": {
+              "outcome": "fail",
+              "seconds": 1200
+          }
+      },
+      "createdAt": "2026-04-21T13:27:46.081538Z",
+      "executionState": "finished",
+      "passFail": "fail",
+      "outcomeReason": "policy_breach",
+      "effectiveSummary": {
+          "count": 4,
+          "count_by": {
+              "result_type": {
+                  "secrets": 4
+              },
+              "severity": {
+                  "critical": 1,
+                  "high": 3,
+                  "low": 0,
+                  "medium": 0
+              }
+          }
+      },
+      "rawSummary": {
+          "count": 5,
+          "count_by": {
+              "result_type": {
+                  "secrets": 5
+              },
+              "severity": {
+                  "critical": 1,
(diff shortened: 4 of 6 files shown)


internal/presenters/templates/ufm.sarif.tmpl
```diff
@@ -222,6 +222,15 @@
 					{{- end }}
 				},
 				*/ -}}
+				{{- $reportUrl := or (index $metadata "project-page-link") "" }}
```
danskmt: Initially I interpreted of having a direct link to the history / snapshot URL, e.g. 
https://app.dev.snyk.io/org/platform_hammerhead_testing/project/bf44b59c-627b-483c-9247-4bb97aec114d/history/a5e3d271-5796-4b21-af6b-6c9d227b8c88

However, the test API’s run identifier (TestID) does not match the Web UI’s history / snapshot id. 
After talking to Catalin, we'll be using just the existing field project-page-link that contains the reference to the project and thus brings to the Overview page that is the latest run, i.e.:
https://app.dev.snyk.io/org/platform_hammerhead_testing/project/bf44b59c-627b-483c-9247-4bb97aec114d

In a CI environment we might want to have the exact link to the history item (I believe it is desirable), but at the moment we are not able to get such value. So for now, we are going to use `project-page-link`.

CatalinSnyk: suggestion: I think the `or` is redundant here. The `index` function would return a 0 value which if the key is not there.

internal/presenters/templates/ufm.sarif.tmpl
```diff
@@ -222,6 +222,15 @@
...
+				{{- if $reportUrl }}
+				"properties": {
+					"uploadResult": {
+						"projectId": {{ getQuotedString (printf "%s" (or (index $metadata "project-id") "")) }},
```
CatalinSnyk: suggestion: same suggestion as above regarding the `or`

internal/presenters/presenter_ufm_test.go
```diff
@@ -27,6 +27,14 @@ import (
 	"github.com/snyk/go-application-framework/pkg/utils/ufm"
 )
 
+const regenerateTestResults = false
```
CatalinSnyk: nit: I would say this could be done in the other ticket around ensuring the fixtures files are up to date with the TestAPI spec. Maybe we keep these out for now?

internal/presenters/presenter_ufm_test.go
```diff
@@ -561,6 +569,7 @@ func Test_UfmPresenter_Sarif(t *testing.T) {
 
 	testCases := []struct {
 		name               string
+		source             string // optional: repo URL used to generate the fixture
```
CatalinSnyk: nit: ditto as above.

internal/presenters/testdata/ufm/secrets.with-report.testresult.json
```diff
@@ -0,0 +1,355 @@
...
+                  ],
+                  "policy_modifications": [],
+                  "problems": null,
+                  "rating": {
+                      "severity": "critical"
+                  },
+                  "risk": {},
+                  "title": "Stripe Secret Key"
+              },
+              "id": "00000001-0001-4001-a001-000000000001",
```
CatalinSnyk, line 149: question/suggestion: I think this mapping in the `_problemRefs`  is not correct. What we normally call `FindingID` (you can also look into the human readable snapshots from the tests) is derived from the `"key":"<value>` propriety. This is also noted in the TestAPI generated file like so:
https://github.com/snyk/go-application-framework/blob/828033ca578853ac9da0c19e0433a5e2642dcb50/pkg/apiclients/testapi/testapi.gen.go#L858-L862

This `FindingAttributes.Id` identifier is more or less not used by our presenter. So the problem store should hold the references for the values from the `"key"` propriety.

internal/presenters/templates/ufm.sarif.tmpl
```diff
@@ -222,6 +222,14 @@
 					{{- end }}
 				},
 				*/ -}}
+				{{- $reportUrl := index $metadata "project-page-link" }}
```
CatalinSnyk: super-small nit: Could have probably used the `with index $metadata "project-page-link" ....` Like the usage in the human readable template to get this to a 1 line conditional.

internal/presenters/templates/ufm.human.tmpl
```diff
@@ -292,6 +292,16 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
     {{- /* Render test summary */}}
     {{- box (renderToString "testSummary" $) }}
 
+    {{- /* Render report link when available */ -}}
+    {{- with $.GetMetadataValue "project-page-link" -}}
+        {{- "\n" -}}
+        {{- "Report" | bold -}}
+        {{- "\n" -}}
+        Your test results are available at:
+        {{- " " -}}
```
CatalinSnyk: question/nit: Can't this space be at the end of the sentence above? We could also add it in the `printf` statement, but I think it's not necessary.

internal/presenters/templates/ufm.human.tmpl
```diff
@@ -292,6 +292,16 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
     {{- /* Render test summary */}}
     {{- box (renderToString "testSummary" $) }}
 
+    {{- /* Render report link when available */ -}}
+    {{- with $.GetMetadataValue "project-page-link" -}}
```
PeterSchafer: question: From a semantical standpoint, why don't we call this link `report-link` or `test-result-uri`? `project-page-link` seems to be a single temporary use case.

internal/presenters/testdata/ufm/secrets.with-report.testresult.json
danskmt: Hmm Attributes key is different from Findings ID. I see that in _human readable_ you are using Finding ID [from attributes.key](https://github.com/snyk/go-application-framework/blob/b9ea94f06443a5d09a1a04fd8d277938116001f1/pkg/apiclients/testapi/issues.go#L791), 
image

And in [this flow](https://github.com/snyk/go-application-framework/blob/a1fb6f13accd32148cc7fcf290ddae1eb5a096fc/pkg/utils/ufm/json_test_result.go#L298) it is by finding.id:
image

danskmt: you mean for us to change this flow..? I didn't get much what you mean in your comment

CatalinSnyk: Ah, good catch, my bad on this one. I looked at https://github.com/snyk/go-application-framework/blob/828033ca578853ac9da0c19e0433a5e2642dcb50/internal/presenters/testdata/ufm/secrets.duplicated-sarif-rules.testresult.json#L99-L124 and I thought the key was also used in the problemRef/store structs. The comment can be disregarded.

internal/presenters/templates/ufm.human.tmpl
CatalinSnyk: Well, it is indeed more of a temporary fix/placeholder until the API would support this flow better with returning the URL and snapshot ID in a more direct way. And technically, the report link is a link to the WebUI project page 🤷 

I don't have any hard feelings around this one, so if you think it's worth to rename (maybe also moving it to a constant?), it can definitely be done. But I'm also happy with keeping this one like this as well for the time being.

danskmt: I renamed it to `report-url` to be consistent with the rest.
But yeah, I thought about keeping the original name of what was indeed being used... Maybe we should map a task to not forget about this needs to be fixed at some point?

CatalinSnyk: Good, but we might wanna raise a PR to also update the [current usage](https://github.com/snyk/cli-extension-secrets/blob/0648d310ddd78cdf903330415d62b31ee9f0f67a/internal/commands/secretstest/command.go#L286) or at least reach out about this. (after this gets merged)

internal/presenters/testdata/ufm/secrets.with-report.testresult.json
danskmt: No worries, thanks for the comment!

internal/presenters/templates/ufm.sarif.tmpl
danskmt: Thanks! Changed it

danskmt:
```
{{- with index $metadata "report-url" }}
```

internal/presenters/templates/ufm.human.tmpl
danskmt: Thanks, fixed it

internal/presenters/templates/ufm.human.tmpl
PeterSchafer: The reason why I mentioned this is, that the current limitation affects secrets only. imagine next comes OSS and they might have the correct URL, then they would be using a semantically incorrect name. Also please think of this as an API. changing the name is a breaking change.

danskmt: Related PR for this change:
https://github.com/snyk/cli-extension-secrets/pull/69

internal/presenters/templates/ufm.human.tmpl
```diff
@@ -292,6 +292,15 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
     {{- /* Render test summary */}}
     {{- box (renderToString "testSummary" $) }}
 
+    {{- /* Render report link when available */ -}}
+    {{- with $.GetMetadataValue "report-url" -}}
+        {{- "\n" -}}
```
CatalinSnyk: issue: I think this is not working as expected to the the last `-}}`. I just tested the output and the newline is not rendered.

internal/presenters/templates/ufm.human.tmpl
```diff
@@ -292,6 +292,15 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
...
+        {{- "Report" | bold -}}
+        {{- "\n" -}}
```
CatalinSnyk: Same as above.

internal/presenters/templates/ufm.human.tmpl
CatalinSnyk: You can also see it in the generated snapshot.

CatalinSnyk: Image

CatalinSnyk (approved):

PR #599, opened 2026-05-01
chore: cleans up TestResult interface

j-luong:
### Description

This PR updates the `testapi` `TestResult` interface by introducing a general `Get()` getter method and deprecating several specific getters from the interface. The `Get()` method will eventually replace the deprecated ones.

The outdated getters were deprecated instead of removed so as not to introduce a breaking change and block this PR from being merged. This would allow dependencies to update their implementations before we eventually remove the deprecated methods.

### Checklist

- [x] Linted (`make lint`)

### DIFF
--- a/pkg/apiclients/testapi/testapi.go
+++ b/pkg/apiclients/testapi/testapi.go
@@ -20,6 +20,19 @@ import (
 
 //go:generate go tool github.com/golang/mock/mockgen -source=testapi.go -destination ../mocks/testapi.go -package mocks -imports testapi=github.com/snyk/go-application-framework/pkg/apiclients/testapi
 
+// TestResultKeys represents the keys for accessing test result data.
+type TestResultKeys string
+
+const (
+	TestResultTestSubject      TestResultKeys = "subject"
+	TestResultSubjectLocators  TestResultKeys = "subject_locators"
+	TestResultTestResources    TestResultKeys = "resources"
+	TestResultRawSummary       TestResultKeys = "raw_summary"
+	TestResultTestFacts        TestResultKeys = "test_facts"
+	TestResultBreachedPolicies TestResultKeys = "breached_policies"
+	TestResultMetadata         TestResultKeys = "metadata"
+)
+
 // config holds configuration for the test API client, set using ConfigOption functions.
 type config struct {
 	PollInterval          time.Duration // Default: 2000ms, Min: 1000ms
@@ -107,8 +120,14 @@ type TestResult interface {
 	GetTestID() *uuid.UUID
 	GetTestConfiguration() *TestConfiguration
 	GetCreatedAt() *time.Time
+
+	Get(key TestResultKeys) interface{}
+
+	// Deprecated: Use Get(testapi.TestResultTestSubject) instead.
 	GetTestSubject() *TestSubject
+	// Deprecated: Use Get(testapi.TestResultSubjectLocators) instead.
 	GetSubjectLocators() *[]TestSubjectLocator
+	// Deprecated: Use Get(testapi.TestResultTestResources) instead.
 	GetTestResources() *[]TestResource
 
 	GetExecutionState() TestExecutionStates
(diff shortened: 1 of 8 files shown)


pkg/utils/ufm/json_test_result.go
```diff
@@ -60,16 +82,22 @@ func (j *jsonTestResult) GetCreatedAt() *time.Time {
 	return j.CreatedAt
 }
 
+// Deprecated: use Get(testapi.TestResultTestSubject) instead
+//
 // GetTestSubject returns the test subject.
 func (j *jsonTestResult) GetTestSubject() *testapi.TestSubject {
 	return j.TestSubject
 }
```
danskmt, line 90: I believe we should have PRs for making these changes in the repos that are using this? e.g. https://github.com/snyk/cli-extension-secrets/pull/59/changes
https://github.com/snyk/cli-extension-os-flows/pull/223/changes

pkg/apiclients/testapi/testapi.go
```diff
@@ -102,13 +102,31 @@ type client struct {
...
+
+const (
+	TestResultTestSubject      TestResultKeys = "subject"
+	TestResultSubjectLocators  TestResultKeys = "subject_locators"
+	TestResultTestResources    TestResultKeys = "resources"
+	TestResultRawSummary       TestResultKeys = "raw_summary"
+	TestResultTestFacts        TestResultKeys = "test_facts"
+	TestResultBreachedPolicies TestResultKeys = "breached_policies"
+	TestResultMetadata         TestResultKeys = "metadata"
+)
```
danskmt: wondering if would make sense having these constants at the top of the file, or maybe somewhere else (separate file) specific for these definitions.

pkg/apiclients/testapi/testapi.go
```diff
@@ -117,14 +135,18 @@ type TestResult interface {
...
 	GetEffectiveSummary() *FindingSummary
+	// Deprecated: Use Get(testapi.TestResultRawSummary) instead.
 	GetRawSummary() *FindingSummary
+	// Deprecated: Use Get(testapi.TestResultTestFacts) instead.
 	GetTestFacts() *[]TestFact
 
 	SetMetadata(key string, value interface{})
 	GetMetadataValue(key string) interface{}
+	// Deprecated: Use Get(testapi.TestResultMetadata) instead.
 	GetMetadata() map[string]interface{}
```
danskmt, line 151: since GetMetadata is being deprecated, should GetMetadataValue also be?

danskmt: Maybe should change this to use the `Get(testapi.TestResultMetadata)` instead of direct access to  `r.metadata[key]`?
```go
// GetMetadataValue returns the metadata value for the given key.
func (r *testResult) GetMetadataValue(key string) interface{} {
	return r.metadata[key]
}
```

danskmt (approved):
LGTM - Left some comments

pkg/utils/ufm/json_test_result.go
j-luong: Thanks, I can update those PRs instead of raising new ones

pkg/apiclients/testapi/testapi.go
j-luong: I've no strong opinion, I'll move it to the top (as it's more common) instead of keeping it close to where it's used

pkg/apiclients/testapi/testapi.go
j-luong: I think it's ok for the public interface (`GetMetaDataValue`) to access the internal implementation (`r.metadata`) like this - or did you mean something else?

j-luong: > since GetMetadata is being deprecated, should GetMetadataValue also be?

I had considered this, but IMO this `GetMetadataValue` is the cleanest way to access metadata. Is there a better alternative?

danskmt: I guess is more like "What if we change the way of accessing this value", i.e. **GetMetadata()** previously would return `r.metadata`, so actually the implementation of **GetMetadataValue** should be at that point:
```
k := r.GetMetadata()
return k[key]
```
I know that can be "a little too much", but it is a bit different as it transfers the responsibility to GetMetadata function.
I'm ok with any approach you decide suits best, both works.

PR #601, opened 2026-05-06
chore: remove usage of deprecated functions

j-luong:
### Description

This PR removes the usage of deprecated methods from the `TestResult` interface. Extra type checks are included to address a potential concern:

> In NewSerializableTestResult, the code uses comma-ok type assertions (e.g., tr.Get(testapi.TestResultTestSubject).(*testapi.TestSubject)) to populate the result fields. If the Get() method of the source TestResult returns a non-pointer type or a type that doesn't exactly match the assertion, the field is silently skipped and remains nil. This results in data loss in the serialized output without any error or log, which is a regression from the previous direct method calls like tr.GetTestSubject() which were type-safe at compile time.

### Checklist

- [x] Linted (`make lint`)


pkg/utils/ufm/json_test_result.go
```diff
@@ -220,28 +220,61 @@ func NewSerializableTestResult(ctx context.Context, tr testapi.TestResult) (test
...
+	if err != nil {
+		return nil, err
 	}
 
 	return result, nil
 }
 
+// getResultProperty extracts a property from the test result and returns it with proper type checking.
+// This is a generic helper function that safely extracts typed values from interface{}.
+func getResultProperty[T any](raw any, property testapi.TestResultKeys) (T, error) {
```
CatalinSnyk, line 266: suggestion/question: I'm wondering if it's worth moving this generic into the `.Get` method directly, just to make it easier for the testapi consumers. WDYT?

pkg/utils/ufm/json_test_result.go
CatalinSnyk, line 219: Maybe a bit unrelated to the PR, but it seems we never set the `TestResources` propriety here.

j-luong: would it make sense to include it in this PR? IMO a quick follow up PR makes more sense

pkg/utils/ufm/json_test_result.go
j-luong: could you give an example? the `.Get` returns `any` and its the reason why we need these type checks. As far as I'm aware, interface methods can't have generics so:

```
// TestResult defines the contract for accessing test result information.
type TestResult interface {
	Get[T any](key TestResultKeys) (T, error)
    ...
```

Is not allowed

CatalinSnyk: Ah, I keep forgetting about that... my bad!

CatalinSnyk: I was just thinking from an ergonomic POV that most of the time you would have the type cast check after calling the get method, and using the generic you could maybe get away with just one.

pkg/utils/ufm/json_test_result.go
CatalinSnyk: I don't have any preferences on it, just thought I would mention it since it looked like a gap in the struct.

CatalinSnyk (approved):

PR #604, opened 2026-05-08
feat: add handling for `X-RateLimit-Reset` header in retry middleware…

robertolopezlopez:
### Description

X-RateLimit-Reset response header support.

When the user runs into rate limiting issues, Envoy returns a 429 response with this header. According to envoy docs and confirmed by platform team, it has numeric int format.

### Checklist

- [x] Linted (`make lint`)

[CLI-1452]: https://snyksec.atlassian.net/browse/CLI-1452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -134,7 +134,9 @@ func errFromStatusCode(code int) error {
 	case http.StatusInternalServerError:
 		return snyk.NewServerError("Internal server error.")
 	case http.StatusTooManyRequests:
-		return snyk.NewTooManyRequestsError("Too many requests.")
+		err := snyk.NewTooManyRequestsError("Too many requests.")
+		err.Level = "error"
+		return err
 	default:
 		return nil
 	}
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -133,6 +133,8 @@ func errFromStatusCode(code int) error {
 		return snyk.NewBadRequestError("The request cannot be processed.")
 	case http.StatusInternalServerError:
 		return snyk.NewServerError("Internal server error.")
+	case http.StatusTooManyRequests:
+		return snyk.NewTooManyRequestsError("Too many requests.")
 	default:
 		return nil
 	}
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -133,13 +133,20 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	backoffMethod.InitialInterval = time.Duration(retryAfterSeconds) * time.Second
 	finalResponse, finalError = backoff.Retry(req.Context(), op, backoff.WithBackOff(backoffMethod))
 
-	// if retries fail to resolve the issue, we need to unset the locally used error type to not return it from the RoundTripper
-	if errors.Is(finalError, errRetryNecessary) {
-		rm.logger.Warn().Msgf("Retry ultimately failed after %d attempts", actualAttempts)
-		finalError = nil
-	}
+	return finalResponse, rm.maskInternalRetryError(finalError, actualAttempts)
+}
 
-	return finalResponse, finalError
+// maskInternalRetryError strips sentinel errors used only inside the retry loop so callers receive the last HTTP response.
+func (rm RetryMiddleware) maskInternalRetryError(err error, actualAttempts int) error {
+	if errors.Is(err, errRetryNecessary) {
+		rm.logger.Warn().Msgf("Retry ultimately failed after %d attempts", actualAttempts)
+		return nil
+	}
+	if errors.Is(err, errRetryDelayMaxExceeded) {
+		rm.logger.Warn().Msg("Suggested retry delay from Retry-After or X-RateLimit-Reset exceeds maximum allowed wait; returning last HTTP response")
+		return nil
+	}
+	return err
 }
 
 func getMaxRetryAttempts(response *http.Response, maxAttempts int) int {
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -133,11 +133,12 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	backoffMethod.InitialInterval = time.Duration(retryAfterSeconds) * time.Second
 	finalResponse, finalError = backoff.Retry(req.Context(), op, backoff.WithBackOff(backoffMethod))
 
-	return finalResponse, rm.maskInternalRetryError(finalError, actualAttempts)
+	finalError = rm.filterRetryError(finalError, actualAttempts)
+	return finalResponse, finalError
 }
 
-// maskInternalRetryError strips sentinel errors used only inside the retry loop so callers receive the last HTTP response.
-func (rm RetryMiddleware) maskInternalRetryError(err error, actualAttempts int) error {
+// filterRetryError strips sentinel errors used only inside the retry loop so callers receive the last HTTP response.
+func (rm RetryMiddleware) filterRetryError(err error, actualAttempts int) error {
 	if errors.Is(err, errRetryNecessary) {
 		rm.logger.Warn().Msgf("Retry ultimately failed after %d attempts", actualAttempts)
 		return nil
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -42,8 +42,7 @@ var statusCodesToRetryLUT = map[int]retryLogic{
 }
 
 var errRetryNecessary = errors.New("retry with backoff")
-var errRetryAfterHeaderError = errors.New("retry-after is too much in the future")
-var errXRateLimitResetHeaderError = errors.New("X-RateLimit-Reset is too far in the future")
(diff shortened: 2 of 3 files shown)


pkg/networking/middleware/retry_middleware.go
```diff
@@ -172,12 +179,23 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
...
-		// if the fix retry delay is too big, we rather fail permanently than blocking too long
-		if fixRetryDelay > maxRetryAfter {
-			return backoff.Permanent(errRetryAfterHeaderError)
+		if fixRetryDelay == 0 {
+			// try to read X-RateLimit-Reset header if available
+			// according to envoy docs: number of seconds until reset of the current time-window
+			if headerXRateLimitResetValue := response.Header.Get("X-RateLimit-Reset"); len(headerXRateLimitResetValue) > 0 {
+				fixRetryDelay = parseRetryDelay(headerXRateLimitResetValue)
+			}
+			if fixRetryDelay > maxRetryAfter {
```
PeterSchafer: Question: @robertolopezlopez did we verify if the existing `maxRetryAfter` is still a sensible max value for `X-RateLimit-Reset`?

robertolopezlopez: we probably need to find out first how long the rate limiting window lasts, I have just asked the platform team

PeterSchafer:
@robertolopezlopez did you create a CLI branch for this change already? Would be great to do, so that we can get these changes tested and merged there.

robertolopezlopez:
> @robertolopezlopez did you create a CLI branch for this change already? Would be great to do, so that we can get these changes tested and merged there.

https://github.com/snyk/cli/pull/6797

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: > The max value WE put in the reset (right now) is 3600 (one hour). But this may change in the future.
> 
> This is due to the buckets that we have on most (if not all) our rate limits:
> • Second
> • Minute
> • Hour
> 
> This is all configured in the API Gateway
> https://github.com/snyk/api-gateway/blob/main/helm/templates/rate-limits/api-rest.yaml

https://snyk.slack.com/archives/C08LCKLSPLZ/p1778596702460069?thread_ts=1778501326.321139&cid=C08LCKLSPLZ

So:

- 1 hour
- This span length may change in the future
- The linked helm template does not give us any significant information, as the real values are not injected yet (there are just placeholders).

In our code: `maxRetryAfter == 10 min`. Comparing that to 1 hour, I think `maxRetryAfter` value makes sense or at least is within the retry window. Does it make sense to you? (you indeed have better insight)

pkg/networking/middleware/retry_middleware.go
```diff
@@ -133,13 +133,20 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	backoffMethod.InitialInterval = time.Duration(retryAfterSeconds) * time.Second
 	finalResponse, finalError = backoff.Retry(req.Context(), op, backoff.WithBackOff(backoffMethod))
 
-	// if retries fail to resolve the issue, we need to unset the locally used error type to not return it from the RoundTripper
-	if errors.Is(finalError, errRetryNecessary) {
+	return finalResponse, rm.maskInternalRetryError(finalError, actualAttempts)
```
danskmt: I suggest separating this like:
```go
maskedErr := rm.maskInternalRetryError(finalError, actualAttempts)
return finalResponse, maskedErr
```

pkg/networking/middleware/retry_middleware.go
```diff
@@ -133,13 +133,20 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		return nil
 	}
-
-	return finalResponse, finalError
+	if errors.Is(err, errRetryDelayMaxExceeded) {
+		rm.logger.Warn().Msg("Suggested retry delay from Retry-After or X-RateLimit-Reset exceeds maximum allowed wait; returning last HTTP response")
+		return nil
+	}
+	return err
 }
```
danskmt: The name of the function is **maskInternalRetryError** but we are just getting rid of them when the conditions satisfy, i.e. returning `nil`. I'm wondering if would be more appropriate to still return an error in these cases, not exactly the same, but something that carries the information

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: refactored as we agreed

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: as agreed, I changed the name of this function to `filterRetryError`

pkg/networking/middleware/retry_middleware.go
```diff
@@ -133,13 +133,21 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
```
danskmt: We can change this to processRetryError or something like that where we still return an error with information so we can surface it.
It seems we have some ErrorCatalog we can use
cc: @PeterSchafer

danskmt: FYI this PR does this: https://github.com/snyk/go-application-framework/pull/610
Let's part of the implementation from my PR to this one, so we keep concerns separated.

danskmt: image

danskmt: PS: we probably want to consider some adequate types for other error codes as well?

pkg/networking/middleware/retry_middleware_test.go
```diff
@@ -112,7 +113,7 @@ func TestNewRetryMiddleware(t *testing.T) {
 		sut := NewRetryMiddleware(config, &logger, failRoundtripper)
 		response, err := sut.RoundTrip(httptest.NewRequest(http.MethodGet, "/", nil))
 
-		assert.NoError(t, err)
+		assert.Nil(t, err)
```
PeterSchafer: nit: using NoError() is actually semantically more helpful and used throughout the codebase

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: @danskmt : after yesterday and today syncs, do you agree to resolve this thread?

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: resolving this thread after clarification

pkg/networking/middleware/retry_middleware.go
```diff
@@ -172,12 +187,23 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
...
+			// if the fix retry delay is too big, we rather fail permanently than blocking too long
+			if fixRetryDelay > maxRetryAfter {
+				return backoff.Permanent(errRetryDelayMaxExceeded)
+			}
 		}
 
-		// if the fix retry delay is too big, we rather fail permanently than blocking too long
-		if fixRetryDelay > maxRetryAfter {
-			return backoff.Permanent(errRetryAfterHeaderError)
+		if fixRetryDelay == 0 {
```
PeterSchafer: Suggestion: I'd like to +1 on the [bot comment](https://github.com/snyk/go-application-framework/pull/604#issuecomment-4486975890) and instead of having one value take precedence over the other evaluate both if existing and use the higher value.

pkg/networking/middleware/response.go
```diff
@@ -133,6 +133,8 @@ func errFromStatusCode(code int) error {
 		return snyk.NewBadRequestError("The request cannot be processed.")
 	case http.StatusInternalServerError:
 		return snyk.NewServerError("Internal server error.")
+	case http.StatusTooManyRequests:
+		return snyk.NewTooManyRequestsError("Too many requests.")
```
danskmt: @PeterSchafer @robertolopezlopez 

Do we want to change this "warn" to "error" here?

Image

Or another option is to do this:
```go
case http.StatusTooManyRequests:
		err := snyk.NewTooManyRequestsError("")
		err.Level = "error"
		return err
```

PeterSchafer: Good catch!

danskmt: @robertolopezlopez also we probably want to have a test here pkg/networking/middleware/response_test.go

https://github.com/snyk/go-application-framework/pull/610/changes

Image

Maybe we should add in this PR? wdyt?

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: @PeterSchafer yesterday we didn't sync about this, what do you think of my modification?

PeterSchafer: @robertolopezlopez I don't think the current change addresses the concern raised.

Please assume the following case:
* Retry After-Header is set to 2 seconds
* X-Ratelimit-Reset Header is set to 10 seconds

The current implementation has a strict precedence that Retry-After has priority and anything else gets ignored, but what if X-Ratelimit-Reset has the greater value. 

And after thinking a bit more about it, I think using the greater of the two header values is the more reliable approach

PeterSchafer (approved):

PR #606, opened 2026-05-11
fix: apply invocation config to workflow Data after callback

danskmt:
### Description

Workflows can return `Data` created with `workflow.NewData` without `WithConfiguration`, so `IN_MEMORY_THRESHOLD_BYTES` and `TEMP_DIR_PATH` were not applied to those payloads.

After each workflow callback returns, `EngineImpl.Invoke` calls `applyConfiguration` on each `*DataImpl` in the output slice so in-memory payloads can spill to the configured temp directory when above threshold. `applyConfiguration` is package-private on `DataImpl` (same package as the engine) so it is not exported from the workflow package.

Adds tests in `dataimpl_test.go` for spill and no-op cases, and in `engine_test.go` for behavior after `Invoke`.

### Checklist

- [x] Linted (`make lint`)
  
PR in CLI https://github.com/snyk/cli/pull/6792

### Where reviewers should start

- `pkg/workflow/dataimpl.go` — `applyConfiguration`
- `pkg/workflow/engineimpl.go` — post-process loop after `callback`

### Risk assessment

**Low.** Default production behavior is unchanged unless spill-related settings are enabled; when they are, returned `Data` placement aligns with engine configuration instead of leaving eligible payloads in memory.

___

### **PR Type**
Bug fix, Enhancement, Tests

___

### **Description**
- Apply engine configuration to workflow output Data.

- Enable payload spilling based on engine settings.

- Introduce `DataImpl.applyConfiguration` method logic.

- Add comprehensive tests for configuration application.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  engine_invoke["EngineImpl.Invoke"] --> workflow_callback["Workflow Callback"];
  workflow_callback --> output_data["Output Data"];
  output_data --> post_processing_loop["Apply Config Loop"];
  post_processing_loop --> dataimpl_apply_config["DataImpl.applyConfiguration"];
  dataimpl_apply_config --> spill_to_disk["Payload Spills to Disk (if needed)"];
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>dataimpl.go</strong><dd><code>Implement DataImpl configuration application logic</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/workflow/dataimpl.go

<ul><li>Modified <code>WithConfiguration</code> to conditionally apply threshold and temp <br>dir values.<br> <li> Introduced <code>applyConfiguration</code> method to relocate in-memory payloads to <br>disk based on configuration.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/606/changes#diff-1752c5ed22bf801a49fb09a0c131b3d5315cca0612ae39d0b43a503f37b17bfb">+32/-2</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>dataimpl_test.go</strong><dd><code>Add tests for DataImpl applyConfiguration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/workflow/dataimpl_test.go

<ul><li>Added several new test cases for <code>applyConfiguration</code>.<br> <li> Tests cover spill behavior, no-op scenarios, default values, and <br>missing temp directory paths.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/606/changes#diff-a282c7d1ff03bff6144f2fe2b46aeb71a92245dad129535e74546eedb7ee591b">+135/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>engine_test.go</strong><dd><code>Test engine Invoke output configuration application</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/workflow/engine_test.go

<ul><li>Added a new test <code>Test_Invoke_AppliesConfigurationToOutput</code>.<br> <li> Verifies that engine's invocation applies configuration to output <br>data, causing spills.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/606/changes#diff-95bcd88858f99e1bfecd015f2c3490e3072dbcb550ac8f7cb385b7462ae929cc">+39/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>engineimpl.go</strong><dd><code>Apply engine configuration to Invoke output</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/workflow/engineimpl.go

<ul><li>Modified <code>Invoke</code> to iterate over workflow output <code>Data</code>.<br> <li> Calls <code>di.applyConfiguration(options.config)</code> on each <code>DataImpl</code> to apply <br>engine settings.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/606/changes#diff-672848374dc64d0b5059745d90ef50be714f0f3d9791ef4f6db7174eff68c460">+9/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/pkg/workflow/dataimpl.go
+++ b/pkg/workflow/dataimpl.go
@@ -66,10 +66,18 @@ func withPayload(payload *interface{}) Option {
 	}
 }
 
+// WithConfiguration copies IN_MEMORY_THRESHOLD_BYTES and TEMP_DIR_PATH from
+// config when each key resolves to a value (including AddDefaultValue). If
+// TEMP_DIR_PATH is absent, d.tempDirPath is left unchanged (typically ""), and
+// os.CreateTemp uses the process default temp directory.
 func WithConfiguration(config configuration.Configuration) Option {
 	return func(d *DataImpl) {
-		d.inMemoryThreshold = config.GetInt(configuration.IN_MEMORY_THRESHOLD_BYTES)
-		d.tempDirPath = config.GetString(configuration.TEMP_DIR_PATH)
+		if v := config.Get(configuration.IN_MEMORY_THRESHOLD_BYTES); v != nil {
+			d.inMemoryThreshold = config.GetInt(configuration.IN_MEMORY_THRESHOLD_BYTES)
+		}
+		if v := config.Get(configuration.TEMP_DIR_PATH); v != nil {
+			d.tempDirPath = config.GetString(configuration.TEMP_DIR_PATH)
+		}
 	}
 }
 
@@ -254,6 +262,28 @@ func (d *DataImpl) GetErrorList() []snyk_errors.Error {
 
 func (d *DataImpl) AddError(err snyk_errors.Error) {
 	d.errors = append(d.errors, err)
+}
+
+// applyConfiguration re-evaluates the payload location using the given
+// configuration. Field updates use the same rules as WithConfiguration.
+// If the payload is currently in memory and exceeds the
+// configured threshold, it is written to disk under the configured temp
+// directory. This allows the engine to apply its configuration to Data
+// objects that were created without WithConfiguration.
+func (d *DataImpl) applyConfiguration(config configuration.Configuration) {
+	if config.Get(configuration.IN_MEMORY_THRESHOLD_BYTES) == nil {
+		return
+	}
+
+	WithConfiguration(config)(d)
+
+	if d.payloadLocation.Type == InMemory && d.payload != nil {
+		d.payloadLocation = setPayloadLocation(d.identifier, d.inMemoryThreshold, d.tempDirPath, d.payload, d.logger)
+		if d.payloadLocation.Type == OnDisk {
+			d.logger.Debug().Msg("payload relocated to disk after applyConfiguration")
+			d.payload = nil
+		}
+	}
 }
 
 func setPayloadLocation(id Identifier, inMemoryThreshold int, tempDirPath string, payload interface{}, logger *zerolog.Logger) Location {
--- a/pkg/workflow/engineimpl.go
+++ b/pkg/workflow/engineimpl.go
@@ -338,6 +338,15 @@ func (e *EngineImpl) Invoke(
 			localLogger.Printf("Workflow Start")
 			output, err = callback(invocationCtx, options.input)
 			localLogger.Printf("Workflow End")
+
+			// Apply the engine's configuration to output data so that
+			// IN_MEMORY_THRESHOLD_BYTES and TEMP_DIR_PATH are respected
+			// even when workflows create Data without WithConfiguration.
+			for _, d := range output {
+				if di, ok := d.(*DataImpl); ok {
+					di.applyConfiguration(options.config)
+				}
+			}
 		}
 	} else {
 		err = fmt.Errorf("workflow '%v' not found", id)
--- a/pkg/workflow/dataimpl_test.go
+++ b/pkg/workflow/dataimpl_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/snyk/error-catalog-golang-public/snyk_errors"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func Test_NewDataFromInput(t *testing.T) {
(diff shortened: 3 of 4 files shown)


pkg/workflow/dataimpl.go
```diff
@@ -256,6 +256,24 @@ func (d *DataImpl) AddError(err snyk_errors.Error) {
 	d.errors = append(d.errors, err)
 }
 
+// applyConfiguration re-evaluates the payload location using the given
+// configuration. If the payload is currently in memory and exceeds the
+// configured threshold, it is written to disk under the configured temp
+// directory. This allows the engine to apply its configuration to Data
+// objects that were created without WithConfiguration.
+func (d *DataImpl) applyConfiguration(config configuration.Configuration) {
+	d.inMemoryThreshold = config.GetInt(configuration.IN_MEMORY_THRESHOLD_BYTES)
```
CatalinSnyk: suggestion: I think the PR bot feedback is correct in this case: we should override this only if the configuration value is set, maybe an early return like:
```go
if !config.IsSet(configuration.IN_MEMORY_THRESHOLD_BYTES) {
    return
}
```
would do the trick?

CatalinSnyk: @PeterSchafer could you also take a look if you have the time?

danskmt: Using IsSet wouldn't work here: **IN_MEMORY_THRESHOLD_BYTES** is registered via [AddDefaultValue](https://github.com/snyk/go-application-framework/blob/e94ad3cfdf43bad5898921d8dbe11d0729466025/pkg/app/app.go#L278), and `IsSet` returns false for those. That would skip applyConfiguration entirely and disable disk-spilling.

Instead we can use config.Get(key) == nil, which respects both Set and AddDefaultValue. It only skips when the key is truly absent (the case where GetInt would return 0).

pkg/workflow/dataimpl.go
```diff
@@ -66,10 +66,18 @@ func withPayload(payload *interface{}) Option {
 	}
 }
 
+// WithConfiguration copies IN_MEMORY_THRESHOLD_BYTES and TEMP_DIR_PATH from
+// config when each key resolves to a value (including AddDefaultValue). If
+// TEMP_DIR_PATH is absent, d.tempDirPath is left unchanged (typically ""), and
+// os.CreateTemp uses the process default temp directory.
 func WithConfiguration(config configuration.Configuration) Option {
```
robertolopezlopez, line 73: in an ideal world (at least the one in my head) there should be a receiver `func (d *DataImpl) WithConfiguration(...)`. As this may be problematic to implement now (too many usages off the task scope), I suggest the following:

```
func WithConfiguration(config configuration.Configuration) Option {
   return func(d *DataImpl) { d.configureFrom(config) }
}

func (d *DataImpl) configureFrom(config configuration.Configuration) {
   // copy in memory threshold and temp dir path if present in config
}

func (d *DataImpl) applyConfiguration(config configuration.Configuration) {
   ... your nil check
   d.configureFrom(config)
   ... rest of logic from your method
}
```

what do you think - this doesn't change the existing signatures which is probably used from other places, while keeping the state in the receiver

pkg/workflow/dataimpl.go
```diff
@@ -256,6 +264,28 @@ func (d *DataImpl) AddError(err snyk_errors.Error) {
...
 }
 
+// applyConfiguration re-evaluates the payload location using the given
+// configuration. Field updates use the same rules as WithConfiguration.
+// If the payload is currently in memory and exceeds the
+// configured threshold, it is written to disk under the configured temp
+// directory. This allows the engine to apply its configuration to Data
+// objects that were created without WithConfiguration.
+func (d *DataImpl) applyConfiguration(config configuration.Configuration) {
+	if config.Get(configuration.IN_MEMORY_THRESHOLD_BYTES) == nil {
```
basti-snyk, line 274: **Should-Fix (correctness): new nil-config panic path.** `Invoke` now unconditionally calls `applyConfiguration(options.config)` for every `*DataImpl` output, and this first line dereferences the config. When a caller passes a nil config — `Invoke(id, WithConfig(nil))` or the deprecated `InvokeWithConfig(id, nil)` — `options.config` is nil, so this panics (nil interface method call). Before this PR that path was harmless: the engine only stored the config (`newInvocationContext`, `SetConfiguration`) and never dereferenced it. Root-cause fix is one line at the top of this method: `if config == nil { return }`. Please also add a test: `Invoke` with `WithConfig(nil)` and a workflow returning `NewData(...)` must not panic. — AI review

pkg/workflow/engineimpl.go
```diff
@@ -338,6 +338,15 @@ func (e *EngineImpl) Invoke(
 			localLogger.Printf("Workflow Start")
 			output, err = callback(invocationCtx, options.input)
 			localLogger.Printf("Workflow End")
+
+			// Apply the engine's configuration to output data so that
+			// IN_MEMORY_THRESHOLD_BYTES and TEMP_DIR_PATH are respected
+			// even when workflows create Data without WithConfiguration.
+			for _, d := range output {
+				if di, ok := d.(*DataImpl); ok {
+					di.applyConfiguration(options.config)
```
basti-snyk, line 347: **Should-Fix (contract): engine silently overrides a workflow's explicit `WithConfiguration`.** This loop applies the engine config to *every* returned `*DataImpl`, and `applyConfiguration` re-runs `WithConfiguration`, overwriting `inMemoryThreshold`/`tempDirPath`. A workflow that deliberately built its Data with `NewData(..., WithConfiguration(customCfg))` (e.g. a higher threshold to keep a payload in memory on purpose) has that intent silently replaced by the engine defaults and force-spilled. `applyConfiguration` can't distinguish "never configured" (the case this PR targets) from "deliberately configured". Either apply only when the Data was never configured, or document on `Invoke` that engine config always wins last. This override path is also untested (new tests only cover Data created without `WithConfiguration`). — AI review

pkg/workflow/engineimpl.go
```diff
@@ -338,6 +338,15 @@ func (e *EngineImpl) Invoke(
 			localLogger.Printf("Workflow Start")
 			output, err = callback(invocationCtx, options.input)
 			localLogger.Printf("Workflow End")
+
+			// Apply the engine's configuration to output data so that
+			// IN_MEMORY_THRESHOLD_BYTES and TEMP_DIR_PATH are respected
+			// even when workflows create Data without WithConfiguration.
+			for _, d := range output {
```
basti-snyk, line 345: **Suggestion: relocation runs even when the callback returned an error.** This loop is outside any `if err == nil` guard, so when a workflow returns partial `output` alongside a non-nil `err`, that soon-to-be-discarded output is still spilled to disk, leaving orphan temp files for data nobody consumes. Guarding the loop with `if err == nil` avoids the wasted I/O. — AI review

pkg/workflow/dataimpl.go
```diff
@@ -66,10 +66,18 @@ func withPayload(payload *interface{}) Option {
...
 
+// WithConfiguration copies IN_MEMORY_THRESHOLD_BYTES and TEMP_DIR_PATH from
+// config when each key resolves to a value (including AddDefaultValue). If
+// TEMP_DIR_PATH is absent, d.tempDirPath is left unchanged (typically ""), and
+// os.CreateTemp uses the process default temp directory.
 func WithConfiguration(config configuration.Configuration) Option {
 	return func(d *DataImpl) {
-		d.inMemoryThreshold = config.GetInt(configuration.IN_MEMORY_THRESHOLD_BYTES)
-		d.tempDirPath = config.GetString(configuration.TEMP_DIR_PATH)
+		if v := config.Get(configuration.IN_MEMORY_THRESHOLD_BYTES); v != nil {
```
basti-snyk, line 75: **Suggestion: exported-API semantic change worth a direct-caller test.** `WithConfiguration` previously set `inMemoryThreshold`/`tempDirPath` unconditionally (unset key → `0`/`""`); it now skips a key when `config.Get(...) == nil`. This is a genuine fix (an unset threshold no longer means "spill everything"), but `WithConfiguration` is public and no test asserts the new behavior for direct external callers when one key is set and the other is unset. Any caller that relied on "unset threshold ⇒ 0 ⇒ spill" changes silently. Add a focused unit test for the mixed set/unset case. — AI review

PR #609, opened 2026-05-18
chore: adds org lookup type to instrumentation data

j-luong:
### Description

This PR adds some extra instrumentation data to how we perform org lookups.

### Checklist

- [ ] Linted (`make lint`)

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -22,6 +22,7 @@ import (
 	"github.com/snyk/go-application-framework/internal/constants"
 	"github.com/snyk/go-application-framework/internal/presenters"
 	"github.com/snyk/go-application-framework/internal/utils"
+	"github.com/snyk/go-application-framework/pkg/analytics"
 	"github.com/snyk/go-application-framework/pkg/auth"
 	"github.com/snyk/go-application-framework/pkg/configuration"
 	localworkflows "github.com/snyk/go-application-framework/pkg/local_workflows"
@@ -30,6 +31,47 @@ import (
 	"github.com/snyk/go-application-framework/pkg/utils/conversion"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
+
+// OrgLookup records how the ORGANIZATION value was resolved.
+type OrgLookup string
+
+const (
+	// orgLookupSourceKey is the instrumentation extension key for the OrgLookup value.
+	orgLookupSourceKey = "gaf.app.defaultfunc.organization.lookup"
+
+	// Values for orgLookupSourceKey.
+	orgLookupProvidedID         OrgLookup = "provided_id"
+	orgLookupProvidedSlug       OrgLookup = "provided_slug"
+	orgLookupProvidedSlugFailed OrgLookup = "provided_slug_failed"
+	orgLookupDefault            OrgLookup = "default"
+	orgLookupDefaultFailed      OrgLookup = "default_failed"
+)
+
+type instrumentationRecorder struct {
+	collector analytics.InstrumentationCollector
+	logger    *zerolog.Logger
+}
+
+// newInstrumentationRecorder wraps the analytics.InstrumentationCollector to prevent nil-check panics
+// (e.g. during engine Init, before e.analytics is assigned).
+func newInstrumentationRecorder(engine workflow.Engine, logger *zerolog.Logger) instrumentationRecorder {
+	r := instrumentationRecorder{
+		logger: logger,
+	}
+	if a := engine.GetAnalytics(); a != nil {
+		r.collector = a.GetInstrumentation()
+	}
+	return r
+}
+
+//nolint:unparam // the method is general-purpose and can receive any key, value
+func (r instrumentationRecorder) addExtension(key string, value any) {
+	if r.collector == nil {
+		r.logger.Debug().Msgf("Instrumentation collector is nil, skipping extension for key: %s, value: %v", key, value)
+		return
+	}
+	r.collector.AddExtension(key, value)
+}
 
 func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Configuration, logger *zerolog.Logger, apiClientFactory func(url string, client *http.Client) api.ApiClient) configuration.DefaultValueFunction {
 	err := config.AddKeyDependency(configuration.ORGANIZATION_SLUG, configuration.ORGANIZATION)
@@ -49,7 +91,7 @@ func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Co
 		}
 		slugName, err := apiClient.GetSlugFromOrgId(context.Background(), orgId)
 		if err != nil {
-			logger.Print("Failed to determine default value for \"ORGANIZATION_SLUG\":", err)
+			logger.Print("Failed to determine slugname for \"ORGANIZATION_SLUG\":", err)
 		}
 		return slugName, nil
 	}
@@ -65,6 +107,8 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 	callback := func(_ configuration.Configuration, existingValue interface{}) (interface{}, error) {
 		// TODO - This function uses the outer (global) config and network access, so will not respect values set in the closures' (potentially cloned) configs.
 		client := engine.GetNetworkAccess().GetHttpClient()
+		instrumentation := newInstrumentationRecorder(engine, logger)
+		instrumentation.addExtension(orgLookupSourceKey, orgLookupDefault)
 		url := config.GetString(configuration.API_URL)
 		apiClient := apiClientFactory(url, client)
 		existingString, ok := existingValue.(string)
@@ -75,20 +119,25 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 			if isSlugName {
 				orgId, err = apiClient.GetOrgIdFromSlug(context.Background(), existingString)
 				if err != nil {
-					logger.Print("Failed to determine default value for \"ORGANIZATION\":", err)
+					instrumentation.addExtension(orgLookupSourceKey, orgLookupProvidedSlugFailed)
+					logger.Print("Failed to lookup \"ORGANIZATION\" id from slugname :", err)
 				} else {
+					instrumentation.addExtension(orgLookupSourceKey, orgLookupProvidedSlug)
 					return orgId, nil
 				}
 			} else {
+				// TODO [CLI-436] - we always return the org ID here, even if it is not valid, which is not what we do for slugs
+				// it makes this datapoint a bit ambiguous as a result
+				instrumentation.addExtension(orgLookupSourceKey, orgLookupProvidedID)
 				return orgId, nil
 			}
 		}
 
 		orgId, err := apiClient.GetDefaultOrgId(context.Background())
 		if err != nil {
+			instrumentation.addExtension(orgLookupSourceKey, orgLookupDefaultFailed)
 			logger.Print("Failed to determine default value for \"ORGANIZATION\":", err)
 		}
-
 		return orgId, err
 	}
 	return callback
--- a/pkg/workflow/engineimpl.go
+++ b/pkg/workflow/engineimpl.go
@@ -184,7 +184,6 @@ func (e *EngineImpl) initAnalytics() analytics.Analytics {
 	a := analytics.New()
 	a.SetIntegration(e.config.GetString(configuration.INTEGRATION_NAME), e.config.GetString(configuration.INTEGRATION_VERSION))
 	a.SetApiUrl(e.config.GetString(configuration.API_URL))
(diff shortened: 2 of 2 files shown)


pkg/app/app.go
```diff
@@ -65,6 +66,16 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
...
+		// Analytics can be nil here during engine Init,
+		var instrumentation analytics.InstrumentationCollector
+		if a := engine.GetAnalytics(); a != nil {
+			instrumentation = a.GetInstrumentation()
+		}
+		addLookupType := func(value string) {
+			if instrumentation != nil {
+				instrumentation.AddExtension("organization-lookup-type", value)
+			}
+		}
```
danskmt: I'd suggest refactoring to something like this:
```go
// outside
var instrumentation *analytics.InstrumentationCollector = nil

func getInstrumentation(engine workflow.Engine) *analytics.InstrumentationCollector {
  if instrumentation == nil {
     if a := engine.GetAnalytics(); a != nil {
	    instrumentation = a.GetInstrumentation()
     }
  }
  
  return instrumentation
}

func addLookupType(orgLookupType string) bool {
   i := getInstrumentation(engine)
  if i != nil {
		i.AddExtension("organization-lookup-type", orgLookupType)
		return true
  }
  return false
}
```

danskmt: Actually my suggestion may not be the best though, so feel free to take something from it if that makes sense.
The idea is that we can have a better way to use it outside of the function they are currently defined and reducing the size of the function at the same time.

danskmt: we also can make use if the lookup type was applied or not, maybe logging etc.

pkg/app/app.go
```diff
@@ -75,11 +86,14 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 			if isSlugName {
 				orgId, err = apiClient.GetOrgIdFromSlug(context.Background(), existingString)
 				if err != nil {
-					logger.Print("Failed to determine default value for \"ORGANIZATION\":", err)
+					addLookupType("invalid")
+					logger.Print("Failed to lookup slugname for \"ORGANIZATION\":", err)
```
PeterSchafer: Suggestion: I think this log message is misleading.
```suggestion
					logger.Print("Failed to lookup \"ORGANIZATION\" id from slugname :", err)
```

pkg/app/app.go
```diff
@@ -75,11 +86,14 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
 			if isSlugName {
 				orgId, err = apiClient.GetOrgIdFromSlug(context.Background(), existingString)
 				if err != nil {
-					logger.Print("Failed to determine default value for \"ORGANIZATION\":", err)
+					addLookupType("invalid")
```
PeterSchafer: Suggestion: maybe be a bit more explicit and self explaining in the data. Consider a case that someone will just have the instrumentation data, will it be self explaining?

PeterSchafer: thinking of the more complete picture (name: value)
* organization-id-origin: defaultid-failed-slugname-lookup
* organization-id-origin: defaultid-not-user-supplied
* organization-id-origin: slugname-lookup
* organization-id-origin: user-supplied

I don't like my proposal but maybe still food for thought?

pkg/app/app.go
robertolopezlopez: I agree with Daniel, but also I would like to add my extra 2 cents to back his proposal.

- In general yes, closures affect readability.
- Also, **not in this case**, in certain scenarios multiple closure invocations may lead to memory allocation degraded performance.
- And most importantly: **a closure modifying variables in the parent scope can lead very easily to unexpected behavior and bugs**. I would recommend running away from these "false closures".

Thus I vote for Daniel's proposal.

robertolopezlopez (changes requested):

pkg/app/app.go
```diff
@@ -49,7 +75,8 @@ func defaultFuncOrganizationSlug(engine workflow.Engine, config configuration.Co
 		}
 		slugName, err := apiClient.GetSlugFromOrgId(context.Background(), orgId)
 		if err != nil {
-			logger.Print("Failed to determine default value for \"ORGANIZATION_SLUG\":", err)
+			recorder.addExtension(orgLookupKey, "invalid")
```
j-luong: context: a command like  `snyk test --org=<INVALID_UUID>` will result in the `orgLookupkey` value being `orgid` instead of `invalid`

pkg/workflow/engineimpl.go
```diff
@@ -184,7 +184,6 @@ func (e *EngineImpl) initAnalytics() analytics.Analytics {
 	a := analytics.New()
 	a.SetIntegration(e.config.GetString(configuration.INTEGRATION_NAME), e.config.GetString(configuration.INTEGRATION_VERSION))
 	a.SetApiUrl(e.config.GetString(configuration.API_URL))
-	a.SetOrg(e.config.GetString(configuration.ORGANIZATION))
```
j-luong, line 187: context: calling this before we finish initialising our engine means our analytics is not initialised and will lead to it being `nil` when trying to use it in our `config.DefaultValueFunction`.

The consequence is that we have to set this where we use GAF (e.g. CLI) instead.

Possibly a controversial change but, without this, we won't get analytics implemented in the rest of the PR. Additionally, the argument could be made that none of these `"Sets"` should be a GAF concern.

pkg/app/app.go
```diff
@@ -31,13 +32,47 @@ import (
...
+// (e.g. during engine Init, before e.analytics is assigned).
+func newInstrumentationRecorder(engine workflow.Engine) instrumentationRecorder {
+	r := instrumentationRecorder{}
+	if a := engine.GetAnalytics(); a != nil {
+		r.ic = a.GetInstrumentation()
+	}
+	return r
+}
+
+//nolint:unparam // the method is general-purpose and can receive any key, value
```
robertolopezlopez, line 67: thanks @j-luong for this change. Just to clarify, why do you need this lint directive here?

j-luong: The linter fails as I'm using the same value for `key` wherever this fn is called - it thinks we can just remove the `key` param as a result

robertolopezlopez (approved):
LGTM

pkg/app/app.go
j-luong: made an opinionated change, let me know if it works

pkg/app/app.go
```diff
@@ -31,6 +32,39 @@ import (
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
+const (
+	// orgLookupSourceKey records how the ORGANIZATION value was resolved.
+	orgLookupSourceKey = "gaf.app.defaultfunc.organization.lookup"
```
j-luong, line 40: context: key tries to define the kind of data we're assessing in OTel standard dot `.` notation.
The idea is to break it down as follows: `<project>.<package|scope>.<method>.<property>.<detail>`.

danskmt: @j-luong I'm wondering if "gaf" is the correct name to put here because the project is actually `go-application-framework`. Or is this key already defined somewhere else?

pkg/app/app.go
```diff
@@ -75,20 +111,25 @@ func defaultFuncOrganization(engine workflow.Engine, config configuration.Config
...
+					logger.Print("Failed to lookup \"ORGANIZATION\" id from slugname :", err)
 				} else {
+					instrumentation.addExtension(orgLookupSourceKey, orgLookupProvidedSlug)
 					return orgId, nil
 				}
 			} else {
+				// TODO [CLI-1372] - we always return the org ID here, even if it is not valid, which is not what we do for slugs
+				// it makes this datapoint a bit ambiguous as a result
+				instrumentation.addExtension(orgLookupSourceKey, orgLookupProvidedID)
 				return orgId, nil
```
danskmt: I see this is a TODO for the current branch name of this PR, will you address this TODO here or in a follow-up task?

pkg/workflow/engineimpl.go
danskmt: Since GAF is a framework that is used by other repos such as the CLI, do you know what other repos will possibly be affected by this change?

Also, is this change in the CLI that replaces this? I think it is not straightforward to understand that it is setting org by just reading that code
Image

pkg/app/app.go
```diff
@@ -31,6 +32,39 @@ import (
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
+const (
+	// orgLookupSourceKey records how the ORGANIZATION value was resolved.
+	orgLookupSourceKey = "gaf.app.defaultfunc.organization.lookup"
+
+	// Values for orgIDLookupSourceKey.
+	orgLookupProvidedID         = "provided_id"
```
robertolopezlopez: nit: why not creating a data type for this? i.e. the following:

```go
type OrgLookup string

const (
    orgLookupSourceKey = "gaf.app.defaultfunc.organization.lookup"
    orgLookupProvidedID       OrgLookup  = "provided_id"
	orgLookupProvidedSlug     OrgLookup  = "provided_slug"
	orgLookupProvidedSlugFailed OrgLookup = "provided_slug_failed"
	orgLookupDefault           OrgLookup = "default"
	orgLookupDefaultFailed    OrgLookup  = "default_failed"
)
```

and reuse the type. Cleaner IMO

pkg/app/app.go
```diff
@@ -31,6 +32,39 @@ import (
...
+	orgLookupProvidedSlug       = "provided_slug"
+	orgLookupProvidedSlugFailed = "provided_slug_failed"
+	orgLookupDefault            = "default"
+	orgLookupDefaultFailed      = "default_failed"
+)
+
+type instrumentationRecorder struct {
+	ic analytics.InstrumentationCollector
```
robertolopezlopez:
```suggestion
	collector analytics.InstrumentationCollector
```

pkg/app/app.go
```diff
@@ -31,6 +32,39 @@ import (
...
+func newInstrumentationRecorder(engine workflow.Engine) instrumentationRecorder {
+	r := instrumentationRecorder{}
+	if a := engine.GetAnalytics(); a != nil {
+		r.ic = a.GetInstrumentation()
+	}
+	return r
+}
+
+//nolint:unparam // the method is general-purpose and can receive any key, value
+func (r instrumentationRecorder) addExtension(key string, value any) {
```
robertolopezlopez, line 68:
```suggestion
func (r instrumentationRecorder) addExtension(key string, value OrgLookup) {
```

why `any`?? maybe `interface{}` as it is defined in the `InstrumentationCollector` interface? Actually as you are just using this here, I would use either `string` or `OrgLookup` if you like my other suggestion

pkg/app/app.go
```diff
@@ -31,6 +32,39 @@ import (
...
+	if r.ic != nil {
+		r.ic.AddExtension(key, value)
+	}
```
robertolopezlopez: and what if ic == nil? would that be normal or do we need to log it?

pkg/app/app.go
j-luong: it's just a shorthand to make it less verbose.

pkg/app/app.go
j-luong: > why any?? maybe interface{}

`any === interface{}` and is quicker to type for lazy people.

I kept the key/value parameters as it makes the method more general and not tied to org lookup

pkg/app/app.go
j-luong: The way we initialise these parameters means that it is very possible that the instrumentation collector can be nil when this is called, without this check, the application will panic.

If it is nil, then it will not add the instrumentation data. It does this silently at the moment as it is not on the critical path if this data is not added and adding logs might make things very verbose.

If we find that we need to log this for whatever reason, then we can add logging later.

pkg/app/app.go
j-luong: This TODO just describes the existing implementation, that has a follow on impact on how we collect the org lookup data.

The fix is not part of this PR as we don't want to make business logic changes here. The TODO references the ticket to fix this

pkg/app/app.go
robertolopezlopez: mmmmh I would argue that they are the same: `any` is mostly used for generics (which you are not using here) while `interface{}` is used for dynamic typing

pkg/app/app.go
robertolopezlopez: Sure, I am not wondering if it may happen or not. I am just asking if we should log it here, early

pkg/workflow/engineimpl.go
j-luong: No, it should only be for the CLI with the change you highlighted.

I added the context comment to help make it a bit more digestible.

pkg/app/app.go
j-luong: Ok, I'll add the logging

danskmt (approved):

PR #612, opened 2026-05-22
feat: show rate-limit UI feedback during 429 retry backoff

danskmt:
### Description

When the CLI is rate-limited (HTTP 429), retry backoff previously happened silently, making the CLI look hung. This PR adds user-visible feedback during 429 retry waits and records retry rate-limit events in v2 instrumentation.

**During 429 backoff (between attempts):**
- `RetryMiddleware` wraps retry failures in `RetryAttemptError` and invokes `backoff.WithNotify`.
- `notifyRetry` maps 429 attempts to a warn-level catalog error via `retryAttemptNotification` (detail: `Automatically retrying in N seconds... (attempt X/Y).`), with the original `RetryAttemptError` attached as `Cause`, then calls the registered `ErrorHandlerFunc`.
- Only 429 attempts produce this notification; other retried status codes (5xx, etc.) do not.

**Consumer-side (CLI):**
- The CLI's `AddErrorHandler` callback detects retry notifications via `errors.As(err, &RetryAttemptError)` on the error's cause chain, then calls `OutputError` and `AddError`, and returns `nil` so the event is not collected as a command error.
- GAF owns mapping and notification shape; the CLI only handles display and analytics.

**When 429 retries are exhausted:**
- `RetryMiddleware` still returns the last 429 response with `nil` error (unchanged).
- Downstream `ResponseMiddleware` maps 429 to SNYK-0001 (`TooManyRequestsError`, level `error`) — covered by a test in `response_test.go`; `response.go` is unchanged.

**Plumbing:**
- `retry_middleware.go`: `RetryAttemptError`, unexported `retryAttemptNotification`; optional variadic `errorHandler` on `NewRetryMiddleware`; `notifyRetry` calls the error handler with the catalog error
- `networking.go`: passes `errorHandler` into `NewRetryMiddleware` (no new network-layer interfaces)
- Tests: `TestRetryAttemptNotification`, retry `errorHandler` / exhaustion tests, 429 mapping in `response_test.go`

### Checklist

- [x] Linted (`make lint`)
  2. Run `go get github.com/snyk/go-application-framework@<commit>` in the `cliv2` directory.
  - Once this PR is merged, repeat these steps pointing at the latest GAF commit on main and update your CLI PR.

CLI PR: https://github.com/snyk/cli/pull/6836

### DIFF
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -162,7 +162,7 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	finalResponse, finalError = backoff.Retry(reqCtx, op,
 		backoff.WithBackOff(backoffMethod),
 		backoff.WithNotify(func(err error, duration time.Duration) {
-			rm.notifyRetry(err, duration, reqCtx)
+			rm.notifyRetry(reqCtx, err, duration)
 		}),
 	)
 
@@ -170,14 +170,14 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	return finalResponse, finalError
 }
 
-func (rm RetryMiddleware) notifyRetry(err error, duration time.Duration, ctx context.Context) {
+func (rm RetryMiddleware) notifyRetry(ctx context.Context, err error, duration time.Duration) {
 	if rm.errorHandler == nil {
 		return
 	}
 	var retryErr *RetryAttemptError
 	if errors.As(err, &retryErr) {
 		retryErr.Duration = duration
-		if catalogErr, ok := RetryAttemptNotification(retryErr); ok {
+		if catalogErr, ok := retryAttemptNotification(retryErr); ok {
 			if handlerErr := rm.errorHandler(catalogErr, ctx); handlerErr != nil {
 				rm.logger.Debug().Err(handlerErr).Msg("error handler failed during retry notification")
 			}
@@ -263,7 +263,7 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
 	return nil
 }
 
-func RetryAttemptNotification(err error) (error, bool) {
+func retryAttemptNotification(err error) (error, bool) {
 	var attempt *RetryAttemptError
 	if !errors.As(err, &attempt) || attempt.StatusCode != http.StatusTooManyRequests {
 		return nil, false
@@ -274,12 +274,11 @@ func RetryAttemptNotification(err error) (error, bool) {
 		totalSecs = 1
 	}
 
-	return snyk_errors.Error{
-		Title:  "Rate limited",
-		Detail: fmt.Sprintf("Waiting up to %ds before retry (attempt %d/%d).", totalSecs, attempt.Attempt, attempt.MaxAttempts),
-		Level:  "warn",
-		Cause:  attempt,
-	}, true
+	notifMsg := fmt.Sprintf("Automatically retrying in %ds seconds... (attempt %d/%d).", totalSecs, attempt.Attempt, attempt.MaxAttempts)
+	notif := snyk.NewTooManyRequestsError(notifMsg, snyk_errors.WithCause(attempt))
+	notif.Description = ""
+
+	return notif, true
 }
 
 func parseRetryDelay(headerRetryAfterValue string) time.Duration {
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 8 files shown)


pkg/networking/middleware/retry_middleware_test.go
```diff
@@ -550,3 +553,201 @@ func Test_parseRetryDelay(t *testing.T) {
...
+	}
+
+	t.Run("returns last 429 response and nil error", func(t *testing.T) {
+		sut, attemptCount := newSUT(t, nil)
+		resp, err := sut.RoundTrip(httptest.NewRequest(http.MethodGet, "/", nil))
+
+		assert.NoError(t, err)
+		assert.NotNil(t, resp)
+		assert.Equal(t, http.StatusTooManyRequests, resp.StatusCode)
+		assert.Equal(t, 3, *attemptCount, "429 override enforces at least 3 attempts even when maxAttempts=1")
```
danskmt: This assertion for this value "3" is due to this definition in retry_midleware.go:
```go
var attemptThreeNetworkRequests int = 3
```

Image

robertolopezlopez (approved):

pkg/networking/middleware/retry_middleware.go
```diff
@@ -44,21 +44,45 @@ var statusCodesToRetryLUT = map[int]retryLogic{
...
 	config           configuration.Configuration
 	logger           *zerolog.Logger
+	onRetryNotify    RetryNotifyFunc
 }
 
-func NewRetryMiddleware(config configuration.Configuration, logger *zerolog.Logger, roundTripper http.RoundTripper) *RetryMiddleware {
+func NewRetryMiddleware(config configuration.Configuration, logger *zerolog.Logger, roundTripper http.RoundTripper, onRetryNotify ...RetryNotifyFunc) http.RoundTripper {
+	var notify RetryNotifyFunc
+	if len(onRetryNotify) > 0 {
+		notify = onRetryNotify[0]
```
robertolopezlopez: being this a variadic function, why do you just use the first element of `onRetryModify`?

pkg/networking/middleware/retry_middleware.go
```diff
@@ -44,21 +44,45 @@ var statusCodesToRetryLUT = map[int]retryLogic{
...
 	return &RetryMiddleware{
 		nextRoundtripper: roundTripper,
 		config:           config,
 		logger:           logger,
+		onRetryNotify:    notify,
 	}
 }
 
-func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
+func (rm *RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) { //nolint:gocyclo // retry loop with body buffering and notify wiring
```
robertolopezlopez: I do not see the reason you have changed the receiver to a pointer, could you please tell? :-) (also in other receiving functions)

pkg/networking/middleware/retry_middleware.go
```diff
@@ -123,15 +147,34 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+				MaxAttempts: *cachedMaxRetries,
+				Err:         retryError,
+			}
 		}
 
 		return response, nil
 	}
 
+	notify := func(err error, duration time.Duration) {
+		if rm.onRetryNotify == nil {
```
robertolopezlopez: please pass this as a parameter to avoid referencing variables outside of the closure, and/or ideally: make this function a real function

pkg/networking/middleware/retry_middleware_test.go
robertolopezlopez: maybe worth using that variable here instead of using this magic number?

pkg/networking/middleware/retry_middleware.go
danskmt: that's a good point, it is actually suppose to be only 1, but instead of "failing hard", just accepting the first one here should be enough. I'm not sure what is the best semantics to use here, but using panic is not one of them

pkg/networking/middleware/retry_middleware.go
danskmt: No strong reason, but doesn't seem correct to not be a pointer as in the builder for example we use a pointer...
I'll remove it just to keep same as before.

pkg/networking/middleware/retry_middleware.go
danskmt: Thanks for the suggestion! will update

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: my head was rather leaning towards removing the "variadic" part

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: actually that's a good reason - consistency, if the builder returns pointer... sounds ok to me

pkg/networking/middleware/retry_middleware.go
```diff
@@ -123,20 +145,44 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		backoff.WithNotify(func(err error, duration time.Duration) {
+			rm.notifyRetry(err, duration, reqCtx)
+		}),
+	)
 
 	finalError = rm.filterRetryError(finalError, actualAttempts)
 	return finalResponse, finalError
 }
 
+func (rm RetryMiddleware) notifyRetry(err error, duration time.Duration, ctx context.Context) {
```
PeterSchafer: nitpick: ctx is often the first parameter

pkg/networking/middleware/retry_middleware.go
```diff
@@ -44,17 +48,39 @@ var statusCodesToRetryLUT = map[int]retryLogic{
...
+
 type RetryMiddleware struct {
 	nextRoundtripper http.RoundTripper
 	config           configuration.Configuration
 	logger           *zerolog.Logger
+	errorHandler     networktypes.ErrorHandlerFunc
 }
 
-func NewRetryMiddleware(config configuration.Configuration, logger *zerolog.Logger, roundTripper http.RoundTripper) *RetryMiddleware {
+func NewRetryMiddleware(config configuration.Configuration, logger *zerolog.Logger, roundTripper http.RoundTripper, errorHandler ...networktypes.ErrorHandlerFunc) http.RoundTripper {
```
PeterSchafer: Suggestion: Please consider using the Functional Options Pattern for optional parameter

PeterSchafer: currently it appears that it is valid to add multiple error handler functions

danskmt: @PeterSchafer thanks for the suggestion! I improved it.

PeterSchafer (approved):

PR #613, opened 2026-05-22
fix: isolate test environment from external variables

robertolopezlopez:
### Description

This PR adds an extendable mechanism to ignore environment variables in unit tests.

### Checklist

- [x] Linted (`make lint`)

### DIFF
--- a/pkg/configtest/env.go
+++ b/pkg/configtest/env.go
@@ -2,13 +2,13 @@ package configtest
 
 import "testing"
 
-// APIURLViperEnvironmentKey is the AutomaticEnv name for configuration.API_URL (viper key "snyk_api").
+// snykAPIKey is the AutomaticEnv name for configuration.API_URL (viper key "snyk_api").
 // Keep aligned with pkg/configuration/constants.go.
-const APIURLViperEnvironmentKey = "SNYK_API"
+const snykAPIKey = "SNYK_API"
 
 // KnownLeakEnvironmentKeys is cleared when [IsolateEnvironmentForTest] is called with no arguments.
 var KnownLeakEnvironmentKeys = []string{
-	APIURLViperEnvironmentKey,
+	snykAPIKey,
 }
 
 // IsolateEnvironmentForTest runs t.Setenv(k, "") for each key. With no keys it clears [KnownLeakEnvironmentKeys].
--- a/pkg/configtest/env.go
+++ b/pkg/configtest/env.go
@@ -1,6 +1,8 @@
 package configtest
 
-import "testing"
+import (
+	"testing"
+)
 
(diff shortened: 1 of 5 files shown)


pkg/configtest/env.go
```diff
@@ -0,0 +1,28 @@
...
+import "testing"
+
+// snykAPIKey is the AutomaticEnv name for configuration.API_URL (viper key "snyk_api").
+// Keep aligned with pkg/configuration/constants.go.
+const snykAPIKey = "SNYK_API"
+
+// KnownLeakEnvironmentKeys is cleared when [IsolateEnvironmentForTest] is called with no arguments.
+var KnownLeakEnvironmentKeys = []string{
+	snykAPIKey,
+}
```
danskmt, line 14: I guess you can use the value from pkg/configuration/constants.go directly here

pkg/configtest/env.go
```diff
@@ -0,0 +1,28 @@
...
+	toClear := keys
+	if len(toClear) == 0 {
+		toClear = KnownLeakEnvironmentKeys
+	}
+	for _, k := range toClear {
+		if k == "" {
+			continue
+		}
+		t.Setenv(k, "")
+	}
```
danskmt, line 31: should consider if is lowercase / uppercase?

pkg/configtest/env.go
robertolopezlopez: there would be a circular dependency

image

pkg/configtest/env.go
robertolopezlopez: Not really, I believe:

1. Viper CLI transforms every environment variable key to uppercase
2. Our `configuration.go` also looks up the keys as uppercase (line 700).
3. `configuration_test.go` in line 490 also uses `strings.ToUpper()`

pkg/configtest/env.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// IsolateEnvironmentForTest runs t.Setenv(k, "") for each key. With no keys it clears [KnownLeakEnvironmentKeys].
+// With keys, only those are cleared (not merged with defaults). Empty keys are skipped.
+func IsolateEnvironmentForTest(t *testing.T, keys ...string) {
+	t.Helper()
+	toClear := keys
+	if len(toClear) == 0 {
+		toClear = KnownLeakEnvironmentKeys
+	}
```
danskmt, line 25: Should we ALWAYS consider ignoring the **KnownLeakEnvironmentKeys** values?
If so, we should concat always.
In the current way, if parameters for keys are passed, the **KnownLeakEnvironmentKeys** are not being considered

danskmt (approved):

pkg/configtest/env.go
robertolopezlopez: this behavior has been documented in my last commit

PR #615, opened 2026-05-25
chore: automatic semver [CLI-1465]

octavian-snyk:
### Description

This PR introduces SemVer release tagging workflow for this repo.

Note: When this PR is merged, the repository should bump to version: v0.1.0. Please tag the last commit manually to v0.1.0 before merging this PR.

[CLI-1465](https://snyksec.atlassian.net/browse/CLI-1465)

[CLI-1465]: https://snyksec.atlassian.net/browse/CLI-1465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,71 @@
+name: Release
+on:
+  push:
+    branches:
+      - 'main'
+      - 'release/**'
+
+permissions: {}
+
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  compute-semver:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      id-token: none
+    outputs:
+      next: ${{ steps.semver.outputs.next }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: '0'
+          persist-credentials: false
+
+      - name: Get Next Version
+        id: semver
+        uses: ietf-tools/semver-action@c90370b2958652d71c06a3484129a4d423a6d8a8  # v1.11.0
+        with:
+          token: ${{ github.token }}
+          branch: ${{ github.ref_name }}
+          minorList: feat
+          patchList: fix, revert
+          noVersionBumpBehavior: warn
+
+      - name: Validate version string
+        if: ${{ steps.semver.outputs.next != '' }}
+        env:
+          NEXT_VERSION: ${{ steps.semver.outputs.next }}
+        run: |
+          if ! printf '%s' "$NEXT_VERSION" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "Unexpected version format: '$NEXT_VERSION'" >&2
+            exit 1
+          fi
+
+  release:
+    needs: compute-semver
+    if: needs.compute-semver.outputs.next != ''
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
+      contents: write
+      id-token: none
+    env:
+      NEXT_VERSION: ${{ needs.compute-semver.outputs.next }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+      - name: Tag commit
+        run: git tag -- "$NEXT_VERSION"
+
+      - name: Push tag
+        run: git push origin -- "refs/tags/$NEXT_VERSION"
+
+      - name: Release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release create "$NEXT_VERSION" --target "$GITHUB_SHA" --generate-notes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 This repo is intended for internal (Snyk) contributions only at this time.
 
-# Creating a new extension
+## Creating a new extension
 
 The `go-application-framework` makes it easy to add functionality to the Snyk CLI by taking care of the orchestration, requirements, and state management when running the CLI; this all happens behind the scenes via the packages provided by the framework. It means we can focus on building the core business logic for our extension and not worry about much else.
 
@@ -12,7 +12,7 @@ Creating a new extension using the `go-application-framework` can be done in thr
 2. Register the extension with the workflow engine
 3. Initialise the extension when using the `go-application-framework`
 
-# Create the extension workflow
+### Create the extension workflow
 
 The extension workflow is the main component of your extension, it should:
 
@@ -20,11 +20,11 @@ The extension workflow is the main component of your extension, it should:
 2. Initialise your workflow (registering the extension with the workflow engine should happen here)
 3. Have an entry point containing the extension business logic
 
-# Creating `snyk whoami`
+### Creating `snyk whoami`
 
 The `snyk whoami` command was created following the steps outlined above; the following outlines the extension implementation in detail.
 
-## Uniquely identify your workflow
+#### Uniquely identify your workflow
 
 Before we can start creating our extension workflow, we'll need to create our workflow identifier. An identifier is required by the workflow engine to identify our workflow. The workflow identifier should also contain the name of the new extension command.
 
@@ -39,7 +39,7 @@ var workflowName = "whoami"
 var WORKFLOWID_WHOAMI workflow.Identifier = workflow.NewWorkflowIdentifier(workflowName)
 ```
 
-## Initialise your workflow
+#### Initialise your workflow
 
 Now that we have a workflow identifier, we can continue with the implementation of the workflow's initialiser via an `Init()` function.
 
@@ -58,19 +58,19 @@ func Init(engine workflow.Engine) error {
 }
 ```
 
-### Initialise configuration
+##### Initialise configuration
 
 We use the `pflag` package in order to create POSIX/GNU-style --flags. The extension should support the `--json` flag, so we add it to the configuration via `whoAmIConfig.Bool()`.
 
-### Mark experimental versions of an extension workflow
+##### Mark experimental versions of an extension workflow
 
 There is a common pattern to mark workflows as experimental by using [MarkAsExperimental()](https://github.com/snyk/go-application-framework/blob/main/pkg/local_workflows/config_utils/experimental.go#L28) in order to indicate early non production ready versions of a workflow. In the CLI this will automatically reflect in a required argument `--experimental`, if not provided the CLI will show an error and not call the workflow.
 
-### Register with the workflow engine
+##### Register with the workflow engine
 
 Next we must register the extension with the workflow engine. The `workflow` package is used again here in, firstly to abstract away the engine/workflow registration logic via `engine.Register`, and again to configure the workflow's configuration via `workflow.ConfigurationOptionsFromFlagset`.
 
-## Implement business logic via the workflow entry point
+#### Implement business logic via the workflow entry point
 
 Now we can define the business logic for our extension workflow. This can be done in an `entryPoint()` function, which has two parameters; `invocationContext` and `input []workflow.Data`.
 
@@ -128,7 +128,7 @@ func entryPoint(invocationCtx workflow.InvocationContext, _ []workflow.Data) (ou
 }
 ```
 
-### Get necessary invocation context objects
+##### Get necessary invocation context objects
 
 The first step is to retrieve the necessary objects the workflow requires.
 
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -296,3 +296,22 @@ make regenerate-expected
 ```
 
 This runs the presenter tests with `UFM_REGEN=1`, which overwrites the expected files with the current presenter output and skips the comparison. The same effect can be achieved by flipping the `regenerateExpectedFiles` constant at the top of `internal/presenters/presenter_ufm_test.go` to `true` and running the tests directly (handy when stepping through with a debugger). Review the diff before committing.
+
+# Creating commits
+
+Commits must follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) structure.
+
```
danskmt: In the [CLI's CONTRIBUTING.md](https://github.com/snyk/cli/blob/main/CONTRIBUTING.md) also says about _Writing commit messages_, by specifying "what" rather than "how". Maybe is worth saying it here as well.

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Get Next Version
+        id: semver
+        uses: ietf-tools/semver-action@v1
+        with:
+          token: ${{ github.token }}
+          branch: ${{ github.ref_name }}
+          minorList: feat, revert
+          patchList: fix
+          noVersionBumpBehavior: silent
```
danskmt: in `code-client-go` the config is:
```
     - name: Get Next Version
        id: semver
        uses: ietf-tools/semver-action@v1
        with:
          token: ${{ github.token }}
          branch: main
          noVersionBumpBehavior: patch
```

here fixes branch to **main** instead getting _ref_name_, and **noVersionBumpBehavior** sets to _patch_.
Are we sure about not following the pattern that has in `code-client-go`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Create tag and Push tag
+        if: ${{ steps.semver.outputs.next != '' }}
+        run: |
+          git tag ${{ steps.semver.outputs.next }}
+          git push origin ${{ steps.semver.outputs.next }}
```
danskmt: maybe have separate steps for creating tag and then pushing? 
ref: https://github.com/snyk/code-client-go/blob/main/.github/workflows/release.yml

CONTRIBUTING.md
```diff
@@ -296,3 +296,22 @@ make regenerate-expected
...
+| Type            | Description                                      | Version bump |
+| --------------- | ------------------------------------------------ | ------------ |
+| feat            | A new user-facing feature.                       | minor        |
+| fix             | A bug fix for an existing feature.               | patch        |
+| refactor        | Changes which do not affect existing features.   | none         |
+| test            | Changes to tests for existing features.          | none         |
+| docs            | Changes to documentation for existing features.  | none         |
+| chore           | Build, workflow and pipeline changes.            | none         |
+| revert          | Reverting a previous commit.                     | minor        |
+| BREAKING CHANGE | Commit message contains BREAKING CHANGE          | major        |
```
danskmt: Not sure if this BREAKING CHANGE should be in this table as the others are defined at the beginning of the commit message, and BREAKING CHANGE is something that is probably in the commit's body description

CONTRIBUTING.md
```diff
@@ -296,3 +296,22 @@ make regenerate-expected
 ```
 
 This runs the presenter tests with `UFM_REGEN=1`, which overwrites the expected files with the current presenter output and skips the comparison. The same effect can be achieved by flipping the `regenerateExpectedFiles` constant at the top of `internal/presenters/presenter_ufm_test.go` to `true` and running the tests directly (handy when stepping through with a debugger). Review the diff before committing.
+
+# Creating commits
```
robertolopezlopez: I guess this is not the main header. Maybe `##` or `###`?

octavian-snyk: Thanks for keeping an eye on the formatting. 
It's formatted like this to keep the GitHub outline preview tidy. Using ## or ### I believe would accidentally nest this under "Creating a new extension". I might be misunderstanding the layout you have in mind.

robertolopezlopez: but in theory there should be one header (`#`) per file - is this not the case here? oh I can see there are 3 defined in this file, I do not think that's correct (although not your task, I would fix that)

.github/workflows/release.yml
PeterSchafer: 1. We are using ref_name to ensure that it works correctly for the different branches it runs on.
2. `noVersionBumpBehavior: silent` ensures that chores are not bumping the version I think.

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - 'main'
+      - 'release/**'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
```
PeterSchafer: Suggestion: please pin the version

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          fetch-depth: '0'
+
+      - name: Get Next Version
+        id: semver
+        uses: ietf-tools/semver-action@v1
```
PeterSchafer: Suggestion: please pin the version

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          token: ${{ github.token }}
+          branch: ${{ github.ref_name }}
+          minorList: feat, revert
+          patchList: fix
+          noVersionBumpBehavior: silent
+
+      - name: Create tag
+        if: ${{ steps.semver.outputs.next != '' }}
+        run: git tag ${{ steps.semver.outputs.next }}
```
PeterSchafer: Suggestion: Let's ensure that `steps.semver.outputs.next` is a valid semver and not an arbitrary string

PeterSchafer: same below

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+      next: ${{ steps.semver.outputs.next }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: '0'
+          persist-credentials: false
+
+      - name: Get Next Version
+        id: semver
+        uses: ietf-tools/semver-action@c90370b2958652d71c06a3484129a4d423a6d8a8  # v1.11.0
```
robertolopezlopez, line 31: can't you just use the tag instead of the commit hash? If I look at their github repository, it seems so https://github.com/ietf-tools/semver-action#full-basic-workflow-with-release


```suggestion
        uses: ietf-tools/semver-action@v1.11.0
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+  compute-semver:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      id-token: none
+    outputs:
+      next: ${{ steps.semver.outputs.next }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```
robertolopezlopez, line 24: Same here

https://github.com/ietf-tools/semver-action#full-basic-workflow-with-release


```suggestion
      - uses: actions/checkout@v6.0.2
```

robertolopezlopez: Do the same all along this file, please

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+          fetch-depth: '0'
+          persist-credentials: false
+
+      - name: Get Next Version
+        id: semver
+        uses: ietf-tools/semver-action@c90370b2958652d71c06a3484129a4d423a6d8a8  # v1.11.0
+        with:
+          token: ${{ github.token }}
+          branch: ${{ github.ref_name }}
+          minorList: feat, revert
```
robertolopezlopez: is "revert" really minor? This is a little bit ambiguous, as we may be reverting a fix as well

I would consider it as patch

CONTRIBUTING.md
```diff
@@ -296,3 +296,47 @@ make regenerate-expected
...
+
+| Type            | Description                                      		| Version bump |
+| --------------- | ------------------------------------------------------- | ------------ |
+| feat            | A new user-facing feature.                       		| minor        |
+| fix             | A bug fix for an existing feature.               		| patch        |
+| refactor        | Changes which do not affect existing features.   		| none         |
+| test            | Changes to tests for existing features.          		| none         |
+| docs            | Changes to documentation for existing features.  		| none         |
+| chore           | Build, workflow and pipeline changes.            		| none         |
+| revert          | Reverting a previous commit.                     		| minor        |
```
robertolopezlopez: same comment as [here](https://github.com/snyk/go-application-framework/pull/615/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R35)

.github/workflows/release.yml
octavian-snyk: I believe the intention is to use the `revert:` commit type when rolling back a feature. If we don't, our only option might be using `feat:` to undo a feature. On the other hand, if we're reverting a bug fix or a patch, we could use `fix: revert ...` which will trigger a patch release.

[CLI-1465 comment](https://snyksec.atlassian.net/browse/CLI-1465?focusedCommentId=1529396)

robertolopezlopez: that's a nice hack in my opinion, but if the plan is that it's fine by me

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+    if: needs.compute-semver.outputs.next != ''
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
+      contents: write
+      id-token: none
+    env:
+      NEXT_VERSION: ${{ needs.compute-semver.outputs.next }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```
robertolopezlopez, line 60: also here: please use the release tag instead of commit hash

CatalinSnyk: Commit SHAs are the recommended way for pinning AFAIK: https://wellarchitected.github.com/library/application-security/recommendations/actions-security/#pinning-actions-based-on-a-version-tag & https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/#enforce-sha-pinning

.github/workflows/release.yml
j-luong: Actually, on reflection... I agree with @robertolopezlopez, our typical use of `revert` is when we can't apply a `fix` and have to undo a whole feature. If a `minor` is only for introducing a new feature, then a `patch` makes sense for a `revert`.

An argument can be made that `major` could be used if we're reverting something that has already been widely adopted, but we usually only apply `revert` on recently released features with little/no establed usage

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+          persist-credentials: false
+
+      - name: Get Next Version
+        id: semver
+        uses: ietf-tools/semver-action@c90370b2958652d71c06a3484129a4d423a6d8a8  # v1.11.0
+        with:
+          token: ${{ github.token }}
+          branch: ${{ github.ref_name }}
+          minorList: feat, revert
+          patchList: fix
```
j-luong: suggest: per my previous comment `patchList: fix, revert`

.github/workflows/release.yml
j-luong: suggest: `noVersionBumpBehavior: warn` will be more useful if we want some kind of log message when `Get Next Version` results in no version change.

CONTRIBUTING.md
```diff
@@ -296,3 +296,47 @@ make regenerate-expected
...
+| _any type_**!** | Appending a ! to any type denotes a breaking change.	| major        |
```
j-luong, line 336: question: is this tested/correct? I know it's in the conventional commit standard, but is it configured for the action we use, `ietf-tools/semver-action`? Looking at the [inputs](https://github.com/ietf-tools/semver-action#inputs) it doesn't seem like it.

j-luong:
> Note: When this PR is merged, the repository will bump to version: v1.0.0.

Why is this? The latest tag is `v0.1.0` and none of the commits in this PR should result in a major version bump (or any bump in fact).

CONTRIBUTING.md
octavian-snyk: I tested it, it works. 😅

j-luong (approved):

PR #619, opened 2026-06-02
feat(configuration): add GetDefaultValueFunc to Configuration interface

nick-y-snyk:
## Summary

- Adds `GetDefaultValueFunc(key string) DefaultValueFunction` to the `Configuration` interface and `extendedViper` implementation
- Regenerates the `MockConfiguration` mock
- Enables callers to capture an existing default-value function before overwriting it with `AddDefaultValue`, allowing layered/chained behaviour

## Motivation

Currently `AddDefaultValue` replaces the registered function for a key with no way to retrieve the prior one. This makes it impossible to build wrappers that guard (e.g. skip network calls when offline/unauthenticated) and then delegate to the original function. Without this, a wrapper permanently displaces the original GAF default func (e.g. `defaultFuncOrganization`), silently breaking org resolution for callers that go through `conf.GetString(configuration.ORGANIZATION)` when no cached value exists.

## Test plan

- [ ] `go test ./pkg/configuration/...` passes (180 tests, verified locally)
- [ ] `go build ./pkg/...` passes
- [ ] Mock regenerated via `go generate ./pkg/configuration/...`

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -57,6 +57,7 @@ type Configuration interface {
 	AddFlagSet(flagset *pflag.FlagSet) error
 	AllKeys() []string
 	AddDefaultValue(key string, defaultValue DefaultValueFunction)
+	GetDefaultValueFunc(key string) DefaultValueFunction
 	AddAlternativeKeys(key string, altKeys []string)
 	GetAlternativeKeys(key string) []string
 	GetAllKeysThatContainValues(key string) []string
@@ -674,6 +675,15 @@ func (ev *extendedViper) AddDefaultValue(key string, defaultValue DefaultValueFu
 	defer ev.mutex.Unlock()
 
 	ev.defaultValues[key] = defaultValue
+}
+
+// GetDefaultValueFunc returns the DefaultValueFunction registered for key, or nil if none is registered.
+// This allows callers to capture the existing function before overwriting it with AddDefaultValue,
+// enabling layered/chained default-value behaviour.
+func (ev *extendedViper) GetDefaultValueFunc(key string) DefaultValueFunction {
+	ev.mutex.RLock()
+	defer ev.mutex.RUnlock()
+	return ev.defaultValues[key]
 }
 
 // AddAlternativeKeys adds alternative keys to the configuration.
--- a/pkg/mocks/configuration.go
+++ b/pkg/mocks/configuration.go
@@ -226,6 +226,20 @@ func (mr *MockConfigurationMockRecorder) GetBoolWithError(key interface{}) *gomo
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetBoolWithError", reflect.TypeOf((*MockConfiguration)(nil).GetBoolWithError), key)
 }
 
+// GetDefaultValueFunc mocks base method.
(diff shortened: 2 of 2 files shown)


pkg/configuration/configuration.go
```diff
@@ -676,6 +677,15 @@ func (ev *extendedViper) AddDefaultValue(key string, defaultValue DefaultValueFu
 	ev.defaultValues[key] = defaultValue
 }
 
+// GetDefaultValueFunc returns the DefaultValueFunction registered for key, or nil if none is registered.
+// This allows callers to capture the existing function before overwriting it with AddDefaultValue,
+// enabling layered/chained default-value behaviour.
+func (ev *extendedViper) GetDefaultValueFunc(key string) DefaultValueFunction {
```
bastiandoetsch, line 683: [Suggestion] No direct test for `GetDefaultValueFunc` — every other accessor in this file has explicit coverage and `AddDefaultValue` is tested in ~12 places, but there's no test here. The 'capture-before-overwrite' layering use case the method exists to enable is non-obvious and currently unproven. A ~10-line table test would lock it in: register a func via `AddDefaultValue`, assert `GetDefaultValueFunc(key)` returns a non-nil func whose result matches, and assert `GetDefaultValueFunc("unregistered")` returns nil.

— AI review

bastiandoetsch:
[Suggestion · release-notes] Adding a method to the exported `Configuration` interface is a breaking change for any external consumer that hand-rolls the interface (rather than embedding the generated `MockConfiguration`). That's a conventional, accepted Go evolution cost — but worth a changelog note and at least a minor SemVer bump so downstream upgraders aren't surprised. No code change recommended (an optional side-interface would add more complexity than the break costs).

— AI review

bastiandoetsch (approved):
Full verification pass (semantic, adversarial, security, code-review). Clean, minimal addition: `GetDefaultValueFunc` uses `RLock` symmetric to `AddDefaultValue`'s write `Lock`, the mock is properly regenerated (interface/impl/mock all consistent, build+vet pass), and nil-on-missing-key is correct and documented. No correctness or security issues. Two non-blocking suggestions: add a unit test for the new accessor, and note the public-interface break in release notes. Approving.

— AI review

pkg/configuration/configuration.go
```diff
@@ -676,6 +677,15 @@ func (ev *extendedViper) AddDefaultValue(key string, defaultValue DefaultValueFu
 	ev.defaultValues[key] = defaultValue
 }
 
+// GetDefaultValueFunc returns the DefaultValueFunction registered for key, or nil if none is registered.
+// This allows callers to capture the existing function before overwriting it with AddDefaultValue,
```
basti-snyk, line 681: **Suggestion (non-blocking):** The doc comment recommends callers "capture the existing function before overwriting it with `AddDefaultValue`" to chain defaults. That get-then-set sequence is not atomic: `GetDefaultValueFunc` takes `RLock` and releases it, then `AddDefaultValue` takes `Lock` separately. Under concurrent registration two goroutines can both read the same base func and one chain link is silently lost. The accessor itself is race-free — only the composition pattern the comment endorses is racy. Consider either noting that chaining must happen during single-threaded setup, or offering an atomic read-modify-write helper. In practice config is wired at init single-threaded, so severity is low.

— AI review

PR #620, opened 2026-06-03
feat: add opt-in middleware to block unauthenticated Snyk API requests [IDE-1988]

nick-y-snyk:
### Description

Extends `AuthHeaderMiddleware` with an opt-in circuit-breaker: when `internal_stop_requests_without_auth` is `true`, requests to Snyk API URLs with no `Authorization` header are blocked with a synthetic 401 instead of hitting the wire.

This prevents unauthenticated requests from triggering retry/error storms, as part of the auth/network state gating effort in [IDE-1988].

**Changes:**
- `AuthHeaderMiddleware.RoundTrip` checks the flag and short-circuits with 401 + `UnauthorisedError` if no token was set for a Snyk API URL
- New config constant `STOP_REQUESTS_WITHOUT_AUTH` (default: `false`)
- The gate is enforced in the round-tripper only; `AddHeaders` callers are out of scope (documented)

### Checklist

- [ ] Linted (`make lint`)

[IDE-1988]: https://snyksec.atlassian.net/browse/IDE-1988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

___

### **PR Type**
Enhancement

___

### **Description**
- Introduce opt-in middleware to block unauthenticated Snyk API requests.

- Prevents unauthenticated requests from hitting the wire.

- Refactored `AuthHeaderMiddleware` for clearer logic.

- Added comprehensive tests for new functionality.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[HTTP Request] --> B{AuthHeaderMiddleware};
  B -- Config: STOP_REQUESTS_WITHOUT_AUTH enabled --> C{Check Authorization Header};
  C -- Missing --> D[Block with 401];
  C -- Present --> E[Proceed to Next RT];
  B -- Config: STOP_REQUESTS_WITHOUT_AUTH disabled --> E;
  B -- Not Snyk API --> E;
  E --> F[Next HTTP RoundTripper];
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Configuration changes</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>constants.go</strong><dd><code>Add new configuration constant for request blocking</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/configuration/constants.go

<ul><li>Defines a new boolean configuration constant <br><code>STOP_REQUESTS_WITHOUT_AUTH</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/620/changes#diff-b346bdeac320c8fe8bc31a635e59da3e8bf59c7c00ed11bf8cd6a17dccda15b7">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>auth_header.go</strong><dd><code>Enhance AuthHeaderMiddleware with unauthenticated request blocking</code></dd></summary>
<hr>

pkg/networking/middleware/auth_header.go

<ul><li>Implements an opt-in feature to block unauthenticated Snyk API <br>requests with a synthetic 401.<br> <li> Adds logging for authentication checks and blocked requests.<br> <li> Refactors <code>AddAuthenticationHeader</code> to return authentication requirement <br>status and integrates logger.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/620/changes#diff-d428da578d8f6b9c705434c936c9839326c0ea3b3100e1a78a035739a672f027">+53/-8</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>networking.go</strong><dd><code>Update network implementation to use logger middleware</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/networking/networking.go

<ul><li>Updates the network implementation to use the logger-aware middleware <br>constructor.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/620/changes#diff-99beec8232d7e64bc806ee6a0d9ee88bbff437a1a93094027686253ab80725a5">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>auth_header_test.go</strong><dd><code>Add tests for new middleware functionality</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/networking/middleware/auth_header_test.go

<ul><li>Adds comprehensive tests for the new unauthenticated request blocking <br>feature.<br> <li> Tests cover various flag states, token presence, and URL types.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/620/changes#diff-e0788a25efc84932b61127fa4bb772464fc6c6a2c14ebf8007fa37c1119d35bd">+107/-0</a>&nbsp; </td>

</tr>

</details>

___


pkg/networking/middleware/no_auth_intercept_middleware.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	apiUrl := m.config.GetString(configuration.API_URL)
+	additionalSubdomains := m.config.GetStringSlice(configuration.AUTHENTICATION_SUBDOMAINS)
+	additionalUrls := m.config.GetStringSlice(configuration.AUTHENTICATION_ADDITIONAL_URLS)
+
+	requiresAuth, err := ShouldRequireAuthentication(apiUrl, req.URL, additionalSubdomains, additionalUrls)
+	if err != nil || !requiresAuth {
+		return m.next.RoundTrip(req)
+	}
```
robertolopezlopez: hi - if `err != nil`: shouldn't it be handled or logged in any way?

Maybe muting the error is acceptable now but on the long term it might lead to trouble.

nick-y-snyk: Good catch! The error here can only come from `url.Parse` inside `DeriveSubdomainUrl`, which is called with the `API_URL` config value — that's always validated before storage, so in practice it's unreachable. That said, swallowing it silently is still bad form.

I've refactored the middleware since this comment was written (the logic now lives in `AuthHeaderMiddleware` rather than a separate type), and I've added explicit debug logging for the error case — it lets the request through and logs why. Fixed in e340c52.

pkg/networking/middleware/auth_header_test.go
```diff
@@ -106,6 +111,95 @@ func Test_AddAuthenticationHeader(t *testing.T) {
...
+		next := roundTripFunc(func(r *http.Request) (*http.Response, error) {
+			called = true
+			return &http.Response{StatusCode: http.StatusOK}, nil
+		})
+		ctrl := gomock.NewController(t)
+		auth := mocks.NewMockAuthenticator(ctrl)
+		auth.EXPECT().AddAuthenticationHeader(gomock.Any()).Return(nil).Times(1)
+
+		m := middleware.NewAuthHeaderMiddlewareWithLogger(newConfig(true), auth, next, &logger)
+		resp, err := m.RoundTrip(newRequest(t, "https://app.snyk.io/rest/endpoint"))
```
bastiandoetsch, line 169: [Should Fix] `[1/4]` The test "flag on, no token, Snyk API URL — returns 401" uses `https://app.snyk.io/rest/endpoint` as the request URL, but `ShouldRequireAuthentication` checks whether the URL has the `api.snyk.io` prefix via `api.GetCanonicalApiUrl`. If `app.snyk.io` does not canonicalize to the API host, `requiresAuth` will be false and the 401 block is never reached — the test passes vacuously while providing false confidence that the gate works.

Fix: use `https://api.snyk.io/rest/endpoint` to match the configuration value set in `newConfig`, or add an assertion that `ShouldRequireAuthentication` returns true for the chosen URL before relying on it.
```go
resp, err := m.RoundTrip(newRequest(t, "https://api.snyk.io/rest/endpoint"))
```

— AI review

bastiandoetsch:
[Should Fix] `[3/4] ⚑` **retry_middleware_test.go:486** — The test case name "Retryable status code (429) Retry-After takes precedence over X-RateLimit-Reset" documents a policy that does not exist in the implementation. `retry_middleware.go:260` uses `max(fixRetryDelay, fixXRateLimitReset)` — the larger value wins, not Retry-After specifically. With `Retry-After: 3` and `X-RateLimit-Reset: 10` the result is 10 s; the assertion passes by coincidence. A developer reading this test will infer the wrong mental model and remove the `max()`, introducing a regression.

Fix: rename the test case and add a symmetric case where Retry-After exceeds X-RateLimit-Reset.
```go
name: "Retryable status code (429) both headers present — picks the larger value (max strategy)",
```
Confirmed independently by 3 of 4 reviewers.

— AI review

bastiandoetsch:
[Should Fix] `[3/4] ⚑` **retry_middleware.go:190** — `notifyRetry` mutates `retryErr.Duration` in-place via the pointer returned by `errors.As(err, &retryErr)`. The `RetryAttemptError` is the same struct that the `backoff` library holds in its retry loop; writing to it while it is live creates an implicit aliasing dependency on the library's internal sequencing. The mutation is safe in the current version (notify is called synchronously), but it is fragile against library upgrades and makes the function read as a side-channel state injection rather than a clean notification.

Fix: copy the struct before mutating.
```go
notifCopy := *retryErr
notifCopy.Duration = duration
retryAttemptNotification(&notifCopy)
```
Confirmed independently by 3 of 4 reviewers.

— AI review

bastiandoetsch:
[Should Fix] `[2/4]` **retry_middleware_test.go:697** — The test "500 exhaustion does not invoke callback" has a vacuous assertion body and a contradictory name. `notifyRetry` only calls `errorHandler` when `retryAttemptNotification` returns `ok=true`, which exclusively happens for status 429. For a 500, `tracker.errors` stays empty, so the `for _, e := range tracker.errors` loop at line 718 never executes — all assertions inside it are dead. The test passes but verifies nothing.

Fix: replace the vacuous loop with an explicit assertion and rename the test to match.
```go
t.Run("500 exhaustion does not invoke errorHandler — only 429 does", func(t *testing.T) {
    // ...setup...
    assert.Empty(t, tracker.errors, "500 should not trigger errorHandler")
})
```

— AI review

bastiandoetsch:
[Should Fix] `[1/4]` **retry_middleware.go:260** — When both `Retry-After` and `X-RateLimit-Reset` headers are present, `max(fixRetryDelay, fixXRateLimitReset)` always picks the *larger* value. RFC 6585 §4 specifies `Retry-After` as the server-authoritative wait time for 429 responses. If a server sets `Retry-After: 3` (token bucket refills quickly) alongside `X-RateLimit-Reset: 30` (window end), the client will wait 30 s instead of the server-intended 3 s, reducing effective throughput by 10×.

Fix: prefer `Retry-After` when it is present, falling back to `X-RateLimit-Reset` only if `Retry-After` is absent. If the intentional design is to always wait for the longer of the two, document that choice explicitly.

— AI review

bastiandoetsch:
[Should Fix] `[1/4]` **retry_middleware_test.go:217,245** — The test names "429 with Retry-After beyond max wait returns SNYK-0001 and last response" and "429 with X-RateLimit-Reset beyond max wait returns SNYK-0001 and last response" are inaccurate. When `Retry-After` or `X-RateLimit-Reset` exceeds `maxRetryAfter`, `filterRetryError` suppresses the error and returns `nil` — no SNYK-0001 error reaches the caller. The tests correctly assert `NoError` and the 429 response, but the names imply an error is surfaced.

Fix: rename both tests.
```go
"429 with Retry-After beyond max wait — blocks immediately, returns last 429 response with no error"
```

— AI review

bastiandoetsch:
[Should Fix] `[1/4]` **pkg/configtest/env.go:30** — `IsolateEnvironmentForTest` uses `t.Setenv(k, "")` to clear each key. Setting a variable to an empty string is not the same as unsetting it: `os.LookupEnv(k)` returns `("", true)` after `Setenv(k, "")`, rather than `("", false)`. Any code under test that uses `os.LookupEnv` to distinguish "configured to empty" from "not set" (a pattern used in several configuration paths) will behave differently in isolated tests than in production where the variable is absent, producing false-passing tests.

Fix: use `t.Unsetenv(k)` instead of `t.Setenv(k, "")`.
```go
for _, k := range toClear {
    if k == "" { continue }
    t.Unsetenv(k)
}
```

— AI review

bastiandoetsch:
[Should Fix] `[2/4]` **pkg/networking/networking.go:196-200** — `n.errorHandler` is registered at two distinct points in the middleware stack with no documented contract distinguishing what each layer delivers. The retry middleware (line 196) fires the handler with a `TooManyRequestsError` on 429 retries; `NewNetworkStackErrorHandlerMiddleware` (line 199) fires it for non-retryable terminal errors. Callers of `SetErrorHandler` must inspect the error type to know the context, but there is nothing in the API contract documenting what types they should expect.

Fix: add inline comments at the middleware-construction site.
```go
// RetryMiddleware fires errorHandler with *TooManyRequestsError on each 429 retry notification.
crt = middleware.NewRetryMiddleware(n.config, n.logger, crt, middleware.WithErrorHandler(n.errorHandler))
// NetworkStackErrorHandlerMiddleware fires errorHandler for non-retryable terminal transport errors.
if n.errorHandler != nil {
    crt = middleware.NewNetworkStackErrorHandlerMiddleware(crt, n.errorHandler)
}
```

— AI review

pkg/networking/middleware/auth_header.go
```diff
@@ -30,18 +34,42 @@ func NewAuthHeaderMiddleware(
...
 
 	// RoundTrippers should not modify the source request according to the docs, so cloning is used.
 	newRequest := request.Clone(request.Context())
-	err := AddAuthenticationHeader(n.authenticator, n.config, newRequest)
+	requiresAuth, err := addAuthenticationHeader(n.authenticator, n.config, newRequest)
 	if err != nil {
 		return nil, err
 	}
 
+	if n.config.GetBool(configuration.STOP_REQUESTS_WITHOUT_AUTH) && requiresAuth && newRequest.Header.Get("Authorization") == "" {
```
bastiandoetsch, line 66: [Suggestion] The no-token signal here is inferred from `Authorization == ""` after `addAuthenticationHeader` returns. That is exact for `tokenAuthenticator` (returns nil, sets no header → the 401 fires), but `oAuth2Authenticator` returns an *error* when its token is nil or refresh fails, so it trips the `err != nil` guard above and the caller gets `(nil, err)` instead of the synthetic 401.

Impact is bounded: `CreateAuthenticator` selects OAuth only when `IsSupported()` is true, which requires a non-nil token — so a plain no-token user always falls back to `tokenAuthenticator` and does get the 401. The only affected case is an OAuth user whose token is present but expired/unrefreshable. The request is still not sent (the error pre-empts it), so the breaker's primary goal holds; only the response contract differs (error vs 401).

Fix: document on the intercept that the breaker emits a 401 for token/PAT auth and an error for OAuth-refresh-failure, so callers do not assume a 401 is always produced.

— AI review

pkg/networking/middleware/auth_header.go
```diff
@@ -30,18 +34,42 @@ func NewAuthHeaderMiddleware(
...
+
 func (n *AuthHeaderMiddleware) RoundTrip(request *http.Request) (*http.Response, error) {
 	if request.URL == nil {
 		return n.next.RoundTrip(request)
 	}
 
 	// RoundTrippers should not modify the source request according to the docs, so cloning is used.
 	newRequest := request.Clone(request.Context())
-	err := AddAuthenticationHeader(n.authenticator, n.config, newRequest)
+	requiresAuth, err := addAuthenticationHeader(n.authenticator, n.config, newRequest)
```
bastiandoetsch, line 55: [Should Fix] When `ShouldRequireAuthentication` errors, `addAuthenticationHeader` returns `(false, err)` and `RoundTrip` returns `(nil, err)` without logging. Commit e340c52 ("log ShouldRequireAuthentication error instead of silently ignoring") added a debug-log-and-passthrough for exactly this case to resolve a review thread, but the later refactors dropped it — so the resolved thread now describes behavior the code no longer has. This matches the origin/main baseline, so it is not a functional regression; the issue is that the reviewed resolution was reverted.

Fix: restore the debug log on the error path, or reply on the thread that the resolution went back to baseline error-return semantics so reviewers are not misled.

— AI review

pkg/networking/middleware/auth_header.go
```diff
@@ -30,18 +34,42 @@ func NewAuthHeaderMiddleware(
...
 	if err != nil {
 		return nil, err
 	}
 
+	if n.config.GetBool(configuration.STOP_REQUESTS_WITHOUT_AUTH) && requiresAuth && newRequest.Header.Get("Authorization") == "" {
+		if n.logger != nil {
+			n.logger.Debug().Str("url", newRequest.URL.String()).Msg("request requires auth but no token present, blocking with 401")
+		}
+		return &http.Response{
+			StatusCode: http.StatusUnauthorized,
```
bastiandoetsch, line 71: [Suggestion] The synthetic 401 differs from a real upstream 401 in two ways that may matter to consumers:

1. It returns `(resp, nil)` — no error — so it bypasses `ResponseMiddleware`'s Error Catalog mapping. A real API 401 becomes an actionable `UnauthorisedError` ("Use `snyk auth` to authenticate."); a caller relying on the registered error handler gets a bare 401 with nil error instead.
2. Per RFC 7235 §3.1 a 401 should carry a `WWW-Authenticate` header; this one omits it.

Neither breaks a consumer in this repo today, but if the IDE auth-prompt UX keys off the catalog error, the synthetic path will not trigger it.

Fix: if that UX matters, synthesize the same `UnauthorisedError` the real-401 path produces; and add `Header.Set("WWW-Authenticate", "Bearer")` for spec-compliance.

— AI review

bastiandoetsch:
> ℹ️ Two non-blocking notes from a re-review — neither is on a line changed by this PR.

[Suggestion · non-blocking] The block is only enforced in `GetRoundTripper`'s middleware. `NetworkAccess.AddHeaders()` (networking.go:167) calls `middleware.AddAuthenticationHeader` directly, which has no intercept — a consumer that builds a request via `AddHeaders()` and sends it through their own client will still send the unauthenticated request with the flag on. If the flag's guarantee is meant to be global, `AddHeaders` needs the same gate; if it is intentionally scoped to the round-tripper, worth a one-line comment saying so.

[Suggestion · non-blocking] The PR description states "`Authenticator.AddAuthenticationHeader` now returns `(bool, error)`" — it does not (the interface still returns `error`; the `(bool, error)` approach was tried in b75257b and abandoned in favour of header inspection). Worth updating so reviewers are not evaluating the abandoned design — the stale "breaking interface change" bot comments stem from this same confusion.

— AI review

pkg/networking/middleware/auth_header_test.go
nick-y-snyk: `app.snyk.io` does canonicalize to `api.snyk.io` via `GetCanonicalApiUrl` (the `app.` → `api.` replacement), so `requiresAuth` is true and the 401 gate fires correctly — the test isn't vacuous. That said, the implicit canonicalization is easy to miss, so I've added an explicit pre-assertion that `ShouldRequireAuthentication` returns `true` for the URL before relying on it.

pkg/networking/middleware/auth_header.go
nick-y-snyk: Restored: when `ShouldRequireAuthentication` errors, `RoundTrip` now logs a debug message and passes the request through rather than returning the error. This matches the behavior from e340c52 that was dropped during subsequent refactors.

pkg/networking/middleware/auth_header.go
nick-y-snyk: Added a comment on the intercept documenting this contract: the 401 gate fires only for token/PAT auth (no token → no `Authorization` header → 401). For OAuth, an expired/unrefreshable token causes `addAuthenticationHeader` to return an error, which is handled above (log + passthrough) — so the `Authorization` check is never reached in that case.

pkg/networking/middleware/auth_header.go
nick-y-snyk: Addressed. The synthetic 401 now returns both the response (callers reading `resp.StatusCode` get 401) and the error catalog error via `errFromStatusCode(http.StatusUnauthorized)` — the same function `ResponseMiddleware` uses for real upstream 401s. No string duplication. Test updated to assert both.

andrewrobinsonhodges-snyk (approved):

PR #622, opened 2026-06-05
fix: preserve GetBody for HTTP/2 retry

robertolopezlopez:
### Description

This PR addressed problems executing Snyk test with reachability

### Checklist

- [x] Linted (`make lint`)

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Preserve `GetBody` for HTTP/2 retry compatibility.

- Avoid buffering `http.NoBody` when retries are enabled.

- Add comprehensive tests for body handling during retries.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[HTTP Request with Body] --> B(RetryMiddleware);
  B -- Body Exists & Retries > 1 --> C{Buffer Body};
  C --> D[Set req.GetBody function];
  B -- Retry Logic --> E{Invoke req.GetBody};
  D -- GetBody Function --> F(Returns Buffered Data);
  E -- Request Data --> F;
  B -- NoBody or Retries <= 1 --> G(No Body Buffering);
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>retry_middleware.go</strong><dd><code>Add `GetBody` for retry middleware</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/networking/middleware/retry_middleware.go

<ul><li>Implemented <code>req.GetBody</code> when the request body is buffered for retries.<br> <li> Added a check to ensure <code>http.NoBody</code> is not buffered, as it's already <br>empty and doesn't require re-reading.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/622/changes#diff-295d20b71bb7f7fdc47d3503d7fb17595142bb2c48dd5801cc4ad0517afb1a55">+7/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>retry_middleware_test.go</strong><dd><code>Enhance retry middleware body handling tests</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/networking/middleware/retry_middleware_test.go

<ul><li>Introduced a new <code>getBodyRetryRoundTripper</code> test helper.<br> <li> Added a test case to verify <code>http.NoBody</code> is not buffered when retries <br>are enabled.<br> <li> Added a test case to confirm that buffered request bodies correctly <br>set <code>GetBody</code> for HTTP/2 compatibility.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/622/changes#diff-781d13f3ad7692fdffe97f224e749e9e4cbbab29fbefafcfb11f78de8772aa26">+72/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/pkg/networking/logging.go
+++ b/pkg/networking/logging.go
@@ -37,16 +37,21 @@ func shouldNotLog(currentLevel zerolog.Level, levelToLogAt zerolog.Level) bool {
 }
 
 func getResponseBody(response *http.Response) io.ReadCloser {
-	if response.Body != nil {
-		bodyBytes, bodyErr := io.ReadAll(response.Body)
-		if bodyErr == nil {
-			response.Body.Close()
-			bodyReader := io.NopCloser(bytes.NewBuffer(bodyBytes))
-			response.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
-			return bodyReader
-		}
+	if response.Body == nil {
+		return nil
 	}
-	return nil
+
+	bodyBytes, readErr := io.ReadAll(response.Body)
+	closeErr := response.Body.Close()
+	if readErr != nil {
+		return nil
+	}
+
+	bodyReader := io.NopCloser(bytes.NewBuffer(bodyBytes))
+	response.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+	// closeErr is non-fatal once ReadAll succeeded; the body is fully buffered.
+	_ = closeErr
+	return bodyReader
 }
 
 func getRequestBody(request *http.Request) io.ReadCloser {
@@ -58,37 +63,44 @@ func getRequestBody(request *http.Request) io.ReadCloser {
 		return bodyReader
 	}
 
-	if request.Body != nil && request.Body != http.NoBody {
-		bodyBytes, bodyErr := io.ReadAll(request.Body)
-		if bodyErr == nil {
-			request.Body.Close()
-			bodyReader := io.NopCloser(bytes.NewBuffer(bodyBytes))
-			request.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
-			request.GetBody = func() (io.ReadCloser, error) {
-				return io.NopCloser(bytes.NewBuffer(bodyBytes)), nil
-			}
-			return bodyReader
-		}
+	if request.Body == nil || request.Body == http.NoBody {
+		return nil
 	}
 
-	return nil
+	bodyBytes, readErr := io.ReadAll(request.Body)
+	closeErr := request.Body.Close()
+	if readErr != nil {
+		return nil
+	}
+
+	bodyReader := io.NopCloser(bytes.NewBuffer(bodyBytes))
+	request.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+	request.GetBody = func() (io.ReadCloser, error) {
+		return io.NopCloser(bytes.NewBuffer(bodyBytes)), nil
+	}
+	// closeErr is non-fatal once ReadAll succeeded; Body/GetBody must stay restored.
+	_ = closeErr
+	return bodyReader
 }
 
 func decodeBody(bodyBytes []byte, contentEncoding string) (string, error) {
-	if contentEncoding == "gzip" {
-		reader, err := gzip.NewReader(bytes.NewReader(bodyBytes))
-		if err != nil {
-			return "", errors.Wrap(err, "failed to create gzip reader")
-		}
-		defer reader.Close()
-		decodedBytes, err := io.ReadAll(reader)
-		if err != nil {
-			return "", errors.Wrap(err, "failed to read decoded body")
-		}
-		return string(decodedBytes), nil
-	} else {
+	if contentEncoding != "gzip" {
 		return string(bodyBytes), nil
 	}
+
+	reader, err := gzip.NewReader(bytes.NewReader(bodyBytes))
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create gzip reader")
+	}
+
+	decodedBytes, readErr := io.ReadAll(reader)
+	closeErr := reader.Close()
+	if readErr != nil {
+		return "", errors.Wrap(readErr, "failed to read decoded body")
+	}
+	// closeErr is non-fatal once decoding succeeded; return content for logging.
+	_ = closeErr
+	return string(decodedBytes), nil
 }
 
 func logBody(logger *zerolog.Logger, logLevel zerolog.Level, logPrefix string, body io.ReadCloser, header http.Header, maxBodyCharacters int64) {
--- a/pkg/networking/logging.go
+++ b/pkg/networking/logging.go
@@ -36,7 +36,7 @@ func shouldNotLog(currentLevel zerolog.Level, levelToLogAt zerolog.Level) bool {
 	return currentLevel > levelToLogAt
 }
 
-func getResponseBody(response *http.Response) io.ReadCloser {
+func getResponseBody(response *http.Response, logger *zerolog.Logger) io.ReadCloser {
 	if response.Body == nil {
 		return nil
 	}
@@ -50,11 +50,13 @@ func getResponseBody(response *http.Response) io.ReadCloser {
 	bodyReader := io.NopCloser(bytes.NewBuffer(bodyBytes))
 	response.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
 	// closeErr is non-fatal once ReadAll succeeded; the body is fully buffered.
-	_ = closeErr
+	if closeErr != nil && logger != nil {
+		logger.WithLevel(defaultNetworkLogLevel).Err(closeErr).Msg("failed to close response body")
+	}
 	return bodyReader
 }
 
-func getRequestBody(request *http.Request) io.ReadCloser {
+func getRequestBody(request *http.Request, logger *zerolog.Logger) io.ReadCloser {
 	if request.GetBody != nil {
 		bodyReader, bodyErr := request.GetBody()
 		if bodyErr != nil {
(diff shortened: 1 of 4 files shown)


pkg/networking/middleware/retry_middleware_test.go
```diff
@@ -339,6 +339,57 @@ func TestNewRetryMiddleware(t *testing.T) {
...
+		require.Equal(t, http.NoBody, req.Body)
+
+		sut := NewRetryMiddleware(config, &logger, rt)
+		_, err = sut.RoundTrip(req)
+		require.NoError(t, err)
+
+		assert.Equal(t, http.NoBody, capturedBody)
+	})
+
+	t.Run("buffered request body sets GetBody for HTTP/2 retry compatibility", func(t *testing.T) {
```
bastiandoetsch, line 400: [Should Fix] This test is named for the feature but doesn't exercise it. The `roundTripFn` returns 200 on the first attempt, so `backoff.Retry` runs `op` exactly once and no retry happens; it then calls `capturedGetBody()` a single time. That proves `GetBody` was *assigned* and returns the buffer once — it does not prove the across-retry rewind contract HTTP/2 actually depends on (a *fresh, full* reader on the *second* invocation). A regression where `GetBody` returned an already-drained reader on the 2nd call would still pass. Note `failRoundtripper.RoundTrip` reads `req.Body`, never `req.GetBody`, so the existing multi-attempt tests don't cover this either.

Fix: drive an actual retry (return 503 then 200, or reuse `failRoundtripper` with `NumberOfAttemptsUntilSuccess`) and on each attempt call `req.GetBody()` + `io.ReadAll`, asserting the full expected body every time — modelling what the net/http transport does. The production closure is correct on inspection (it closes over an immutable buffer); this is purely about proving it.

— AI review

bastiandoetsch:
The **production code is correct** — `GetBody` is preserved in both the retry middleware and `getRequestBody` via a closure over an immutable buffer, `http.NoBody` is properly guarded (with genuine behaviour-pinning tests), and the read-error paths correctly avoid closing/mutating the body. Security scan clean (no new logging; the change doesn't widen what's emitted). The two Should-Fix items above are about test validity, not the code.

Non-blocking suggestions:
- [Suggestion] The 'buffer body → restore Body → set GetBody' logic is now duplicated in `logging.go:getRequestBody` and `retry_middleware.go:RoundTrip`; the 'must also restore GetBody' invariant is an unwritten rule copied by hand and will drift. Extract one `bufferRequestBody(req)` helper used by both.
- [Suggestion] `retry_middleware.go:138` reassigns `localRequest.GetBody` to the same closure already carried by `localRequest := *req` (set at line 126) — redundant; drop one.
- [Suggestion] `getRequestBody` overwrites any pre-existing custom `request.GetBody` with the in-memory buffer closure. This is the *correct* outcome once logging has buffered the body, but worth a comment noting logging takes ownership of the body representation.
- [Suggestion] `_ = closeErr` on `gzip.Reader.Close()` can mask a truncated/corrupt-trailer error; for the gzip case consider logging it at debug rather than discarding silently.
- [Suggestion · separate hardening] Not this PR, but noted in passing: full-header logging (Authorization/Cookie) at logging.go:160/178 is pre-existing and unredacted — worth a separate ticket.

— AI review

pkg/networking/logging_test.go
```diff
@@ -80,6 +81,152 @@ func Test_shouldNotLog(t *testing.T) {
...
+
+	bodyReader := getResponseBody(response)
+	assert.Nil(t, bodyReader)
+	assert.False(t, body.closed, "must not close response body when read fails")
+	assert.Same(t, body, response.Body)
+}
+
+func Test_getRequestBody_restoresBodyWhenCloseFails(t *testing.T) {
+	expectedBody := "hello world"
+	req, err := http.NewRequest(http.MethodPost, "http://localhost/", io.NopCloser(errOnCloseReadCloser{
```
bastiandoetsch: [Should Fix] This close-failure test never exercises the close-failure path. `errOnCloseReadCloser` is wrapped in `io.NopCloser`, which replaces Close with a no-op returning nil — so the injected `assert.AnError` is discarded and `getRequestBody` never sees a failing Close(). The test passes simply because the body is re-buffered; it would still pass if the production code aborted on a close error and failed to restore Body/GetBody, giving false confidence on exactly the behaviour it is named for. Same issue in `Test_getResponseBody_restoresBodyWhenCloseFails` (line 186).

Fix: pass `errOnCloseReadCloser{...}` directly (it already implements Close) without the `io.NopCloser` wrapper, so the erroring Close is actually invoked.

— AI review

pkg/networking/middleware/retry_middleware.go
```diff
@@ -109,7 +109,7 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	// if a body is available, create a local copy to be able to use it multiple times
-	if req.Body != nil && maxAttempts > 1 {
+	if req.Body != nil && req.Body != http.NoBody && maxAttempts > 1 {
```
danskmt, line 112: I think this `req.Body != nil` check is invalid as it sets a value for the Body before RoundTrip: `io.NopCloser`, like 125, wdyt?

pkg/networking/logging_test.go
robertolopezlopez: thanks @bastiandoetsch !

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: OK, after carefully checking this:

`req.Body != http.NoBody` is just a not empty check. Without it, we are replacing the body with a `NopCloser`... which we do not really want -> it was causing problems in request without body (GET). With the check, we just set `req.Body = io.NopCloser(...)` safely: GET request keep the `req.Body == http.NoBody` and do not set `GetBody`.

pkg/networking/middleware/retry_middleware_test.go
robertolopezlopez: adressed, I believe 👍

pkg/networking/logging_test.go
robertolopezlopez: addressed 👍

robertolopezlopez:
> The **production code is correct** — `GetBody` is preserved in both the retry middleware and `getRequestBody` via a closure over an immutable buffer, `http.NoBody` is properly guarded (with genuine behaviour-pinning tests), and the read-error paths correctly avoid closing/mutating the body. Security scan clean (no new logging; the change doesn't widen what's emitted). The two Should-Fix items above are about test validity, not the code.
> 
> Non-blocking suggestions:
> 
> * [Suggestion] The 'buffer body → restore Body → set GetBody' logic is now duplicated in `logging.go:getRequestBody` and `retry_middleware.go:RoundTrip`; the 'must also restore GetBody' invariant is an unwritten rule copied by hand and will drift. Extract one `bufferRequestBody(req)` helper used by both.
> * [Suggestion] `retry_middleware.go:138` reassigns `localRequest.GetBody` to the same closure already carried by `localRequest := *req` (set at line 126) — redundant; drop one.
> * [Suggestion] `getRequestBody` overwrites any pre-existing custom `request.GetBody` with the in-memory buffer closure. This is the _correct_ outcome once logging has buffered the body, but worth a comment noting logging takes ownership of the body representation.
> * [Suggestion] `_ = closeErr` on `gzip.Reader.Close()` can mask a truncated/corrupt-trailer error; for the gzip case consider logging it at debug rather than discarding silently.
> * [Suggestion · separate hardening] Not this PR, but noted in passing: full-header logging (Authorization/Cookie) at logging.go:160/178 is pre-existing and unredacted — worth a separate ticket.
> 
> — AI review

Thanks @bastiandoetsch 

- "Should fix" suggestions: addressed, I believe.
- I will create [a ticket](https://snyksec.atlassian.net/browse/CLI-1559) for the full header logging

robertolopezlopez:
> @robertolopezlopez could you also open a CLI PR to make sure there's nothing breaking from the CLI acceptance tests? I'll give it a test & review as well in a bit.

@CatalinSnyk https://github.com/snyk/cli/pull/6885

pkg/networking/logging_test.go
```diff
@@ -80,6 +81,156 @@ func Test_shouldNotLog(t *testing.T) {
...
+	err error
+}
+
+func (r *failReader) Read([]byte) (int, error) {
+	return 0, r.err
+}
+
+func (r *failReader) Close() error {
+	return nil
+}
```
danskmt: should we move this type and its functions to the top or bottom?

pkg/networking/logging_test.go
```diff
@@ -80,6 +81,156 @@ func Test_shouldNotLog(t *testing.T) {
...
+
+type trackingReadCloser struct {
+	io.Reader
+	closed bool
+}
+
+func (r *trackingReadCloser) Close() error {
+	r.closed = true
+	return nil
+}
```
danskmt: same for these, should be moved to the top or bottom?

pkg/networking/logging_test.go
robertolopezlopez: ok

pkg/networking/logging_test.go
robertolopezlopez: ok - I am moving everything to the bottom

robertolopezlopez: this was addressed as well

pkg/networking/logging_test.go
robertolopezlopez: and this

pkg/networking/logging.go
CatalinSnyk, line 1: suggestion: I do see the vision with the cleanup, but I think most of the changes are unrelated for the fix in `pkg/networking/middleware/retry_middleware.go`, so I would actually prefer we scope those out to a follow-up and keep the scope of the changes to minimum.

robertolopezlopez: cleaned up!

CatalinSnyk (approved):

PR #624, opened 2026-06-09
feat: make dot snyk excludes parameterisable

ancasnyk:
### Description

dotSnyk parsing had code hardcoded. In order to be able to reuse the functionality in the secrets extension, the Filter was made parameterizable with the .snyk exclude sections. 

It remains backwards compatible with the code sections as default.

### Checklist

- [x] Linted (`make lint`)

---

> [!NOTE]
> **Medium Risk**
> Changes which files are excluded when callers opt into different sections; wrong configuration could scan or skip paths unexpectedly, though defaults preserve existing behavior.
> 
> **Overview**
> **`FileFilter`** can now choose which `.snyk` `exclude` sections become ignore globs via **`WithDotSnykSections`**, so products like secrets can honor `secrets` (or custom section names) without hardcoding only `code` and `global`. Defaults stay **`code` + `global`** for backward compatibility; an empty slice turns off all `.snyk` exclusions.
> 
> Parsing uses a **map of YAML nodes** and **decodes each requested section separately**, so a bad section is logged and skipped instead of breaking other sections. **`parseIgnoreRuleToGlobs`** now uses **`slices.Contains`** for invalid-rule checks. Tests cover section selection, ordering, arbitrary section names, and malformed YAML.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8b3797d89e9426d6136ef9d783c121ef0adfb67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -50,9 +50,12 @@ func (s DotSnykExcludeSectionName) String() string {
 
 // DotSnykRule mirrors the relevant parts of a .snyk policy file. Exclude keys
 // its sections by name (e.g. "code", "global") so new sections are picked up by
-// decoding alone, without changes to this type.
+// decoding alone, without changes to this type. Section contents are kept as raw
+// yaml.Node values so each section can be decoded independently: only the
+// sections a caller opts into are decoded into rules, and a malformed section a
+// caller does not request cannot fail the decode of the whole file.
 type DotSnykRule struct {
-	Exclude map[DotSnykExcludeSectionName][]dotSnykExclude `yaml:"exclude"`
+	Exclude map[DotSnykExcludeSectionName]yaml.Node `yaml:"exclude"`
 }
 
 type FileFilterOption func(*FileFilter) error
@@ -215,10 +218,23 @@ func (fw *FileFilter) parseDotSnykFile(content []byte, filePath string) []string
 		return nil
 	}
 
-	// collect rules according to fw.dotSnykSections
+	// collect rules according to fw.dotSnykSections, decoding each requested
+	// section independently so a malformed section we don't care about (or a
+	// malformed sibling of one we do) cannot drop the sections we do apply.
 	var allRules []dotSnykExclude
 	for _, section := range fw.dotSnykSections {
-		allRules = append(allRules, rules.Exclude[section]...)
+		node, ok := rules.Exclude[section]
+		if !ok {
+			continue
+		}
+
+		var sectionRules []dotSnykExclude
+		if err := node.Decode(&sectionRules); err != nil {
+			fw.logger.Error().Msgf("parse .snyk section %q failed: %v", section, err)
+			continue
+		}
+
+		allRules = append(allRules, sectionRules...)
 	}
 
 	var globs []string
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
@@ -22,10 +23,50 @@ import (
 var defaultInvalidRules = []string{}
 
 type FileFilter struct {
-	path         string
-	defaultRules []string
-	logger       *zerolog.Logger
-	max_threads  int64
+	path            string
+	defaultRules    []string
+	logger          *zerolog.Logger
+	max_threads     int64
+	dotSnykSections []DotSnykExcludeSectionName
+}
+
+type DotSnykExcludeSectionName string
+
+const (
+	Global  DotSnykExcludeSectionName = "global"
+	Code    DotSnykExcludeSectionName = "code"
+	Secrets DotSnykExcludeSectionName = "secrets"
+)
+
+// dotSnykExcludeSectionNames lists every valid DotSnykExcludeSectionName value.
+var dotSnykExcludeSectionNames = []DotSnykExcludeSectionName{Global, Code, Secrets}
+
+// String implements fmt.Stringer.
+func (s DotSnykExcludeSectionName) String() string {
+	return string(s)
+}
+
+// IsValid reports whether s is one of the defined section names.
+func (s DotSnykExcludeSectionName) IsValid() bool {
+	return slices.Contains(dotSnykExcludeSectionNames, s)
+}
+
+// ParseDotSnykExcludeSectionName converts a string into a DotSnykExcludeSectionName,
+// returning an error if it does not match a defined section name.
+func ParseDotSnykExcludeSectionName(value string) (DotSnykExcludeSectionName, error) {
+	name := DotSnykExcludeSectionName(value)
+	if !name.IsValid() {
+		return "", fmt.Errorf("invalid .snyk exclude section name %q, must be one of %v", value, dotSnykExcludeSectionNames)
+	}
+	return name, nil
+}
+
+type DotSnykRule struct {
+	Exclude struct {
+		Code    []dotSnykExclude `yaml:"code"`
+		Global  []dotSnykExclude `yaml:"global"`
+		Secrets []dotSnykExclude `yaml:"secrets"`
+	} `yaml:"exclude"`
 }
 
 type FileFilterOption func(*FileFilter) error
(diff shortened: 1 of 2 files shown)


pkg/utils/file_filter.go
```diff
@@ -39,12 +80,24 @@ func WithThreadNumber(maxThreadCount int) FileFilterOption {
...
 	filter := &FileFilter{
-		path:         path,
-		defaultRules: []string{"**/.git/**"},
-		logger:       logger,
-		max_threads:  int64(runtime.NumCPU()),
+		path:            path,
+		defaultRules:    []string{"**/.git/**"},
+		logger:          logger,
+		max_threads:     int64(runtime.NumCPU()),
+		dotSnykSections: []DotSnykExcludeSectionName{Global, Code}, // init default with Global and Code to keep it backwards compatible
```
CatalinSnyk: nit: The previous order seems to be Code and then Global, not sure if there's any concrete cases for this being a problem, but seems like a trivial fix 🤷 

```go
allRules := append(rules.Exclude.Code, rules.Exclude.Global...)
```

pkg/utils/file_filter.go
CatalinSnyk, line 1: suggestion: I think the only thing this might be missing are some test cases 😄

pkg/utils/file_filter.go
```diff
@@ -20,10 +21,50 @@ import (
...
+	defaultRules    []string
+	logger          *zerolog.Logger
+	max_threads     int64
+	dotSnykSections []DotSnykExcludeSectionName
+}
+
+type DotSnykExcludeSectionName string
+
+const (
+	Global  DotSnykExcludeSectionName = "global"
```
CatalinSnyk: thought (non-blocking): Adding a future section would still need a code change here, and functionally this would basically just register another key in the exclude map. And the validation on top of it - which I guess would need some thoughts if it's really needed or not.

Decoding exclude into a `map[string][]dotSnykExclude` and just collecting whichever keys we're configured with would drop the struct fields, the switch, and arguably the validation. This would basically allow extension to register sections without a code change here.

Not a blocker for this PR, but it we wanted to flag this since it would likely be a good thing to support.

pkg/utils/file_filter.go
ancasnyk: I held off on tests until we decided on the approach. Added now :)

pkg/utils/file_filter.go
ancasnyk: reversed them, although it shouldn't make a difference

pkg/utils/file_filter.go
```diff
@@ -22,10 +23,36 @@ import (
...
+// String implements fmt.Stringer.
+func (s DotSnykExcludeSectionName) String() string {
+	return string(s)
+}
+
+// DotSnykRule mirrors the relevant parts of a .snyk policy file. Exclude keys
+// its sections by name (e.g. "code", "global") so new sections are picked up by
+// decoding alone, without changes to this type.
+type DotSnykRule struct {
+	Exclude map[DotSnykExcludeSectionName][]dotSnykExclude `yaml:"exclude"`
```
CatalinSnyk: small issue: Some fault tolerance finding my little AI assistant found: an unexpected section under `exclude:` silently drops all `.snyk` exclusions.

Since Exclude is a `map[...][]dotSnykExclude`, yaml decodes every section before we pick the ones we use. If any of the sections isn't a list, the `yaml.Unmarshal(content, &rules)` call would error => `parseDotSnykFile` bails to nil, and we lost the other sections that we would parse normally, since `GetRules` still returns no error. 

The old version simply didn't care about the other sections, and only parsed and returned the known ones.

CatalinSnyk: Test case used:
```go
func TestFileFilter_GetRules_malformedSectionDoesntWipeOthers(t *testing.T) {
	tempDir := t.TempDir()
	codeFile := "code.ts"
	createFileInPath(t, filepath.Join(tempDir, codeFile), []byte{})

	// `global` is malformed (missing the "- " list marker), a common typo
	snykContent := fmt.Sprintf(`exclude:
  code:
    - %s
  global:
    some-other-file.ts
`, codeFile)
	createFileInPath(t, filepath.Join(tempDir, ".snyk"), []byte(snykContent))

	fileFilter := NewFileFilter(tempDir, &log.Logger)
	actualRules, err := fileFilter.GetRules([]string{".snyk"})
	assert.NoError(t, err)

	// the valid `code` exclusion should still be applied
	expectedRules := append([]string{}, fileFilter.defaultRules...)
	expectedRules = append(expectedRules,
		fmt.Sprintf("%s/**/%s/**", filepath.ToSlash(tempDir), codeFile),
		fmt.Sprintf("%s/**/%s", filepath.ToSlash(tempDir), codeFile),
	)
	assert.ElementsMatch(t, expectedRules, actualRules)
}
```

CatalinSnyk: And suggested fix is to manually decode the yaml nodes (maybe there's better ones 🤷):
```go
Exclude map[DotSnykExcludeSectionName]yaml.Node `yaml:"exclude"`
```

```go
	var allRules []dotSnykExclude
	for _, section := range fw.dotSnykSections {
		node, ok := rules.Exclude[section]
		if !ok {
			continue
		}

		var sectionRules []dotSnykExclude
		if err := node.Decode(&sectionRules); err != nil {
			fw.logger.Error().Msgf("parse .snyk section %q failed: %v", section, err)
			continue
		}
		allRules = append(allRules, sectionRules...)
	}
```

pkg/utils/file_filter.go
```diff
@@ -30,43 +30,29 @@ type FileFilter struct {
...
+// file (e.g. "code", "global", "secrets", "iac"). It is a plain string so callers can
+// opt into sections this package doesn't define constants for without requiring
+// a code change here.
 type DotSnykExcludeSectionName string
 
 const (
 	Global  DotSnykExcludeSectionName = "global"
 	Code    DotSnykExcludeSectionName = "code"
 	Secrets DotSnykExcludeSectionName = "secrets"
+	Iac     DotSnykExcludeSectionName = "iac"
```
CatalinSnyk: question: I think this was not in the initial discussion scope, ~~is this supported?~~ should this be included?

CatalinSnyk: also this addition kinda defeats the purpose of this test case, since it's defined as a constant, just not referenced as one 😅 :
```go
	{
			// no code change in file_filter.go is needed to support a new section
			name:          "supports sections without a predefined constant",
			options:       []FileFilterOption{WithDotSnykSections([]DotSnykExcludeSectionName{"iac"})},
			expectedFiles: []string{customFile},
		},
```

ancasnyk: you're right, it's actually iac-drift https://github.com/snyk/policy/blob/master/lib/add-exclude.ts#L42 
"iac" is only for ignores

the test was changed when the iac constant was added  🙏 
[
[]FileFilterOption{WithDotSnykSections([]DotSnykExcludeSectionName{"newsection"})},](https://github.com/snyk/go-application-framework/pull/624/changes#diff-0525df151b63831e042ffd8a2c7cab466a29f8d6f710cb726c359b6becffa89cR275)

pkg/utils/file_filter.go
```diff
@@ -22,10 +23,39 @@ import (
...
+}
+
+// DotSnykExcludeSectionName is the name of an `exclude` section in a .snyk
+// file (e.g. "code", "global", "secrets", "iac"). It is a plain string so callers can
+// opt into sections this package doesn't define constants for without requiring
+// a code change here.
+type DotSnykExcludeSectionName string
+
+const (
+	Global  DotSnykExcludeSectionName = "global"
```
CatalinSnyk: super small nit: Not sure the names are quite fitting, current usage for the consumers would be `utils.Code`, `utils.Secrets`, etc. I think it's maybe a bit too generic, so perhaps we need some better names for those....

CatalinSnyk (approved):
Pre-approving, please feel free to merge after the CLI PR is raised 😄

pkg/utils/file_filter.go
ancasnyk: changed to something more expressive

PR #625, opened 2026-06-10
feat: support new HTML mime type for HTML rendering

j-luong:
### Description

This PR adds support for a new mime type, `text/html` in our presentation layer. This adds the initial changes necessary for HTML rendering and includes
- a new function map implementation that supports Go's `html/template` package
- an initial `ufm.html.tmpl` template that renders a simple HTML output using the new function map

#626 should be merged into this PR before this PR can be merged into main

### Checklist

- [ ] Linted (`make lint`)

### DIFF
--- a/internal/presenters/funcs.go
+++ b/internal/presenters/funcs.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	htmlTemplate "html/template"
 	"maps"
 	"os"
 	"reflect"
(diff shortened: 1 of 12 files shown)


CatalinSnyk (approved):

internal/presenters/funcs.go
robertolopezlopez, line 1: I have the feeling that this file has become a jumble mumble, kind of `utils.go`. It would be worth to split into separate files with meaningful purposes (i.e. `html-template-utils.go`)

robertolopezlopez (commented):
I do not see any major issue in this PR, other than the inherited [`tools.go`, `other_tools.go`, ...] pattern which is a strong candidate for refactoring (but out of scope this time)

internal/presenters/funcs.go
j-luong: yes I think so...it can be part of the next task to add the html template

PR #629, opened 2026-06-16
fix: Buffer request body for retry overrides, close intermediate response bodies

danskmt:
### Description

Fixes empty request bodies when retrying POST requests. In addition it addresses several connection/resource management gaps in the retry middleware.

#### 1. Always buffer request body for retry overrides

POST request bodies were not preserved across retry attempts when a 429 triggered the per-status-code retry override (3 attempts despite `maxAttempts=1`).

**Root cause:** Body buffering was gated on `maxAttempts > 1`, but the per-status-code override in `getMaxRetryAttempts` is only evaluated after the first response — by which point the body was already consumed and never buffered.

**Fix:** `ensureGetBodyExists` now always prepares a replayable body using three strategies in order of preference:
1. `GetBody` already set (e.g. `http.NewRequest` with `*bytes.Reader`) → use as-is, no copy.
2. Body implements `io.ReadSeeker` (e.g. `*os.File`) → wrap to suppress Close, rewind with `Seek(0)`.
3. Fallback → `io.ReadAll` into memory buffer.

#### 2. Close intermediate response bodies between retries

The middleware returned `(response, retryErr)` on retriable status codes, but `backoff.Retry` discarded intermediate responses without closing their bodies — leaking one TCP connection per retry. This matches the drain+close behavior in stdlib's `Client.do`.

`drainAndClose` is called eagerly when the retry decision is made (non-permanent errors). On permanent 503 where `getErrorList` replaces `response.Body` with a buffer copy, the orphaned original transport body is also closed.

#### 3. Cleanup: remove dead code

Removed redundant `noCloseSeekBody.Read` method — the embedded `io.ReadSeeker` already promotes `Read`.

**Ticket:** [CLI-1591](https://snyksec.atlassian.net/browse/CLI-1591)

### Test coverage

| Scenario | Test |
|----------|------|
| POST body preserved across 429 retries | `TestRetryMiddleware_429_POST_BodyPreservedAcrossRetries` |
| GetBody preset skips buffering | `TestRetryMiddleware_GetBodyPreset_SkipsBuffering` |
| NopCloser body falls back to ReadAll | `TestRetryMiddleware_GetBodyNil_BuffersBody` |
| Seekable body rewinds without copy | `TestRetryMiddleware_SeekableBody_NoBuffer` |
| Intermediate response body closed (1 retry) | `TestRetryMiddleware_IntermediateResponseBodyClosed` |
| All intermediate bodies closed (3 retries) | `TestRetryMiddleware_MultipleIntermediateResponseBodiesClosed` |
| Context cancellation still closes body | `TestRetryMiddleware_ContextCancellation_BodyClosed` |
| Permanent 503 closes orphaned transport body | `TestRetryMiddleware_503Permanent_OriginalBodyClosed` |

### Checklist

- [x] Linted (`make lint`)

PR in CLI: https://github.com/snyk/cli/pull/6912

[CLI-1591]: https://snyksec.atlassian.net/browse/CLI-1591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -240,14 +240,16 @@ func ensureGetBodyExists(req *http.Request) (io.ReadCloser, func() (io.ReadClose
 		return req.Body, req.GetBody, nil, nil
 	}
 
+	// body can be rewound
 	if rs, ok := req.Body.(io.ReadSeeker); ok {
-		wrapped := &noCloseSeekBody{ReadSeeker: rs, realCloser: req.Body}
-		return wrapped, func() (io.ReadCloser, error) {
+		body := &noCloseSeekBody{ReadSeeker: rs, realCloser: req.Body}
+		getBody := func() (io.ReadCloser, error) {
 			if _, seekErr := rs.Seek(0, io.SeekStart); seekErr != nil {
 				return nil, seekErr
 			}
-			return wrapped, nil
-		}, wrapped.RealClose, nil
+			return body, nil
+		}
+		return body, getBody, body.RealClose, nil
 	}
 
 	bodyBytes, readErr := io.ReadAll(req.Body)
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -79,7 +79,6 @@ type noCloseSeekBody struct {
 	realCloser io.Closer
 }
 
-func (b *noCloseSeekBody) Read(p []byte) (int, error) { return b.ReadSeeker.Read(p) }
 func (b *noCloseSeekBody) Close() error               { return nil }
 func (b *noCloseSeekBody) RealClose() error {
 	if b.realCloser != nil {
@@ -115,6 +114,7 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	var retryAfterSeconds = defaultRetryAfterSeconds
 	var actualAttempts int = 0
 	var cachedMaxRetries *int = nil // Per-request cached max retries
+	var prevRespBody io.ReadCloser  // original response body from prior attempt, for drain+close
 
 	if tmp := rm.config.GetInt(ConfigurationKeyRequestAttempts); tmp > 0 {
 		maxAttempts = tmp
@@ -140,6 +140,15 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	op := func() (*http.Response, error) {
 		actualAttempts++
+
+		// Close the previous response body before retrying to avoid leaking
+		// connections.  This matches stdlib's Client.do which drains+closes
+		// intermediate responses between attempts.
+		if prevRespBody != nil {
+			_, _ = io.Copy(io.Discard, prevRespBody)
+			_ = prevRespBody.Close()
+			prevRespBody = nil
+		}
 
 		// create a local copy of the request
 		localRequest := *req
@@ -177,8 +186,10 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 		}
 
 		// depending on the response determine if we should retry
+		origBody := response.Body // capture before shouldRetry may replace it (e.g. getErrorList for 503)
 		if retryError := shouldRetry(response, actualAttempts, *cachedMaxRetries); retryError != nil {
 			rm.logger.Debug().Msgf("Retrying request, reason: %v", retryError)
+			prevRespBody = origBody
 			return response, &RetryAttemptError{
 				StatusCode:  response.StatusCode,
 				Attempt:     actualAttempts,
@@ -199,6 +210,13 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 			rm.notifyRetry(reqCtx, err, duration)
 		}),
 	)
+
+	// Ensure any intermediate body is closed if the loop exited before the
+	// next op() call could clean it up (e.g. context cancellation).
+	if prevRespBody != nil && (finalResponse == nil || prevRespBody != finalResponse.Body) {
+		_, _ = io.Copy(io.Discard, prevRespBody)
+		_ = prevRespBody.Close()
+	}
 
 	finalError = rm.filterRetryError(finalError, actualAttempts)
 	return finalResponse, finalError
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -108,20 +108,13 @@ func NewRetryMiddleware(config configuration.Configuration, logger *zerolog.Logg
 	return rm
 }
 
-func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) { //nolint:gocyclo // complexity inherent to sequential retry logic with per-status-code overrides
+func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 	var finalResponse *http.Response
 	var finalError error
-	var localBodyBuffer []byte
 	var maxAttempts = defaultMaxAttemptsCount
 	var retryAfterSeconds = defaultRetryAfterSeconds
 	var actualAttempts int = 0
 	var cachedMaxRetries *int = nil // Per-request cached max retries
-	var deferredBodyClose func() error
-	defer func() {
-		if deferredBodyClose != nil {
-			_ = deferredBodyClose()
-		}
-	}()
 
 	if tmp := rm.config.GetInt(ConfigurationKeyRequestAttempts); tmp > 0 {
 		maxAttempts = tmp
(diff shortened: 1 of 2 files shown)


pkg/networking/middleware/retry_middleware_test.go
```diff
@@ -837,6 +837,66 @@ func TestRetryAttemptNotification(t *testing.T) {
...
+	response, err := sut.RoundTrip(req)
+
+	assert.NoError(t, err)
+	require.NotNil(t, response)
+	assert.Equal(t, http.StatusOK, response.StatusCode, "Should succeed after retries")
+	assert.Equal(t, 3, attemptCount, "Should make 3 attempts per 429 override")
+
+	require.Len(t, bodiesReceived, 3, "Should have recorded body for each attempt")
+	for i, body := range bodiesReceived {
+		assert.Equal(t, expectedBody, body,
```
robertolopezlopez, line 905: it is good that you are verifying the bodies across retries, but also you may verify `req.GetBody()` (it should give you the same payload)

not a blocker, but nice to have

robertolopezlopez (approved):

pkg/networking/middleware/retry_middleware.go
```diff
@@ -108,8 +108,10 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 		retryAfterSeconds = tmp
 	}
 
-	// if a body is available, create a local copy to be able to use it multiple times
-	if req.Body != nil && req.Body != http.NoBody && maxAttempts > 1 {
+	// if a body is available, create a local copy to be able to use it multiple times.
+	// Always buffer: per-status-code overrides (e.g. 429 → 3 attempts) can trigger retries
+	// regardless of the configured maxAttempts, so the body must be available for reuse.
+	if req.Body != nil && req.Body != http.NoBody {
```
PeterSchafer: Question: Did we evaluate the memory impact of this? Or are we planning to do so?

danskmt: I'll send a message in Slack about this

pkg/networking/middleware/retry_middleware.go
```diff
@@ -138,6 +144,18 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+			// On retry attempts, obtain a fresh body via GetBody.
+			// The first attempt uses the original req.Body so the transport
+			// can read and close it per the http.RoundTripper contract.
+			body, err := req.GetBody()
+			if err != nil {
+				return nil, backoff.Permanent(err)
+			}
+			if body == nil {
+				return nil, backoff.Permanent(fmt.Errorf("GetBody returned nil reader"))
+			}
```
danskmt: @PeterSchafer I've added this extra check FYI

pkg/networking/middleware/retry_middleware.go
```diff
@@ -187,6 +204,52 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	}
+	if closeErr != nil {
+		return nil, nil, nil, closeErr
+	}
+
+	newGetBody := func() (io.ReadCloser, error) {
+		return io.NopCloser(bytes.NewBuffer(bodyBytes)), nil
+	}
+	firstBody, firstBodyErr := newGetBody()
+	return firstBody, newGetBody, nil, firstBodyErr
```
danskmt, line 268: I'm wondering if we should put the:
```go
closeErr := req.Body.Close()
if closeErr != nil {
		return nil, nil, nil, closeErr
}
```

in a **defer** func instead of how it is right now here

pkg/networking/middleware/retry_middleware.go
```diff
@@ -187,6 +204,52 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+//  2. Body implements io.ReadSeeker → wrap to suppress Close, Seek(0) to rewind.
+//  3. Fallback: io.ReadAll into memory buffer.
+func ensureGetBodyExists(req *http.Request) (io.ReadCloser, func() (io.ReadCloser, error), func() error, error) {
+	if req.Body == nil || req.Body == http.NoBody {
+		return req.Body, nil, nil, nil
+	}
+
+	if req.GetBody != nil {
+		return req.Body, req.GetBody, nil, nil
+	}
```
danskmt, line 241: In the first scenario, should we consider providing **req.GetBody** if present as well? i.e.
```go
if req.Body == nil || req.Body == http.NoBody {
        if req.GetBody != nil {
		   return req.Body, req.GetBody, nil, nil
	    }
		return req.Body, nil, nil, nil
}

if req.GetBody != nil {
		return req.Body, req.GetBody, nil, nil
}
```

PeterSchafer: Since this case is the case where there is no body, it doesn't seem required to return a GetBody

pkg/networking/middleware/retry_middleware.go
PeterSchafer: I think it would be optional and might make things even a bit more complicated

pkg/networking/middleware/retry_middleware.go
```diff
@@ -71,6 +71,23 @@ type RetryMiddleware struct {
...
 
+// noCloseSeekBody wraps an io.ReadSeeker so that Close is suppressed during
+// retries.  The underlying stream stays open and is rewound with Seek(0)
+// instead of being copied into memory.  Call RealClose after the retry loop.
+type noCloseSeekBody struct {
+	io.ReadSeeker
+	realCloser io.Closer
+}
+
+func (b *noCloseSeekBody) Read(p []byte) (int, error) { return b.ReadSeeker.Read(p) }
```
PeterSchafer: This can probably be removed, but please verify.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -108,49 +133,50 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
 		if response != nil && response.Header != nil && actualAttempts > 1 {
 			response.Header.Set(retryCountHeaderKey, fmt.Sprintf("%d", actualAttempts))
 		}
 
 		// errors from the next round tripper cannot be retried
-		if err != nil {
-			return response, backoff.Permanent(err)
+		if rtErr != nil {
+			return response, backoff.Permanent(rtErr)
 		}
```
robertolopezlopez, line 180: I am wondering about the order here:

1. Roundtrip
2. set header
3. if error: return response, error

Do you really need to set the header in case of error? Just thinking that the canonical order is usually:

1. Roundtrip
2. if error: return response, error
3. set header

But probably I am just missing the point here. Do you mind to clarify?

pkg/networking/middleware/retry_middleware.go
```diff
@@ -187,6 +222,52 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	}
+	if closeErr != nil {
+		return nil, nil, nil, closeErr
+	}
+
+	newGetBody := func() (io.ReadCloser, error) {
+		return io.NopCloser(bytes.NewBuffer(bodyBytes)), nil
+	}
+	firstBody, firstBodyErr := newGetBody()
+	return firstBody, newGetBody, nil, firstBodyErr
```
robertolopezlopez, line 268: can you maybe return `struct, error` instead of all those `firstBody, newGetBody, error, error`

pkg/networking/middleware/retry_middleware.go
danskmt: Yeah I believe you're right @robertolopezlopez , WDYT @PeterSchafer ?
It is not harmful but we can do this change

pkg/networking/middleware/retry_middleware.go
danskmt: Yeah we can probably consider that. Btw, there is one scenario where it is not nil:
**wrapped.RealClose**
image


Just a little bit hard to read the code being like this

robertolopezlopez: yes I saw it after my initial message ;-)

image

pkg/networking/middleware/retry_middleware.go
danskmt: @robertolopezlopez I believe here is that we want to preserve observability regarding num of attempts so we can leave as is

pkg/networking/middleware/retry_middleware.go
danskmt: I'm thinking of considering this type of cleanup for followup task

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: fair enough, sounds good to me

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: sounds good to me

PeterSchafer (approved):

PR #632, opened 2026-06-19
feat: summarize information from debug logs

CatalinSnyk:
### Description

- Adding some basic matching for notable/useful information from debug logs.
image

### Checklist

- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI - [PR](https://github.com/snyk/cli/pull/6913)

### DIFF
--- a/pkg/local_workflows/doctor_workflow/logsummary/format.go
+++ b/pkg/local_workflows/doctor_workflow/logsummary/format.go
@@ -1,0 +1,41 @@
+package logsummary
+
+import (
+	"fmt"
+	"strings"
+)
+
+func (summary Summary) Format() string {
+	var sb strings.Builder
+
+	sb.WriteString("Snyk Doctor Diagnostic Report\n")
+
+	writeSection(&sb, "Environment", summary.Header)
+
+	sb.WriteString("\n\nNotable Events\n\n")
+	if len(summary.Highlights) == 0 {
+		sb.WriteString("  No failing requests or CLI error entries found in the log body.\n")
+	} else {
+		for _, h := range summary.Highlights {
+			fmt.Fprintf(&sb, "  L%d [%s] %s\n", h.Line, h.Kind, h.Message)
+		}
+		if summary.Truncated {
+			fmt.Fprintf(&sb, "  (showing first %d notable events; remaining log not scanned)\n", maxHighlights)
+		}
+	}
+
+	writeSection(&sb, "Result", summary.Footer)
+
+	return sb.String()
+}
+
+func writeSection(sb *strings.Builder, title, content string) {
+	fmt.Fprintf(sb, "\n\n%s\n\n", title)
+	if content == "" {
+		sb.WriteString("  (not found in the provided log)\n")
+		return
+	}
+	for line := range strings.SplitSeq(content, "\n") {
+		fmt.Fprintf(sb, "  %s\n", strings.TrimRight(line, " "))
+	}
+}
--- a/pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
+++ b/pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
@@ -1,0 +1,213 @@
+package logsummary
+
+import (
+	"regexp"
+	"strings"
+)
+
+type EventKind string
+
+const (
+	EventHTTPError EventKind = "http-error"
+	EventError     EventKind = "error"
+)
+
+const (
+	maxHighlights  = 200
+	summaryMarker  = "------------ Summary ------------"
+	errorsMarker   = "------------ Errors ------------"
+	versionPrefix  = "Version:"
+	exitCodePrefix = "Exit Code:"
+	cliErrorPrefix = "< error:"
+	failedPrefix   = "Failed "
+)
+
+var (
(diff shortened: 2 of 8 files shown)


pkg/local_workflows/doctor_workflow/logsummary/format.go
```diff
@@ -0,0 +1,41 @@
...
+
+	writeSection(&sb, "Environment", summary.Header, "not found in the provided log")
+
+	fmt.Fprintf(&sb, "\n\nNotable Events\n\n")
+	if len(summary.Highlights) == 0 {
+		sb.WriteString("  No failing requests or error/warn entries found in the log body.\n")
+	} else {
+		for _, h := range summary.Highlights {
+			fmt.Fprintf(&sb, "  L%d [%s] %s\n", h.Line, h.Kind, h.Message)
+		}
```
danskmt, line 21: is there any scenario where Line, Kind or Message would be empty / not present?

pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if start < 0 {
+		return span{-1, -1}
+	}
+	return span{start, end}
+}
+
+func isTableRow(msg string) bool {
+	return msg != "" && (msg[0] == ' ' || msg[0] == '\t' || tableRowRe.MatchString(msg))
+}
```
danskmt: this seems very hacky, not sure if there are better ways of doing this

pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if start < 0 {
+		return span{-1, -1}
+	}
+	end := start
+	for i := start + 1; i < len(msgs) && isTableRow(msgs[i]); i++ {
+		end = i
+	}
+	return span{start, end}
+}
```
danskmt: I'm thinking if is possible to work with templates instead of doing these manipulations... Have you considered those / had impediments regarding them?

danskmt:
@CatalinSnyk also, bot left some comments that might be relevant:
image

pkg/local_workflows/doctor_workflow/logsummary/format.go
CatalinSnyk: I think not, Line is always i+1 ≥ 1; Kind is always one of the four constants set by classify; and classify("") returns false, so empty/whitespace messages never become highlights.

pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
CatalinSnyk: Agreed, I'll follow-up on it

pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
CatalinSnyk: Not quite sure, what kind of templates did you have in mind?

Templates that we have in GAF (for our presenter) are better for rendering: if we used templates for rendering the CLI header/footer then I think it could be a bit more viable, but for matching the current log header/footer section I don't think it would make much sense.

pkg/local_workflows/doctor_workflow/logsummary/testdata/wrong-environment.logs
```diff
@@ -0,0 +1,24 @@
...
+2026-06-10T13:20:21Z main -
+2026-06-10T13:20:21Z main - ------------ Summary ------------
+2026-06-10T13:20:21Z main - Version:               1.0.0-test
+2026-06-10T13:20:21Z main - Platform:              linux amd64
+2026-06-10T13:20:21Z main - API:                   https://api.snyk.io
+2026-06-10T13:20:21Z main - Authorization:         tok11111***tok99999  (type=token)
+2026-06-10T13:20:21Z main - Checks:
+2026-06-10T13:20:21Z main -   Configuration:       all good
+2026-06-10T13:20:21Z main -
+2026-06-10T13:20:21Z main - Exit Code:             2
```
danskmt, line 24: we need to support multiple versions of outputs. that we can parse:
BASE_VERSION,
V1,
V2,
...

So a parsing strategy that can be versioned is more ideal for this scenario.
cc: @PeterSchafer @robertolopezlopez

danskmt: Identify by version

Image

pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
danskmt: it was more like to be focused on the structure, i.e. reading the code we can identify the structure more easily. Similar to using a separation between lexer and parser, this could improve visibility of the code and structure.

pkg/local_workflows/doctor_workflow/logsummary/testdata/wrong-environment.logs
robertolopezlopez: yeah indeed a resilient parsing strategy is better, to adapt to breaking changes in the version log lines

danskmt:
Closed in favor of https://github.com/snyk/go-application-framework/pull/643

PR #633, opened 2026-06-22
feat: add asset_name to TestConfiguration

oana-andrisoaei-snyk:
OSF-427

Propagates the new optional `asset_name` field (string, maxLength 256) on `TestConfiguration` introduced upstream in [snyk/test-api-shim#654](https://github.com/snyk/test-api-shim/pull/654) (DGP-1838) so consumers can pass an asset name through to the test-service request.

Because this is a full upstream sync, pulling the spec also brings in the other changes that have landed on `snyk/test-api-shim` main since the last refresh:

- new `monitor` field on `TestConfiguration` (`*bool`, indicates whether the test result should be monitored with recurring tests)
- `io.snyk.api.v1testdepgraph.request.DepGraph` is replaced by `DepGraphRef`, which uses `x-go-type: json.RawMessage` and so generates as `type DepGraphRef = json.RawMessage`
- updated descriptions on `ScmResource` (secrets scans no longer require the repo to be already imported into Snyk) and `SnykReferenceResource` (now documents supported SCA scan via Target ID)

### Checklist

- [x] Tests added and all succeed (`make test`) — `go test ./pkg/apiclients/testapi/...` passes; only the mock test helper signature changed (type rename to `DepGraphRef`).
- [x] Regenerated mocks, etc. (`make generate`) — `make update-local-findings` runs the spec pull and then `make generate`.
- [x] Linted (`make lint`) — `golangci-lint run ./...` reports 0 issues.
- [x] Test your changes work for the CLI — additive optional field with no current consumers yet; `DepGraphRef` alias is source-compatible with the prior `IoSnykApiV1testdepgraphRequestDepGraph` only at the JSON wire level, so any CLI code that referenced the named struct directly will need a follow-up.

Made with [Cursor](https://cursor.com)


CatalinSnyk (approved):

CatalinSnyk (approved):

pkg/apiclients/testapi/testapi.gen.go
```diff
@@ -618,13 +618,16 @@ type DeepcodeBundleSubjectType string
 // DepGraphContent Dep Graph file contents, of the same format as Dep Graphs
 // provided to /v1/test-dep-graph endpoint.
 type DepGraphContent struct {
-	DepGraph IoSnykApiV1testdepgraphRequestDepGraph `json:"dep_graph"`
-	Type     DepGraphContentType                    `json:"type"`
+	DepGraph DepGraphRef         `json:"dep_graph"`
```
PeterSchafer, line 621: Issue: This is a breaking change!

PR #634, opened 2026-06-22
feat: regenerate & document fixture files for UF presenter test suite

robertolopezlopez:
### Description

This PR regenerates and documents UFM presenter live fixtures.

*Fixture catalog, workflow*

- Add `internal/presenters/testdata/ufm/README.md` with per-fixture information.
- Link from `CONTRIBUTING.md`

*`generate-fixture` fix*
- Set `INTERNAL_SNYK_CLI_USE_UNIFIED_TEST_API_FOR_OS_CLI_TEST=true` for `test*` scans so OSS `snyk test` emits `workflow.TestResult` dumps (legacycli path does not).

Live fixture refresh (platform_hammerhead_testing)
- `testresult_cli` -  `snyk test .` on `snyk/cli`
- `webgoat` / `webgoat.ignore` - `snyk test .` on OWASP WebGoat Java (shared input; expected output differs by test config)
- Regenerated expected SARIF/human/HTML snapshots for affected presenter cases

*Live fixture refresh (`platform_hammerhead_testing`)*
- `testresult_cli` - `snyk test .` on `snyk/cli`
- `webgoat` / `webgoat.ignore` - `snyk test .` on OWASP WebGoat Java (shared input; expected output differs by test config)
- Regenerated expected SARIF/human/HTML snapshots for affected presenter cases

*multi_project*
-  Removed legacy `tpwe` (.NET) sub-project from committed `multi_project.testresult.json` and `multi_project.sarif.json` (personal project, no maintained source checkout).
- Regenerated human-readable expected output for the 6 remaining sub-projects.
- Not live-regenerated in this PR.

*Extra*

- `go-git` bump to 5.19.0 because of [this](https://app.circleci.com/pipelines/github/snyk/go-application-framework/4010/workflows/b8e65b3b-b5df-43f5-8c90-e40e2c33a736/jobs/17776)
  - `testify` bump to 1.11.1 is automatic after `go mod tidy`

### Checklist

- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI -> https://github.com/snyk/cli/pull/6945


pkg/utils/ufm/json_test_result_test.go
```diff
@@ -453,6 +453,11 @@ func Test_Findings_ConcurrentAccess(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, testResults, 1)
 
+	expectedFindings, complete, err := testResults[0].Findings(ctx)
```
basti-snyk: **Should Fix — this change defangs `Test_Findings_ConcurrentAccess`.** The new pre-loop call `testResults[0].Findings(ctx)` performs the lazy `ReconstructFindings` once, single-threaded, and memoizes into `fullFindings`. All 10 goroutines below then hit the `fullFindings != nil` fast-return and never exercise the concurrent first-access reconstruction path — which is the exact code this test exists to guard. A future race introduced in lazy reconstruction would now pass clean under `-race`.

Two fixes, ideally both:
1. Derive `expectedCount` without warming the object under test — e.g. parse the bytes into a *separate* `NewSerializableTestResultFromBytes` instance (or count `FindingsData`/`ProblemRefs`), so `testResults[0]` is still cold when the goroutines first hit it.
2. The old `require.Len(..., 47)` also guaranteed a non-trivial payload; the dynamic `expectedCount` passes vacuously if the fixture ever regenerates to 0 findings. Add `require.Greater(t, expectedCount, 0)` (or restore an absolute count, e.g. `55`).

— AI review

robertolopezlopez: addressed with latest commit - thanks!

internal/presenters/testdata/ufm/cli.human.readable
```diff
@@ -1,128 +1,109 @@
-[1mTesting  (package.json) ...[0m
+[1mTesting  (package-lock.json) ...[0m
```
basti-snyk, line 1: **Should Fix (test coverage) — reachability rendering is no longer exercised by any fixture.** Regenerating these inputs via the OSS `snyk test` path dropped all reachability data: `testresult_cli.json` went 26→0 reachability entries and `webgoat.ignore.testresult.json` 47→0, so every `Reachability: …` line disappeared from the golden output (`cli.human.readable` 9→0, `webgoat.ignore.human.readable` 42→0). The UFM presenter still renders reachability, but after this PR **no fixture covers that branch** — a future regression that breaks reachability rendering would pass the whole UFM golden suite clean.

This is new since my last review (the cli fixture also switched `package.json`→`package-lock.json`, which is what dropped the reachability-bearing input).

Suggest either retaining/adding one fixture that carries reachability data so the branch stays covered, or documenting in the README that reachability output is intentionally untested post-CLI-1510 with a follow-up ticket.

— AI review

internal/presenters/testdata/ufm/README.md
```diff
@@ -0,0 +1,142 @@
...
+make generate-fixture \
+  PROJECT=/path/to/snyk/cli \
+  ORG="$ORG" \
+  NAME=testresult_cli \
+  SNYK_BIN="$SNYK_BIN" \
+  SCAN_CMD="test ."
+cp dumps/testresult_cli.testresult.json internal/presenters/testdata/ufm/testresult_cli.json
+```
+
+Verify: `metadata.project-name` is `snyk`, `display-target-file` is `package.json`.
```
basti-snyk: **Suggestion (doc/fixture drift) — `testresult_cli` target-file claim contradicts the regenerated fixture.** This line (and the catalog row at the top) say `display-target-file` is `package.json`, but the regenerated `cli.human.readable` now reports `Testing (package-lock.json)` — the new commits switched the cli scan to the lockfile. Update the README verify step to `package-lock.json` so a contributor regenerating doesn't think the run drifted.

— AI review

scripts/generate-fixture.sh
```diff
@@ -16,6 +16,19 @@ SCAN_CMD="${SCAN_CMD:-${DEFAULT_SCAN_CMD}}"
...
+
+is_oss_test_scan() {
+	[[ "${scan_args[0]:-}" == "test" ]]
+}
+
+# OSS `snyk test` routes to legacycli unless the unified Test API FF is on; only the
+# unified path emits workflow.TestResult payloads we need. Config honors this env var
+# before the org FF gateway (see CLI-1510). Not needed for `secrets test`.
+if is_oss_test_scan; then
+  export INTERNAL_SNYK_CLI_USE_UNIFIED_TEST_API_FOR_OS_CLI_TEST="${INTERNAL_SNYK_CLI_USE_UNIFIED_TEST_API_FOR_OS_CLI_TEST:-true}"
```
basti-snyk: Suggestion (non-blocking): this `export` runs before the `PROJECT`/`ORG`/`NAME` validation block at lines 32-42. If the script is *sourced* (`. ./generate-fixture.sh`) rather than executed, and validation exits early, `INTERNAL_SNYK_CLI_USE_UNIFIED_TEST_API_FOR_OS_CLI_TEST=true` leaks into the caller's interactive shell and silently affects later manual `snyk` runs. Moving the `export` to after the validation block (it's only needed at scan time) avoids the leak. In the normal executed-as-subprocess path this is harmless. — AI review

pkg/utils/ufm/json_test_result_test.go
```diff
@@ -453,6 +453,15 @@ func Test_Findings_ConcurrentAccess(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, testResults, 1)
 
+	expectedResults, err := NewSerializableTestResultFromBytes(testResultsData)
```
CatalinSnyk: issue: I think I understand what the intention of this was, but it's asserting the same thing against each other 😅 since just above testResults is going the same function call
```go
	testResults, err := NewSerializableTestResultFromBytes(testResultsData)
```

robertolopezlopez: xD

scripts/generate-fixture.sh
```diff
@@ -69,24 +85,47 @@ fi
...
+        merged.extend(data)
+    else:
+        merged.append(data)
+for item in merged:
+    item.setdefault("metadata", {})["target-directory"] = target_directory
+merged.sort(key=lambda item: item.get("metadata", {}).get("project-name", ""))
+with open(out_path, "w") as fh:
+    json.dump(merged, fh)
+print(f"merged {len(sys.argv) - 3} dump(s) into {len(merged)} test result(s)")
+PY
```
robertolopezlopez, line 127: necessary to handle json data instead of relying on `jq` - which may be installed or not

PR #635, opened 2026-06-23
fix: return asset link in test result metadata

bsantaus-snyk:
### Description

In support of Scotia commitment for SBOM report, Test API Shim has been updated to return a link to the tested Asset's Inventory page.
This change retrieves the Asset link from the `AdditionalProperties` in the `GetTest_200_Links` and places it into the `testResult`'s `metadata` under `"asset"`.

To do this, overrides for both `MarshalJSON` and `UnmarshalJSON` have been added into `testapi.go`. The override for `MarshalJSON` is only necessary for testing, as the test cases in `testapi_test.go` have been updated to each return an Asset link. The override for `UnmarshalJSON` is the one required for the implementation, as it will properly retrieve the link included in the response under `"asset"`. 

With this change, the OS flows folks can proceed with adding the Asset link into the test output for the SBOM report flow.

### Checklist

- [x] Linted (`make lint`)


pkg/apiclients/testapi/testapi.go
```diff
@@ -703,6 +774,15 @@ func (h *testHandle) fetchResultStatus(ctx context.Context, testID uuid.UUID) (T
 		result.BreachedPolicies = attrs.Outcome.BreachedPolicies
 	}
 
+	links := resp.ApplicationvndApiJSON200.Links
+	if assetLink, ok := links.AdditionalProperties["asset"]; ok {
+		assetLinkStr, err := assetLink.AsIoSnykApiCommonLinkString()
+		if err != nil {
+			h.client.logger.Debug().AnErr("failed parsing asset link string", err)
+		}
+		result.metadata["asset"] = assetLinkStr
```
basti-snyk: **Should-fix (correctness):** This block stores a value even when parsing failed, and only handles the string form of the link.

`io.snyk.api.common.LinkProperty` is `oneOf: [LinkString, LinkObject]`. If the server returns the `asset` link as an object (`{"href": ...}`), `AsIoSnykApiCommonLinkString()` errors, the error is logged at Debug only, and then `result.metadata["asset"] = assetLinkStr` still runs — storing an empty string. Consumers then can't distinguish "unparseable"/"present-but-empty" from "absent".

The sibling pagination helper (`extractNextCursor`, ~line 905) already handles both union variants. Suggest mirroring it: only set metadata on success, and fall back to the object form's `Href`:
```go
if s, err := assetLink.AsIoSnykApiCommonLinkString(); err == nil {
    result.metadata["asset"] = s
} else if obj, err := assetLink.AsIoSnykApiCommonLinkObject(); err == nil {
    result.metadata["asset"] = obj.Href
} else {
    h.client.logger.Debug().Err(err).Msg("failed parsing asset link")
}
```
Note also that the current Debug log builds a zerolog event but never finalizes it with `.Msg()`/`.Send()`, so nothing is actually logged on failure.

— AI review

pkg/apiclients/testapi/testapi.go
```diff
@@ -654,6 +655,76 @@ func (h *testHandle) pollJobToCompletion(ctx context.Context) (*uuid.UUID, error
...
+// additional properties.
+func (gl GetTest_200_Links) MarshalJSON() ([]byte, error) {
+	object := make(map[string]json.RawMessage)
+	var err error
+
+	object["self"], err = json.Marshal(gl.Self)
+	if err != nil {
+		return nil, err
+	}
+	object["related"], err = json.Marshal(gl.Related)
```
basti-snyk: **Suggestion (wire-format):** `Self` and `Related` are marshaled unconditionally, so a links object with nil `Self`/`Related` serializes to `{"self":null,"related":null}` instead of `{}`. The generated struct tags these `omitempty` (`testapi.gen.go`), so this hand-written `MarshalJSON` diverges from the generated/spec wire shape. Guard each write with a nil check (as the generated `IoSnykApiCommonErrorLink` marshaler does):
```go
if gl.Self != nil { object["self"], err = json.Marshal(gl.Self); ... }
if gl.Related != nil { object["related"], err = json.Marshal(gl.Related); ... }
```
Low impact in practice since the client only unmarshals these in production (this method is exercised by the test mock), but worth matching the generated behavior.

— AI review

pkg/apiclients/testapi/testapi.go
oana-andrisoaei-snyk: Addressed in 6b0aa8f. `fetchResultStatus` now tries `AsIoSnykApiCommonLinkString` first, falls back to `AsIoSnykApiCommonLinkObject().Href`, and only writes to `metadata` on success — mirroring the union handling in `extractNextCursor`. The debug log on the final failure path is now finalized with `.Err(err).Msg("failed parsing asset link")` so it actually emits. Also switched the read/write site to a new typed `TestResultMetadataKeyAsset` constant instead of the bare `"asset"` string.

`Test_AssetLink_ObjectForm` covers the object-form path and `Test_AssetLink_Absent` covers the absent case (asserting `GetMetadataValue("asset")` is nil).

pkg/apiclients/testapi/testapi.go
oana-andrisoaei-snyk: Addressed in 6b0aa8f. `Self` and `Related` are now nil-guarded, so an empty `GetTest_200_Links{}` marshals to `{}` instead of `{"self":null,"related":null}`, matching the `omitempty` tags on the generated struct (and the pattern used by `IoSnykApiCommonErrorLink.MarshalJSON`). Inner marshal errors are wrapped with `fmt.Errorf("error marshaling 'self'/'related': %w", err)` for consistency.

Added `Test_GetTest200Links_MarshalJSON_OmitsEmpty` as a direct round-trip test that locks in the omit-empty behavior and the additional-properties path.

internal/presenters/templates/ufm.human.tmpl
```diff
@@ -305,6 +305,12 @@ Tested {{ int $dependencyCount }} dependencies for known issues, found {{ len $o
         {{- printf "  %s" ( . | bold ) }}
     {{- end -}}
 
+    {{- /* Render asset inventory link when available (sbom test --report) */ -}}
+    {{- with $.GetMetadataValue "asset" }}
+        {{- "\n\n" }}
+        {{- printf "View your asset(s) at: %s" ( . | bold ) }}
+    {{- end -}}
```
danskmt, line 312: are we considering rendering this data only to human readable output?

pkg/apiclients/testapi/testapi.go
```diff
@@ -654,6 +661,83 @@ func (h *testHandle) pollJobToCompletion(ctx context.Context) (*uuid.UUID, error
...
+// oapi-codegen does not emit Marshal/UnmarshalJSON for inline link schemas with
+// AdditionalProperties, so the hand-written overrides below are needed to round-
+// trip the Asset link the API returns under "additionalProperties". If a future
+// regeneration of testapi.gen.go starts emitting these methods, the duplicate
+// definitions will fail to compile and these overrides should be removed (or
+// the underlying spec changed to a $ref so codegen handles it natively).
+
+// UnmarshalJSON handles GetTest_200_Links, including link entries returned under
+// additionalProperties (e.g. the "asset" link).
+func (gl *GetTest_200_Links) UnmarshalJSON(b []byte) error {
```
danskmt, line 674: So we need these custom unmarshal / marshal because of this right?

Image

go.mod
```diff
@@ -77,6 +77,7 @@ require (
 	github.com/jcmturner/rpc/v2 v2.0.3 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/kevinburke/ssh_config v1.2.0 // indirect
+	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
```
danskmt, line 80: I see this new dependency have been added, coming from some other bump.
I'm wondering if those bumps are necessary for this PR?

danskmt: Not relevant though, resolving the conversation

PeterSchafer (approved):

PR #637, opened 2026-06-26
feat: add live check for authentication

CatalinSnyk:
### Description

### Checklist

- [x] Linted (`make lint`)

### DIFF
--- a/pkg/local_workflows/doctor_workflow/doctor_workflow.go
+++ b/pkg/local_workflows/doctor_workflow/doctor_workflow.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/pflag"
 	"golang.org/x/term"
 
+	"github.com/snyk/go-application-framework/pkg/local_workflows/doctor_workflow/livecheck"
 	"github.com/snyk/go-application-framework/pkg/local_workflows/doctor_workflow/logsummary"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
@@ -60,6 +61,14 @@ func runDoctor(invocationCtx workflow.InvocationContext, stdin io.Reader, stdinI
 	logger.Debug().Msgf("doctor: summarized debug log (%d notable events from %d bytes)", len(summary.Highlights), len(debugLog))
 
 	report := summary.Format()
+	if config.GetBool(noLiveCheckFlag) {
+		logger.Debug().Msg("doctor: --no-live-check set, skipping live checks")
+	} else {
+		live := livecheck.Run(invocationCtx)
+		logger.Debug().Msgf("doctor: gathered live context (auth ok=%t)", live.Auth.OK)
+		report += live.Format()
+	}
+
 	outputData := workflow.NewData(
 		workflow.NewTypeIdentifier(WORKFLOWID_DOCTOR, doctorWorkflowName),
 		"text/plain",
--- a/pkg/local_workflows/doctor_workflow/livecheck/auth.go
+++ b/pkg/local_workflows/doctor_workflow/livecheck/auth.go
@@ -1,0 +1,47 @@
+package livecheck
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/snyk/go-application-framework/pkg/workflow"
+)
+
+var workflowIDWhoAmI = workflow.NewWorkflowIdentifier("whoami")
+
+type AuthStatus struct {
+	OK           bool
+	Identity     string
+	ErrorMessage string
+}
+
+func checkAuth(invocationCtx workflow.InvocationContext) AuthStatus {
+	config := invocationCtx.GetConfiguration().Clone()
+
+	data, err := invocationCtx.GetEngine().InvokeWithConfig(workflowIDWhoAmI, config)
+	if err != nil {
+		return AuthStatus{ErrorMessage: err.Error()}
+	}
+
+	if len(data) == 0 {
+		return AuthStatus{ErrorMessage: "whoami returned no usable result"}
+	}
+
+	identity, ok := data[0].GetPayload().(string)
+	if !ok {
+		return AuthStatus{ErrorMessage: "whoami returned no usable result"}
+	}
+	return AuthStatus{OK: true, Identity: identity}
+}
+
+func (a AuthStatus) format() string {
+	var sb strings.Builder
+	sb.WriteString("\n\nAuthentication\n\n")
+	if a.OK {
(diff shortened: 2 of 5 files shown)


pkg/local_workflows/doctor_workflow/livecheck/auth.go
```diff
@@ -0,0 +1,47 @@
...
+var workflowIDWhoAmI = workflow.NewWorkflowIdentifier("whoami")
+
+type AuthStatus struct {
+	OK           bool
+	Identity     string
+	ErrorMessage string
+}
+
+func checkAuth(invocationCtx workflow.InvocationContext) AuthStatus {
+	config := invocationCtx.GetConfiguration().Clone()
```
basti-snyk, line 19: > 🤖 This is an automated AI review comment.

**Should Fix — clone is invoked without forcing `json=false`, so `--json` breaks the auth check.**

`whoami` branches on the configuration's `json` flag (`whoami_workflow.go`): when set, it returns an `application/json` `[]byte` payload instead of the default `string`. Because `checkAuth` clones the doctor's config and passes it through unchanged, if `--json` is present on the doctor invocation the `data[0].GetPayload().(string)` assertion below fails and a fully **authenticated** user is reported as "Failed to verify authentication".

Recommend `config.Set(<jsonFlag>, false)` on the cloned config before `InvokeWithConfig`, so the payload-type contract this code depends on holds regardless of inherited flags.

— AI review

pkg/local_workflows/doctor_workflow/livecheck/auth.go
```diff
@@ -0,0 +1,47 @@
...
+		return AuthStatus{ErrorMessage: err.Error()}
+	}
+
+	if len(data) == 0 {
+		return AuthStatus{ErrorMessage: "whoami returned no usable result"}
+	}
+
+	identity, ok := data[0].GetPayload().(string)
+	if !ok {
+		return AuthStatus{ErrorMessage: "whoami returned no usable result"}
```
basti-snyk, line 32: > 🤖 This is an automated AI review comment.

**Suggestion — empty-result and wrong-payload-type branches emit identical messages.**

This `!ok` branch and the `len(data) == 0` branch above both return `"whoami returned no usable result"`. For a diagnostic tool, these are exactly the two cases worth distinguishing: a failed type assertion signals an unexpected payload shape (e.g. the `--json` `[]byte` path), not a genuinely empty result. A distinct message like `"whoami returned an unexpected payload type"` would make a future regression diagnosable straight from the doctor report.

— AI review

pkg/local_workflows/doctor_workflow/livecheck/auth.go
```diff
@@ -0,0 +1,47 @@
+package livecheck
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/snyk/go-application-framework/pkg/workflow"
+)
+
+var workflowIDWhoAmI = workflow.NewWorkflowIdentifier("whoami")
```
basti-snyk, line 10: > 🤖 This is an automated AI review comment.

**Suggestion — reuse the exported whoami workflow identifier instead of redefining it.**

This re-declares the `"whoami"` magic string locally. `localworkflows.WORKFLOWID_WHOAMI` already exists and is reused elsewhere (e.g. `ignore_workflow.go`). If the canonical workflow name ever changes, the exported constant updates automatically but this local copy would silently drift, leaving the live check invoking a non-existent workflow ID. Prefer importing the shared constant.

— AI review

pkg/local_workflows/doctor_workflow/livecheck/auth.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (a AuthStatus) format() string {
+	var sb strings.Builder
+	sb.WriteString("\n\nAuthentication\n\n")
+	if a.OK {
+		fmt.Fprintf(&sb, "  Authenticated as: %s\n", a.Identity)
+	} else {
+		sb.WriteString("  Failed to verify authentication\n")
+		fmt.Fprintf(&sb, "  Error: %s\n", a.ErrorMessage)
```
basti-snyk, line 44: > 🤖 This is an automated AI review comment.

**Suggestion — raw `err.Error()` is rendered verbatim into a report users often share.**

`ErrorMessage` is the wrapped whoami/engine error, which ultimately includes net/http transport detail (status codes and request **URLs**, e.g. the configured API_URL). It does not include the token, so this is low-impact info disclosure rather than a credential leak — but doctor reports are frequently pasted into support tickets / public issues. Consider a user-facing message in the report and keeping the raw error behind debug logging.

— AI review

pkg/local_workflows/doctor_workflow/doctor_workflow.go
```diff
@@ -60,6 +61,14 @@ func runDoctor(invocationCtx workflow.InvocationContext, stdin io.Reader, stdinI
 	logger.Debug().Msgf("doctor: summarized debug log (%d notable events from %d bytes)", len(summary.Highlights), len(debugLog))
 
 	report := summary.Format()
+	if config.GetBool(noLiveCheckFlag) {
+		logger.Debug().Msg("doctor: --no-live-check set, skipping live checks")
+	} else {
+		live := livecheck.Run(invocationCtx)
```
basti-snyk, line 67: > 🤖 This is an automated AI review comment.

**Should Fix — the live check is unbounded and runs even for offline `--input` analysis.**

Two related concerns at this `livecheck.Run` call:

1. **No timeout.** `checkAuth` invokes `whoami` (a real `/self` network call) over `context.Background()`, and the default HTTP client has `Timeout: 0`. `snyk doctor` is a diagnostic run precisely when the network/auth it is checking may be broken — a hung proxy or unreachable API URL will block `doctor` indefinitely, and the already-computed log summary is never emitted because `live.Format()` is appended before the report is returned. Wrapping the invocation in a `context.WithTimeout` would degrade gracefully to a "could not verify auth" line.

2. **Runs on `--input`.** The live check is gated only by `--no-live-check` (default off). When a support engineer runs `snyk doctor --input customer-debug.log` to triage a *captured* log, this fires a live `whoami` against the **engineer's own** credentials and appends "Authenticated as: engineer@…", which can lead to the wrong conclusion ("auth looks fine") about the customer's environment. Consider skipping (or clearly separating) the live check when analyzing a supplied `--input` file.

— AI review

danskmt:
Closing in favor of https://github.com/snyk/go-application-framework/pull/645

PR #638, opened 2026-06-26
fix(ldx_sync_config): support non-Git folders in config resolution [IDE-1993]

rrama:
### Description

Support [Centralised Configuration for non-Git projects](https://snyksec.atlassian.net/browse/IDE-1993).

`GetMergedConfigForFolder` previously returned an error when no git remote was detected, which blocked LDX-Sync Central Config for non-Git folders, Git repos with no remote, and other VCS tools.

This makes `RemoteUrl` optional: git remote detection is attempted, a missing remote is tolerated (logged at debug), and `RemoteUrl` is only sent to the API when present. Non-Git folders fall back to the user's default org server-side. Auto-org selection stays out of scope for non-Git folders (no remote URL to match against).

A small `buildUserConfigParams` helper was extracted to keep `GetMergedConfigForFolder` under the cyclomatic-complexity limit.

**Part of IDE-1993 — paired with the backend change in ldx-sync (snyk/ldx-sync#128). Separate repos, not stacked.**

### Checklist

- [x] Tests added and all succeed (`make test`) — flipped the two prior "expect error" non-Git cases to success; added coverage for non-Git / git-init-no-remote / UUID+slug org / API error paths and the `is_default` org fallback.
- [x] Linted (`make lint`)
- [ ] Test your changes work for the CLI — downstream CLI bump to follow once merged to main.

### DIFF
--- a/pkg/apiclients/ldx_sync_config/resolver.go
+++ b/pkg/apiclients/ldx_sync_config/resolver.go
@@ -80,6 +80,19 @@ func resolveOrgIdToUUID(ctx context.Context, orgId string, engine workflow.Engin
 	return &parsedUUID, nil
 }
 
+// buildUserConfigParams constructs the GetUserConfigParams for the given remoteUrl.
+func buildUserConfigParams(remoteUrl string) *v20260507.GetUserConfigParams {
+	merged := true
+	params := &v20260507.GetUserConfigParams{
+		Version: "2026-05-07",
+		Merged:  &merged,
+	}
+	if remoteUrl != "" {
+		params.RemoteUrl = &remoteUrl
+	}
+	return params
+}
+
 // GetMergedConfigForFolder retrieves merged LDX-Sync user configuration for the given folder
 func GetMergedConfigForFolder(ctx context.Context, engine workflow.Engine, dir string, orgId string) LdxSyncConfigResult {
 	if dir == "" {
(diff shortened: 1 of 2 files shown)


pkg/apiclients/ldx_sync_config/resolver.go
```diff
@@ -92,17 +105,15 @@ func GetMergedConfigForFolder(ctx context.Context, engine workflow.Engine, dir s
 		return LdxSyncConfigResult{Error: fmt.Errorf("failed to create LDX-Sync client: %w", err)}
 	}
 
-	remoteUrl, err := git.GetRemoteUrl(dir)
-	if err != nil || remoteUrl == "" {
-		return LdxSyncConfigResult{Error: fmt.Errorf("git remote detection failed: %w", err)}
+	// remoteErr is intentionally tolerated: non-Git folders, Git repos with no remote,
+	// and other VCS tools are supported per IDE-1993 and fall back to the user's default org.
+	remoteUrl, remoteErr := git.GetRemoteUrl(dir)
```
basti-snyk, line 110: > 🤖 This is an automated AI review comment.

**Should Fix — all `git.GetRemoteUrl` errors are now tolerated, not just "non-Git folder".**

`git.GetRemoteUrl` collapses every failure mode into one `error` return — not-a-repo, no-remote, *and* genuine faults (corrupt/unreadable `.git`, permission denied, transient I/O). With this change all of them fall through to the default-org path, logged only at `Debug`. So a user inside a valid Git repo whose `.git` is momentarily unreadable now silently resolves to their **personal default org** instead of the repo-mapped org, with no user-visible signal — a wrong-org resolution that is hard to diagnose.

IDE-1993 asks to tolerate non-Git / no-remote folders; tolerating genuine git faults too is broader than the stated intent. If `git.GetRemoteUrl` can distinguish "not a repo / no remote" from an operational failure, consider only falling through on the former and surfacing (or at least `Warn`-logging) true failures. If the all-errors-tolerated behavior is deliberate, a one-line confirmation against IDE-1993 plus bumping the genuine-failure case to `Warn` would make this a documented choice rather than an implicit one.

(Non-blocking: this is a design judgment, not a correctness bug — CI is green and the new tests cover the happy path well.)

— AI review

basti-snyk (approved):

PR #639, opened 2026-06-29
chore: improve error details in case of rate limiting

PeterSchafer:
### Description

This PR makes the error message for rate limiting errors more meaningful and actionable.

### Checklist

- [x] Linted (`make lint`)


pkg/networking/middleware/response.go
```diff
@@ -142,6 +161,34 @@ func errFromStatusCode(code int) error {
...
+	if !errors.As(err, &snykErr) {
+		return err
+	}
+
+	retryDelay := rateLimitRetryDelay(res)
+	if retryDelay <= 0 {
+		return err
+	}
+
+	retryMinutes := int(math.Ceil(retryDelay.Minutes()))
```
basti-snyk: **Should-fix — unbounded retry value reaches the user message (no clamp).**

The retry loop caps the delay (`maxRetryAfter`) before acting, but `enrichRateLimitError` only guards `retryDelay <= 0`. A misbehaving/hostile upstream sending e.g. `X-RateLimit-Reset: 99999999` (or an epoch-style value that `parseRetryDelay`'s numeric branch reads as seconds) renders nonsense like `Retry after: ~1666667 min` and stores `Meta["retry-after-seconds"]=99999999`. On 32-bit builds `int(retryDelay.Seconds())` (line 187) can also overflow. Consider clamping `retryDelay` to the existing `maxRetryAfter` ceiling before formatting, and skip enrichment past the cap.

Separately on this line: `math.Ceil(retryDelay.Minutes())` overstates short throttles — `Retry-After: 5` → `~1 min`, and the human-readable minutes then disagree with the exact `retry-after-seconds` meta. Consider rendering seconds for sub-minute delays.

— AI review

pkg/networking/middleware/retry_middleware.go
```diff
@@ -394,3 +379,18 @@ func parseRetryDelay(headerRetryAfterValue string) time.Duration {
...
+	var retryAfter, rateLimitReset time.Duration
+
+	if v := res.Header.Get("Retry-After"); len(v) > 0 {
+		retryAfter = parseRetryDelay(v)
+	}
+	if v := res.Header.Get("X-RateLimit-Reset"); len(v) > 0 {
+		rateLimitReset = parseRetryDelay(v)
+	}
+
+	return max(retryAfter, rateLimitReset)
```
basti-snyk, line 395: **Suggestion — `max(Retry-After, X-RateLimit-Reset)` now drives a user-facing message, not just retry timing.**

Returning the longer of the two headers is reasonable for the retry loop, but this helper's value is now also rendered to users via `enrichRateLimitError`. When a server sends an authoritative short `Retry-After: 30` alongside a window-based `X-RateLimit-Reset: 3600`, the user is told to wait ~60 min. Worth considering whether `Retry-After` (the explicit per-response directive) should take precedence for the displayed guidance, even if `max` is kept for the retry decision.

Minor: the extraction dropped the inline comment that documented the Envoy `X-RateLimit-Reset` "seconds-until-reset" semantics — worth re-adding here so a future reader doesn't mistake it for an absolute epoch.

— AI review

pkg/networking/middleware/response.go
```diff
@@ -142,6 +161,34 @@ func errFromStatusCode(code int) error {
...
+	retryTime := time.Now().Add(retryDelay).UTC()
```
robertolopezlopez: just UTC? Is it possible to use the client TZ instead?

robertolopezlopez: also, maybe it is worth to show the date

pkg/networking/middleware/response.go
```diff
@@ -142,6 +161,34 @@ func errFromStatusCode(code int) error {
...
+
+	// the default description can be replaced with something more actionable
+	snykErr.Detail = fmt.Sprintf("Retry after: ~%d min (\u2248%s UTC).", retryMinutes, retryTime.Format("15:04"))
```
robertolopezlopez:
```suggestion
	snykErr.Detail = fmt.Sprintf("Retry after: ~%d min (\u2248%s).", retryMinutes, retryTime.Format("15:04 UTC"))
```

Or whatever you prefer to show. 

https://pkg.go.dev/time#pkg-constants

pkg/networking/middleware/retry_middleware.go
PeterSchafer: Question: why should the value that is presented to the user be different to the one used for retrying? Eventually in either case we want to ensure that either a manual or an automatic retry is successful so we want to use the correct waiting time.

pkg/networking/middleware/response.go
PeterSchafer: Using the locale timezone now

pkg/networking/middleware/response.go
PeterSchafer: done!

danskmt (approved):
LGTM

PR #640, opened 2026-06-30
fix: Close response body after reading in getErrorList

danskmt:
### Description

Fixes an unclosed response body leak in `getErrorList()`. After `io.ReadAll(res.Body)` consumed the original transport body, it was replaced with a buffer but never closed, leaving the TCP connection open.

The fix closes the original body after a successful read, then replaces `res.Body` with the buffered copy.

Adds direct unit tests for `getErrorList` that verify the original body is closed after reading (valid JSON and invalid JSON).

Fixes [CLI-1626](https://snyksec.atlassian.net/browse/CLI-1626)

### Checklist

- [x] Linted (`make lint`)
- [X] Test your changes work for the CLI

CLI PR: https://github.com/snyk/cli/pull/6959

[CLI-1626]: https://snyksec.atlassian.net/browse/CLI-1626

### DIFF
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -74,11 +74,12 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
 	if res.Body == nil {
 		return []snyk_errors.Error{}
 	}
+	originalBody := res.Body
+	defer originalBody.Close()
 	// get JSONApiErrors from body
 	bodyBytes, err := io.ReadAll(res.Body)
-	closeErr := res.Body.Close()
 	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
-	if err != nil || closeErr != nil {
+	if err != nil {
 		return []snyk_errors.Error{}
 	}
 
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -76,11 +76,11 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
 	}
 	// get JSONApiErrors from body
 	bodyBytes, err := io.ReadAll(res.Body)
-	if err != nil {
+	closeErr := res.Body.Close()
+	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+	if err != nil || closeErr != nil {
 		return []snyk_errors.Error{}
 	}
-
-	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
 
 	errorList, err := snyk_errors.FromJSONAPIErrorBytes(bodyBytes)
 	if err != nil {
--- a/pkg/networking/middleware/response.go
+++ b/pkg/networking/middleware/response.go
@@ -74,14 +74,14 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
 	if res.Body == nil {
 		return []snyk_errors.Error{}
 	}
-	originalBody := res.Body
-	defer originalBody.Close()
 	// get JSONApiErrors from body
 	bodyBytes, err := io.ReadAll(res.Body)
-	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
 	if err != nil {
 		return []snyk_errors.Error{}
 	}
+
+	_ = res.Body.Close()
+	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
 
 	errorList, err := snyk_errors.FromJSONAPIErrorBytes(bodyBytes)
 	if err != nil {
--- a/pkg/networking/middleware/response_internal_test.go
+++ b/pkg/networking/middleware/response_internal_test.go
@@ -38,8 +38,9 @@ func Test_getErrorList_closesBody(t *testing.T) {
 		tb := &trackingBody{ReadCloser: io.NopCloser(strings.NewReader("not json"))}
 		res := &http.Response{Body: tb}
 
-		getErrorList(res)
+		errors := getErrorList(res)
 
+		assert.Len(t, errors, 0)
 		assert.True(t, tb.closed, "original body must be closed even when JSON parsing fails")
 	})
 }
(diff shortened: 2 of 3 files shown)


pkg/networking/middleware/response.go
```diff
@@ -74,12 +74,12 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
...
-	if err != nil {
+	closeErr := res.Body.Close()
+	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+	if err != nil || closeErr != nil {
 		return []snyk_errors.Error{}
 	}
 
-	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
-
 	errorList, err := snyk_errors.FromJSONAPIErrorBytes(bodyBytes)
```
danskmt, line 86: Here we close and reassign regardless of error for scenarios where we have partial bodyBytes (err != nil), otherwise it would just keep the open _res.Body_ after the early return.
Also, instead of using a defer in this case, closing directly seems better otherwise we have to keep track of a `originalBody` variable, because `res.Body` mutates, and we would be postponing the close

pkg/networking/middleware/response.go
```diff
@@ -74,12 +74,12 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
 	}
 	// get JSONApiErrors from body
 	bodyBytes, err := io.ReadAll(res.Body)
-	if err != nil {
+	closeErr := res.Body.Close()
+	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
```
PeterSchafer: Question: in case of a read error, should we really replace the body? this was different before.

pkg/networking/middleware/response.go
```diff
@@ -74,12 +74,12 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
...
+	if err != nil || closeErr != nil {
```
basti-snyk: **Should fix — a benign `Close()` error now discards a valid, already-read error body.** When `io.ReadAll` succeeds, `bodyBytes` holds the full payload and is independent of the connection teardown — yet `if err != nil || closeErr != nil` returns an empty list whenever `res.Body.Close()` returns a (typically transient) error. That regresses both callers of `getErrorList`:

1. `getErrorsFromResponse` (line 100) falls back to `errFromStatusCode`, dropping the server's real error-catalog Title/Detail in favour of a generic status-code message.
2. `shouldRetry` (retry_middleware.go:317-319) gets an empty list, never matches `MaintenanceWindowError`, and keeps retrying through a maintenance window instead of stopping.

This is exactly the flaky-teardown condition the fix targets. The sibling helper `getResponseBody` (pkg/networking/logging.go:39-43) already does read→close→restore and deliberately ignores the Close error — that's the convention to follow here:

```go
bodyBytes, err := io.ReadAll(res.Body)
_ = res.Body.Close() // best-effort, mirrors getResponseBody
res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
if err != nil {
    return []snyk_errors.Error{}
}
```

Also: there are no test changes in this PR, so the new `closeErr` branch ships with no coverage. Please add a `getErrorList` test where `Body.Close()` errors but the body parses, asserting the parsed errors are still returned.

— AI review

pkg/networking/middleware/response.go
basti-snyk: **Suggestion — partial-read edge case.** The `res.Body` reassignment now runs before the `err` check, so on a failed/partial `io.ReadAll` (e.g. unexpected EOF mid-stream) `res.Body` is replaced with a `NopCloser` over the truncated bytes. Downstream consumers then read truncated content with no error and no truncation signal, whereas the previous code left the original error-bearing reader in place. Minor, but worth a guard or a test if this path matters.

— AI review

pkg/networking/middleware/response.go
```diff
@@ -74,12 +74,12 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
 	}
 	// get JSONApiErrors from body
 	bodyBytes, err := io.ReadAll(res.Body)
-	if err != nil {
+	closeErr := res.Body.Close()
```
PeterSchafer: Question: do we need a test for this?

pkg/networking/middleware/response.go
danskmt: Before the fix, the body wasn't closed on the error path. Now that we close unconditionally, we must also replace unconditionally, otherwise `res.Body` would point to a closed reader. The two operations are coupled: close the original, replace with a buffer.

pkg/networking/middleware/response.go
```diff
@@ -72,14 +72,15 @@ func getErrorList(res *http.Response) []snyk_errors.Error {
 	if res.Body == nil {
 		return []snyk_errors.Error{}
 	}
+	originalBody := res.Body
+	defer originalBody.Close()
 	// get JSONApiErrors from body
 	bodyBytes, err := io.ReadAll(res.Body)
+	res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
```
basti-snyk: **Should Fix — silent body truncation on read error.** Moving the `res.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))` reassignment ahead of the `if err != nil` check means that when `io.ReadAll` fails mid-stream (e.g. `unexpected EOF`, connection reset), `res.Body` is now replaced with a `NopCloser` over the **truncated** `bodyBytes`, and the function still early-returns an empty error list. Downstream consumers (`getErrorsFromResponse` and `shouldRetry` via `retry_middleware.go`) then read a short, well-formed-looking body and never observe the transport error or any truncation marker.

This diverges from the sibling convention in `getResponseBody` (`pkg/networking/logging.go`), which only closes+replaces the body when the read succeeds. Given the unconditional `defer originalBody.Close()`, you can't simply early-return the original body any more — so the real question is whether truncated-buffer-as-success is intended. If yes, please make it a documented, tested decision; if not, the read error should propagate. Either way this new path currently ships untested.

_(Non-blocking observations: the core leak fix itself is correct — original transport body is closed exactly once, no double-close / use-after-close. Separately, the `io.ReadAll` on line 78 is unbounded (no `io.LimitReader`), a pre-existing DoS surface worth a follow-up.)_

— AI review

danskmt: tests added

PeterSchafer (approved):

PeterSchafer (approved):

PR #641, opened 2026-06-30
feat: Versioned log parser for snyk doctor debug log extraction

danskmt:
### Description

Alternative implementation for CLI-1573 (extracting events from debug logs). This PR introduces a versioned, extensible log parsing architecture for `snyk doctor` that supports multiple CLI output formats keyed by CLI version.

**Architecture: 3-phase pipeline**

1. **Lexer** (`lexer.go`) -- tokenizes each log line into a typed token (VersionLine, TableRow, HTTPError, CLIError, SummaryMarker, etc.) using a version-specific `LexerSpec`
2. **Landmark splitter** (`sections.go`) -- finds structural anchors in the token stream and carves it into named regions (Header, Body, Summary, Result) without assuming section order
3. **Collect** (`logsummary.go`) -- processes each region independently to extract highlights, header, and footer

**Key design decisions:**

- **Version-keyed specs**: Each `FormatSpec` is bound to a CLI version range via `VersionConstraint`. The parser auto-detects the CLI version from the log's `Version:` line and selects the matching spec.
- **Derivable via functional options**: New versions are defined as `DeriveFormat(parentSpec, ...)` with only the changed fields. Deep-copy safety prevents parent mutation.
- **Landmark-based splitting over state machine**: No ordering assumptions on sections. Future versions can reorder or add sections without rewiring transitions.
- **76 tests across 6 test files**: Lexer unit tests, section split tests, version detection tests, extensibility/derive tests, multi-version integration tests, plus all edge cases and ported colleague tests.

**Files:**

| File | Purpose |
|------|---------|
| `lexer.go` | Token enum, LexerSpec, tokenize(), DeriveLexer(), functional options |
| `sections.go` | Section enum, Landmark, findLandmarks(), splitByLandmarks(), extractHeaderFromRegion() |
| `formatspec.go` | CLIVersion, VersionConstraint, FormatSpec, DeriveFormat(), registry, detectFormat() |
| `logsummary.go` | Summarize() 3-phase pipeline, Summary struct with CLIVersion + FormatSpecID |
| `format.go` | Human-readable report rendering |

### Checklist

- [ ] Linted (`make lint`)


pkg/local_workflows/doctor_workflow/doctor_workflow.go
```diff
@@ -0,0 +1,93 @@
...
+const (
+	doctorWorkflowName = "doctor"
+
+	inputFlag       = "input"
+	noLiveCheckFlag = "no-live-check"
+)
+
+var WORKFLOWID_DOCTOR workflow.Identifier = workflow.NewWorkflowIdentifier(doctorWorkflowName)
+
+func InitDoctorWorkflow(engine workflow.Engine) error {
```
basti-snyk, line 26: **Should fix — unreachable workflow.** `InitDoctorWorkflow` is never registered with the engine: it is absent from the `initMethods` list in `pkg/local_workflows/local_workflows.go` (the only path `app.go` wires via `AddExtensionInitializer`), and there is no non-test caller anywhere. As shipped, `snyk doctor` is not invokable — the whole feature ships dark. If a follow-up PR in the stack does the wiring, please confirm the ordering; otherwise this is dead code.

— AI review

pkg/local_workflows/doctor_workflow/doctor_workflow.go
```diff
@@ -0,0 +1,93 @@
...
+	flags := pflag.NewFlagSet(doctorWorkflowName, pflag.ExitOnError)
+	flags.String(inputFlag, "", "Path to a Snyk CLI debug log file. If absent, reads debug logs from STDIN.")
+	flags.Bool(noLiveCheckFlag, false, "Skip the live authentication and connectivity checks.")
```
basti-snyk, line 29: **Should fix — dead flag promising non-existent behavior.** `--no-live-check` is registered with help text "Skip the live authentication and connectivity checks", but its value is never read and no live-check logic exists in the workflow (`runDoctor` only reads a log and summarizes it). A CLI flag is a public contract; advertising behavior that doesn't exist is a UX defect in a diagnostic tool. Drop the flag (and its existence assertion in the test) until the live-check feature actually lands.

— AI review

pkg/local_workflows/doctor_workflow/logsummary/summarize.go
```diff
@@ -0,0 +1,126 @@
...
+		if _, dup := seen[tok.Message]; dup {
+			continue
+		}
+		seen[tok.Message] = struct{}{}
+		highlights = append(highlights, Highlight{
+			Line:    tok.Number,
+			Kind:    kind,
+			Message: tok.Message,
+		})
+		if len(highlights) == maxHighlights {
```
basti-snyk, line 141: **Should fix — off-by-one in the truncation flag.** When the 200th unique highlight is appended, `len(highlights) == maxHighlights` returns `truncated=true` even if the log ends right at the cap with nothing further dropped. `Format()` then prints "(showing first 200 notable events; remaining log not scanned)" although every event was shown. Mark truncated only when a (cap+1)th qualifying token is actually encountered. Add a test asserting `Truncated == false` at exactly `maxHighlights` events.

— AI review

pkg/local_workflows/doctor_workflow/logsummary/formatspec.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+var BaseSpec = FormatSpec{
+	ID:            "base",
+	Constraint:    versionRange("0.0.0", ""),
+	Lexer:         baseLexer,
+	LandmarkRules: baseLandmarks,
+}
+
+var registry = []FormatSpec{BaseSpec}
```
basti-snyk: **Should fix — speculative generality for a registry of one.** `registry` holds only `BaseSpec` with an unbounded constraint, so `detectFormat` always resolves to `BaseSpec`, and the entire versioned-format apparatus (`DeriveFormat`/`DeriveLexer`/`WithLexerOverrides`/`WithExtraLandmark`/`WithReplacedLandmarks`/`VersionConstraint`) is exercised only by tests — ~180 lines of production code + ~286 lines of tests supporting a second format that does not exist. Since this is a published module, every exported symbol here is a semver commitment. Suggest collapsing to the one concrete lexer/landmark set now and reintroducing the abstraction (a one-file change) when the second format actually arrives; keep parser internals unexported until a real cross-package consumer exists.

— AI review

pkg/local_workflows/doctor_workflow/logsummary/formatspec.go
```diff
@@ -0,0 +1,201 @@
...
+	TableRowRe:     baseTableRe,
+	BodyClassifiers: []BodyClassifier{
+		{Match: func(msg string) bool { return responseRe.MatchString(msg) }, Token: TokenHTTPError},
+		{Match: func(msg string) bool { return strings.HasPrefix(msg, "< error:") }, Token: TokenCLIError},
+		{Match: func(msg string) bool { return strings.HasPrefix(msg, "Failed ") }, Token: TokenFailedLine},
+	},
+}
+
+var baseLandmarks = []LandmarkRule{
+	{AnchorToken: TokenVersionLine, OpensSection: SectionHeader},
```
basti-snyk: **Suggestion — latent section-landmark data loss.** `TokenVersionLine` opens `SectionHeader`. A clean exit-0 run whose Summary block repeats the Version/Platform table (but has no Errors marker and no `Exit Code:`) would have that in-summary region reassigned to `SectionHeader`, and `joinFooter` would fall back to an empty `SectionSummary`, silently dropping the summary detail. Both bundled fixtures carry an `Exit Code:` (which opens `SectionResult` and shields the footer), so the test suite masks this — `Test_Summarize_summaryMarkerButNoErrorsAndNoExitCode` only asserts the weak `NotEmpty`. Worth a fixture that exercises a repeated Version line inside the Summary with no Exit Code.

— AI review

pkg/local_workflows/doctor_workflow/logsummary/formatspec.go
```diff
@@ -0,0 +1,237 @@
...
+
+// BaseSpec is the fallback format spec that matches all CLI versions.
+var BaseSpec = FormatSpec{
+	ID:            "base",
+	Constraint:    versionRange("0.0.0", ""),
+	Lexer:         baseLexer,
+	LandmarkRules: baseLandmarks,
+}
+
+var registry = []FormatSpec{BaseSpec}
```
basti-snyk: **Should fix — versioned-format apparatus is a registry of one, and the lookup has an ordering trap.** `registry = []FormatSpec{BaseSpec}` holds a single spec whose constraint matches every version, so `detectFormat` always resolves to `BaseSpec` and the entire version-selection surface (`DeriveFormat`/`DeriveLexer`/the `With*` options/`VersionConstraint`) is exercised only by tests. Two concrete concerns: (1) since this is a published module, every exported symbol here is a semver commitment for code that currently has no production caller — consider keeping the parser internals unexported until a real second format lands. (2) `detectFormat` returns the *first* matching spec; because `BaseSpec` matches all versions, the natural `append(registry, v2Spec)` extension would place `BaseSpec` first and the specific spec would never be selected. If the layer stays, the specific-before-base ordering is load-bearing and should be documented/enforced. (Re-posting at the current HEAD; my earlier note on this became outdated after the refactor commit.)

— AI review

pkg/local_workflows/doctor_workflow/logsummary/format.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	writeSection(&sb, "Environment", summary.Header)
+
+	sb.WriteString("\n\nNotable Events\n\n")
+	if len(summary.Highlights) == 0 {
+		sb.WriteString("  No failing requests or CLI error entries found in the log body.\n")
+	} else {
+		for _, h := range summary.Highlights {
+			fmt.Fprintf(&sb, "  L%d [%s] %s\n", h.Line, h.Kind, h.Message)
```
basti-snyk, line 25: **Should fix — log content is reproduced verbatim into a shareable report with no redaction.** `Format()` renders the Environment header, the matched highlights, and the Result footer straight from the input log into the plaintext diagnostic report. `snyk doctor --input <path>` accepts any file: while the current CLI pre-masks its own `Authorization` value, older CLI versions or third-party/custom logs may carry unmasked bearer tokens, cookies, org IDs, emails, or internal URLs — all faithfully copied into output that is explicitly intended to be handed to support. The summarizer trusts its input to be pre-sanitized and provides no defense-in-depth. Recommend a redaction/masking pass over header, footer, and highlight `Message` values before they reach the report.

— AI review

pkg/local_workflows/doctor_workflow/doctor_workflow.go
basti-snyk, line 29: **Should fix — `--no-live-check` flag is advertised but never read.** The flag is registered with help text "Skip the live authentication and connectivity checks," but `runDoctor` never reads `noLiveCheckFlag`, and there are no live checks anywhere in this PR. A user who passes `--no-live-check` (or reads `--help`) is promised behavior that does not exist. If the live-check is a later PR in the stack, add the flag there alongside the code that honors it rather than shipping an inert flag now.

— AI review

pkg/local_workflows/doctor_workflow/logsummary/logsummary.go
```diff
@@ -0,0 +1,146 @@
...
+	headerTokens, bodyRemainder := extractHeaderFromRegion(sections[SectionHeader])
+
+	var bodyTokens []TokenizedLine
+	bodyTokens = append(bodyTokens, sections[SectionPreamble]...)
+	bodyTokens = append(bodyTokens, bodyRemainder...)
+	bodyTokens = append(bodyTokens, sections[SectionBody]...)
+
+	highlights, truncated := collectHighlights(bodyTokens)
+
+	cliVer, _ := extractCLIVersion(rawLines)
```
basti-snyk: **Should fix — CLI version is extracted twice per `Summarize`, and the two paths disagree on accepted prefixes.** `detectFormat(rawLines)` (line 47) already calls `extractCLIVersion` internally; line 61 calls it again to populate `Summary.CLIVersion`, so the whole log is scanned twice. Worse, the detection path accepts only `Version:` while the standalone `extractCLIVersion` also accepts `CLI Version:` — a log using only `CLI Version:` yields a populated `CLIVersion` but is still detected as `base`, an internal inconsistency. Have `detectFormat` return the parsed version and thread it through, using one prefix list.

— AI review

pkg/local_workflows/doctor_workflow/logparse/sections.go
```diff
@@ -0,0 +1,139 @@
...
+// extractHeaderFromRegion finds the version line and its trailing table rows
+// within a region. Returns (header tokens, remaining body tokens).
+func extractHeaderFromRegion(region []TokenizedLine) (header, rest []TokenizedLine) {
+	if len(region) == 0 {
+		return nil, nil
+	}
+
+	start := -1
+	for i, tok := range region {
+		if tok.Token == TokenVersionLine {
```
basti-snyk, line 120: **Should fix — a `Version:` line reprinted inside the Summary block is misclassified as a section landmark.** Real Snyk debug logs reprint `Version:`/`Platform:`/`API:` inside the `--- Summary ---` block (see `testdata/empty-config.logs`). Here a `TokenVersionLine` is mapped to open `SectionHeader`, so when it appears inside the summary region it ends `SectionSummary` early and the reprinted environment lines get pulled into the header and dropped. For a log with a Summary marker but no Errors marker / Exit Code, `joinFooter` falls back to `SectionSummary` and the real summary content is lost from the footer. Consider only treating a `Version:` line as a header landmark before the first section opens.

— AI review

pkg/local_workflows/doctor_workflow/logsummary/summarize.go
```diff
@@ -0,0 +1,146 @@
...
+		footerTokens = resultTokens
+	} else if len(summaryTokens) > 0 {
+		footerTokens = summaryTokens
+	}
+
+	result := joinMessages(footerTokens)
+	return trimTrailingBlanks(result)
+}
+
+func trimTrailingBlanks(s string) string {
```
basti-snyk, line 190: **Suggestion — `trimTrailingBlanks` loop is redundant (and misses whitespace-only trailing lines).** `strings.TrimRight(s, " \n")` already strips all trailing spaces and newlines in a single pass, so the surrounding loop and the `"\n "` suffix special-case never iterate more than once; an input ending in `"foo\n   "` is also not handled by the `HasSuffix` guards. Replace the body with a single `return strings.TrimRight(s, " \n")`. No behavior change for the intended cases, plus it fixes the whitespace-only-line case.

— AI review

robertolopezlopez:
please add a matching PR in cli, if there isn't any

robertolopezlopez:
please split `pkg/local_workflows/doctor_workflow/logsummary/*` into sub-packages 🥺

pkg/networking/middleware/retry_middleware.go
```diff
@@ -329,22 +329,7 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
 			return backoff.Permanent(errRetryNecessary)
 		}
 
-		fixRetryDelay := time.Duration(0)
```
robertolopezlopez: this change is good, but it is not strictly related to CLI-1573 - do you mind to drop it off? The PR is already  large

pkg/local_workflows/doctor_workflow/logsummary/lexer.go
```diff
@@ -0,0 +1,203 @@
+package logsummary
```
robertolopezlopez: nit: can you please add comments at +- the beginning of each new file, specifying its intent?

pkg/networking/middleware/retry_middleware.go
danskmt: was outdated, rebased

pkg/local_workflows/doctor_workflow/logsummary/formatspec.go
```diff
@@ -0,0 +1,237 @@
...
+	return CLIVersion{}, false
+}
+
+func detectFormat(rawLines []string) FormatSpec {
+	if ver, ok := extractCLIVersion(rawLines); ok {
+		for _, spec := range registry {
+			if spec.Constraint.Contains(ver) {
+				return spec
+			}
+		}
```
robertolopezlopez: this loop does not make much sense, are you adding any spec to registry apart of BaseSpec? why a loop then

pkg/local_workflows/doctor_workflow/logsummary/formatspec.go
```diff
@@ -0,0 +1,237 @@
...
+					return v, true
+				}
+			}
+		}
+	}
+	return CLIVersion{}, false
+}
+
+func detectFormat(rawLines []string) FormatSpec {
+	if ver, ok := extractCLIVersion(rawLines); ok {
```
robertolopezlopez: should we error out `if !ok` ?

pkg/local_workflows/doctor_workflow/logsummary/multiversion_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Should fall back to base (or latest known), not panic
+	assert.NotEmpty(t, result.FormatSpecID)
+	require.Len(t, result.Highlights, 1)
+	assert.Equal(t, EventError, result.Highlights[0].Kind)
+}
+
+func TestSummarize_twoVersionsSameEngine(t *testing.T) {
+	// Save and restore registry for this test
+	origRegistry := registry
+	defer func() { registry = origRegistry }()
```
robertolopezlopez: Weak test... (writing to package level registry). We can live with this as they are not run in parallel. Fine

pkg/local_workflows/doctor_workflow/logsummary/formatspec.go
danskmt: The registry loop is the version-dispatch mechanism for the multi-format requirement — logparse is a generic engine and can't assume a single spec. It's covered by TestSummarize_twoVersionsSameEngine/TestSummarize_derivedSpecWithExtraClassifier, which register a v2 spec alongside base. The doctor registry has one entry today only because no format has diverged; adding V1/V2 is a data-only change (a DeriveFormat entry), no code change to Detect.

danskmt:
Closed in favor of https://github.com/snyk/go-application-framework/pull/643

PR #645, opened 2026-07-02
feat: add live authentication check

danskmt:
Adds a live authentication check to snyk doctor, ported to the current diagnosis architecture (targets feat/snyk-doctor). A new livecheck package invokes the whoami workflow and maps the result into the shared diagnosis.Finding contract (source: "auth"): success yields an info finding with the identity, failure an error finding with the underlying message. Findings are appended to the report's findings stream, so the existing generic formatter renders them under an "Authentication" section and they appear in --json with no downstream changes. Gated by the existing --no-live-check flag.

This is the same behavior as the original feat/CLI-1574_auth_context (d3619aa), re-expressed against the generic findings contract instead of the old scaffold's string-concatenated report.

### DIFF
--- a/pkg/local_workflows/doctor_workflow/init.go
+++ b/pkg/local_workflows/doctor_workflow/init.go
@@ -9,9 +9,9 @@ import (
 const (
 	doctorWorkflowName = "doctor"
 
-	inputFlag       = "input"
-	noLiveCheckFlag = "no-live-check"
-	jsonFlag        = "json"
+	inputFlag = "input"
+	liveFlag  = "live"
+	jsonFlag  = "json"
 )
 
 var WORKFLOWID_DOCTOR workflow.Identifier = workflow.NewWorkflowIdentifier(doctorWorkflowName)
@@ -19,7 +19,7 @@ var WORKFLOWID_DOCTOR workflow.Identifier = workflow.NewWorkflowIdentifier(docto
 func InitDoctorWorkflow(engine workflow.Engine) error {
 	flags := pflag.NewFlagSet(doctorWorkflowName, pflag.ExitOnError)
 	flags.String(inputFlag, "", "Path to a Snyk CLI debug log file.")
-	flags.Bool(noLiveCheckFlag, false, "Skip live authentication and connectivity checks.")
+	flags.Bool(liveFlag, false, "Run live checks (authentication, connectivity) against the current environment. Off by default.")
 	flags.Bool(jsonFlag, false, "Output the diagnostic report as JSON.")
 
 	_, err := engine.Register(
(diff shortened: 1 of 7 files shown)


pkg/local_workflows/doctor_workflow/livecheck/auth.go
```diff
@@ -0,0 +1,87 @@
...
+	// Force whoami's string-payload path regardless of the doctor run's --json.
+	config.Set(whoamiJSONFlag, false)
+
+	type invokeResult struct {
+		data []workflow.Data
+		err  error
+	}
+	resultCh := make(chan invokeResult, 1)
+	go func() {
+		data, err := invocationCtx.GetEngine().InvokeWithConfig(WhoAmIWorkflowID, config)
```
PeterSchafer: Suggestion: Just use `Invoke(...,WithConfig())` instead of `InvokeWithConfig()`

pkg/local_workflows/doctor_workflow/livecheck/auth.go
```diff
@@ -0,0 +1,87 @@
...
+		resultCh <- invokeResult{data: data, err: err}
+	}()
+
+	select {
+	case <-time.After(authCheckTimeout):
```
PeterSchafer: This seems to unnecessary, just use a context timeout

pkg/local_workflows/doctor_workflow/livecheck/livecheck.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/snyk/go-application-framework/pkg/local_workflows/doctor_workflow/diagnosis"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+)
+
+// Run executes the live checks and returns their findings for inclusion in the
+// doctor report. New checks append their findings here.
+func Run(invocationCtx workflow.InvocationContext) []diagnosis.Finding {
+	return []diagnosis.Finding{
+		checkAuth(invocationCtx).finding(),
```
PeterSchafer, line 15: Praise: very nicely extensible

pkg/local_workflows/doctor_workflow/init.go
```diff
@@ -9,17 +9,17 @@ import (
 const (
 	doctorWorkflowName = "doctor"
 
-	inputFlag       = "input"
-	noLiveCheckFlag = "no-live-check"
-	jsonFlag        = "json"
+	inputFlag = "input"
+	liveFlag  = "live"
```
PeterSchafer, line 13: Praise: makes more sense now

PeterSchafer (approved):

PR #649, opened 2026-07-07
fix(networking): add full jitter to X-RateLimit-Reset retry delay [IDE-2094]

nick-y-snyk:
## Problem

When ldx-sync (or any Snyk API behind the Gloo/Envoy api-gateway) returns a 429, all
clients receive the same `X-RateLimit-Reset` value — the seconds until the token-bucket
window resets (e.g. `1` for a per-second bucket). The retry middleware honoured this
value literally, so all clients woke up at the same millisecond and fired a synchronized
retry wave that immediately re-saturated the rate-limited endpoint.

Root cause traced during ldx-sync thundering-herd analysis (IDE-2094): ~150 concurrent
IDE clients at Monday 9am, all hitting the 160 req/s api-gateway cap, all sleeping
exactly 1 s, all retrying in unison.

## Fix

Apply **full jitter** to the `X-RateLimit-Reset` path: sleep a random duration in
`[0, reset)` instead of the exact reset value. This spreads retries uniformly across
the reset window, eliminating the synchronized wave.

`Retry-After` (RFC 7231) is **not** jittered — its semantics require the client to
wait *at least* the specified duration, and it is typically set by application servers
with deliberate backpressure intent.

## Implementation

- Split `rateLimitRetryDelay` (raw value → used for error reporting: "retry after N
  seconds") from `rateLimitRetryDelayJittered` (jittered value → used for the actual
  sleep in the retry loop).
- `Retry-After` present → return exact, no jitter (takes precedence).
- `X-RateLimit-Reset` only → return `rand.N(reset)`.
- Updated tests: exact-duration assertions on `X-RateLimit-Reset` replaced with
  range assertions `[0, reset)`; `Retry-After` precedence test corrected to assert
  the `Retry-After` value exactly (previous test asserted `max(retryAfter, reset)`
  which was wrong — `Retry-After` should win, not the larger of the two).

## Test plan

- [ ] `go test ./pkg/networking/...` passes (195 → 197 tests, all green)
- [ ] `Test_shouldRetry_rateLimitResetJitter` — new test asserts jitter range and Retry-After precedence
- [ ] `Test_ResponseMiddleware_RateLimitEnrichment` — error metadata still shows raw reset value (300 s), not jittered value

---

> [!NOTE]
> **Medium Risk**
> Changes real sleep timing for all rate-limited HTTP retries across framework clients; impact is bounded (extra wait capped at 2s) but affects shared networking behavior on 429/503 paths.
> 
> **Overview**
> Rate-limit retries no longer sleep only the exact `Retry-After` / `X-RateLimit-Reset` value from `shouldRetry`. When that combined delay (`max` of the two headers) is positive, the middleware now waits **at least** that duration and adds a random extra in **`[0, 2s)`** via `math/rand/v2`, so concurrent clients stagger retries instead of waking together.
> 
> `rateLimitRetryDelay` is documented as the **raw** header-derived delay (still used for max-wait checks and response enrichment); the actual `RetryAfterError` duration includes jitter. Table-driven tests that asserted fixed sleeps were removed in favor of **`Test_shouldRetry_rateLimitResetJitter`**, which checks delays fall in `[rawDelay, rawDelay + maxJitterWindow)`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dcced1082d1ba87efb9f6987d3ebaf6ca1797c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

___

### **PR Type**
Bug fix, Enhancement, Tests

___

### **Description**
- Add random jitter to retry delays.

- Prevent synchronized retry storms.

- Enhance rate limiting handling logic.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[Receive 429 Response] --> B{Extract Retry-After & X-RateLimit-Reset};
  B --> C{Determine Base Delay (max of headers)};
  C --> D{Check Base Delay against maxRetryAfter};
  D -- Exceeds --> E[Permanent Error];
  D -- Within Cap --> F{Calculate Jittered Delay};
  F --> G(Sleep for Jittered Delay);
  G --> H[Retry Request];
  subgraph Jitter Logic
    F -- Base Delay --> F1["Base Delay + Random [0, 2s)"];
  end
  F1 --> G;
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>retry_middleware.go</strong><dd><code>Implement jittered retry delay calculation</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/networking/middleware/retry_middleware.go

<ul><li>Imports <code>math/rand/v2</code> for random number generation.<br> <li> Introduces a <code>maxJitterWindow</code> constant for the jitter duration.<br> <li> Modifies <code>shouldRetry</code> to calculate <code>rawDelay</code> from <code>Retry-After</code> or <br><code>X-RateLimit-Reset</code>.<br> <li> Applies jitter by adding a random duration <code>[0, maxJitterWindow)</code> to the <br><code>rawDelay</code>.<br> <li> Updates comments to reflect that both headers are now treated <br>similarly for jittering.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/649/changes#diff-295d20b71bb7f7fdc47d3503d7fb17595142bb2c48dd5801cc4ad0517afb1a55">+22/-3</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>retry_middleware_test.go</strong><dd><code>Add comprehensive tests for jittered retry delays</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/networking/middleware/retry_middleware_test.go

<ul><li>Removes specific tests for <code>Retry-After</code> and <code>X-RateLimit-Reset</code> from <br><code>Test_shouldRetry_rateLimitResetHeaders</code>.<br> <li> Adds a new test suite <code>Test_shouldRetry_rateLimitResetJitter</code> to cover <br>jittered delays.<br> <li> Tests cover <code>X-RateLimit-Reset</code> alone, <code>Retry-After</code> alone, and cases with <br>both headers present.<br> <li> Verifies that the jittered delay falls within the expected range <br><code>[base_delay, base_delay + maxJitterWindow)</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/649/changes#diff-781d13f3ad7692fdffe97f224e749e9e4cbbab29fbefafcfb11f78de8772aa26">+65/-30</a>&nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -402,6 +402,12 @@ func rateLimitRetryDelay(res *http.Response) time.Duration {
 	return max(retryAfter, rateLimitReset)
 }
 
+// maxJitterWindow is the extra random delay added on top of the reset window.
+// All clients wait at least the reset duration (guaranteeing the bucket has
+// refilled), then each adds a random extra in [0, maxJitterWindow) to
+// desynchronize their retries.
+const maxJitterWindow = 2 * time.Second
+
 // rateLimitRetryDelayJittered returns the actual duration to sleep before retrying
 // a rate-limited request.
 //
@@ -409,11 +415,11 @@ func rateLimitRetryDelay(res *http.Response) time.Duration {
 // at least the specified duration, so no jitter is applied.
 //
 // X-RateLimit-Reset (Envoy/Gloo token-bucket) carries the seconds until the
-// current window resets. All clients that hit the same rate-limit window receive
-// the same reset value, which causes a synchronized retry storm if honored
-// literally. Full jitter — sleeping a random duration in [0, reset) — spreads
-// retries evenly across the reset window and eliminates the thundering herd.
-// If Retry-After is also present it takes precedence (exact, no jitter).
+// current window resets. All clients that received the same 429 share the same
+// reset value, so retrying at exactly reset causes a synchronized spike.
+// The fix: every client waits the full reset (bucket guaranteed refilled) then
+// adds a random extra delay in [0, maxJitterWindow). This desynchronizes retries
+// without risking a retry before the bucket has actually recovered.
 func rateLimitRetryDelayJittered(res *http.Response) time.Duration {
 	if v := res.Header.Get("Retry-After"); len(v) > 0 {
 		if d := parseRetryDelay(v); d > 0 {
@@ -422,9 +428,9 @@ func rateLimitRetryDelayJittered(res *http.Response) time.Duration {
 	}
 	if v := res.Header.Get("X-RateLimit-Reset"); len(v) > 0 {
 		if d := parseRetryDelay(v); d > 0 {
-			// Full jitter: random in [0, d). Desynchronises clients that all
-			// received the same token-bucket reset deadline.
-			return time.Duration(rand.N(int64(d)))
+			// Wait the full reset window (bucket refills), then add jitter so
+			// concurrent clients don't all retry at the same instant.
+			return d + time.Duration(rand.N(int64(maxJitterWindow)))
 		}
 	}
 	return 0
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -410,7 +410,7 @@ func rateLimitRetryDelay(res *http.Response) time.Duration {
 //
 // X-RateLimit-Reset (Envoy/Gloo token-bucket) carries the seconds until the
 // current window resets. All clients that hit the same rate-limit window receive
-// the same reset value, which causes a synchronized retry storm if honoured
+// the same reset value, which causes a synchronized retry storm if honored
 // literally. Full jitter — sleeping a random duration in [0, reset) — spreads
 // retries evenly across the reset window and eliminates the thundering herd.
 // If Retry-After is also present it takes precedence (exact, no jitter).
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand/v2"
 	"net/http"
 	"strconv"
 	"time"
@@ -329,7 +330,7 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
 			return backoff.Permanent(errRetryNecessary)
 		}
 
-		timeToWait := rateLimitRetryDelay(response)
+		timeToWait := rateLimitRetryDelayJittered(response)
 		if timeToWait > maxRetryAfter {
 			return backoff.Permanent(errRetryDelayMaxExceeded)
 		}
@@ -381,7 +382,9 @@ func parseRetryDelay(headerRetryAfterValue string) time.Duration {
 }
 
 // rateLimitRetryDelay extracts the longer of Retry-After and X-RateLimit-Reset
-// durations from the response headers.
+// durations from the response headers. The returned value is the raw header
+// value — suitable for error reporting ("retry after N seconds") but not for
+// the actual sleep, which should use rateLimitRetryDelayJittered.
 func rateLimitRetryDelay(res *http.Response) time.Duration {
 	var retryAfter, rateLimitReset time.Duration
 
@@ -393,4 +396,32 @@ func rateLimitRetryDelay(res *http.Response) time.Duration {
 	}
 
 	return max(retryAfter, rateLimitReset)
+}
+
+// rateLimitRetryDelayJittered returns the actual duration to sleep before retrying
+// a rate-limited request.
+//
+// Retry-After is respected exactly — RFC 7231 semantics require clients to wait
+// at least the specified duration, so no jitter is applied.
+//
+// X-RateLimit-Reset (Envoy/Gloo token-bucket) carries the seconds until the
+// current window resets. All clients that hit the same rate-limit window receive
+// the same reset value, which causes a synchronized retry storm if honoured
+// literally. Full jitter — sleeping a random duration in [0, reset) — spreads
+// retries evenly across the reset window and eliminates the thundering herd.
+// If Retry-After is also present it takes precedence (exact, no jitter).
+func rateLimitRetryDelayJittered(res *http.Response) time.Duration {
+	if v := res.Header.Get("Retry-After"); len(v) > 0 {
+		if d := parseRetryDelay(v); d > 0 {
+			return d
+		}
+	}
+	if v := res.Header.Get("X-RateLimit-Reset"); len(v) > 0 {
+		if d := parseRetryDelay(v); d > 0 {
+			// Full jitter: random in [0, d). Desynchronises clients that all
+			// received the same token-bucket reset deadline.
+			return time.Duration(rand.N(int64(d)))
+		}
+	}
+	return 0
 }
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -330,10 +330,14 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
 			return backoff.Permanent(errRetryNecessary)
 		}
 
-		timeToWait := rateLimitRetryDelayJittered(response)
-		if timeToWait > maxRetryAfter {
+		// Cap check must use the raw header value, not the jittered one.
+		// Jitter applied before the cap could allow a huge reset (e.g. 4 years)
+		// to slip through when rand lands below maxRetryAfter.
+		if rawDelay := rateLimitRetryDelay(response); rawDelay > maxRetryAfter {
 			return backoff.Permanent(errRetryDelayMaxExceeded)
 		}
+
+		timeToWait := rateLimitRetryDelayJittered(response)
 
 		// if a retry after is defined, this is the time to wait for
 		if timeToWait > 0 {
--- a/pkg/networking/middleware/retry_middleware.go
+++ b/pkg/networking/middleware/retry_middleware.go
@@ -10,6 +10,7 @@ import (
 	"math/rand/v2"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff/v5"
@@ -341,9 +342,11 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
 			timeToWait = retryAfter
 		} else if rawDelay > 0 {
 			// X-RateLimit-Reset: wait the full reset window (bucket guaranteed
-			// refilled) then add random extra in [0, maxJitterWindow) to
-			// desynchronize concurrent clients.
-			timeToWait = rawDelay + time.Duration(rand.N(int64(maxJitterWindow)))
+			// refilled) then add random extra in [0, jitter window) to
+			// desynchronize concurrent clients. The jitter window scales with
+			// the bucket's actual capacity (see rateLimitJitterWindow) instead
+			// of a flat constant, so low-capacity buckets get more spread.
+			timeToWait = rawDelay + time.Duration(rand.N(int64(rateLimitJitterWindow(response, rawDelay))))
 		}
 
 		// if a retry after is defined, this is the time to wait for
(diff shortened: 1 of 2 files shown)


pkg/networking/middleware/retry_middleware.go
```diff
@@ -394,3 +401,31 @@ func rateLimitRetryDelay(res *http.Response) time.Duration {
...
+	if v := res.Header.Get("Retry-After"); len(v) > 0 {
+		if d := parseRetryDelay(v); d > 0 {
+			return d
+		}
+	}
+	if v := res.Header.Get("X-RateLimit-Reset"); len(v) > 0 {
+		if d := parseRetryDelay(v); d > 0 {
+			// Full jitter: random in [0, d). Desynchronises clients that all
+			// received the same token-bucket reset deadline.
+			return time.Duration(rand.N(int64(d)))
```
robertolopezlopez: Not AI comment:

I do not fully get the reasoning behind this change. If the server returns a X-RateLimit-Reset of 1h (the maximum value returned by Envoy, or at least 2 months ago it was) but the client retries after 5m, it will still be rate limited for the next 55m (still synchronised to the other clients). If we want to desynchronise clients with the same deadline, I would rather add some seconds delay and get a 200 response instead.

Please let me know your thoughts, maybe I have not understood you properly.

nick-y-snyk: Good catch on the 1h scenario — you're right that jitter doesn't help there, and actually makes it worse by wasting retry attempts during a guaranteed-failure window.

The fix targets a specific scenario: the **per-second bucket** (160 req/s, `X-RateLimit-Reset: 1`). At Monday 9am with ~150 concurrent IDE clients all hitting the per-second cap simultaneously, they all get `X-RateLimit-Reset: 1`. Without jitter: all retry at exactly T+1s → second synchronized wave hits the same 160 req/s wall. With jitter in [0, 1s): retries spread across the 1-second window → ~150 req spread over 1s ≈ 150 req/s, which is just under the 160 req/s cap.

For long resets (hour bucket, `X-RateLimit-Reset: 3600`): jitter here does what you describe — clients retry at random points in the next hour, each getting 429 until the bucket resets, burning their 3 attempts on guaranteed failures. That's a valid concern.

One way to handle both: only apply jitter for short windows (e.g. reset < 60s) and honor the full delay for long ones. That way per-second and per-minute bursts get spread, while per-hour exhaustion waits it out:

```go
if d > 0 && d < 60*time.Second {
    return time.Duration(rand.N(int64(d))) // jitter: spread retries within short reset window
}
return d // long window: honor exactly, don't waste retry attempts
```

Does that address your concern? Happy to update if you prefer a different threshold or approach.

robertolopezlopez: I get it: 150 req/sec getting throttled with the same x-ratelimit-reset.

Which would be the behaviour with reset=2s? `return time.Duration(<0 or 1>)`. The clients would still be hit by the x-ratelimit-reset 429 responses. With 1s, all rate limited requests would be retried immediately (and rate limited)

I still think it would be better to add an extra padding of `rand.N() seconds` to the reset period to somehow spread that spike.

robertolopezlopez: could you please extend your proposed snippet to have a clearer view?

nick-y-snyk: Updating my previous reply — I had the jitter direction wrong.

Sleeping random in `[0, reset)` risks retrying **before** the bucket refills → guaranteed 429 → wasted attempt. That's the bug you were pointing at.

Fixed in the latest commit: sleep `reset + random(0, 2s)`. Every client waits the full reset window (bucket guaranteed recovered), then adds up to 2s of random extra delay. All retries land in `[reset, reset+2s)` — bucket is definitely refilled, clients are desynchronized.

This works for all window sizes without needing the 60s threshold logic I proposed earlier. That was unnecessary complexity.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -329,11 +330,15 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
...
-		timeToWait := rateLimitRetryDelay(response)
-		if timeToWait > maxRetryAfter {
+		// Cap check must use the raw header value, not the jittered one.
+		// Jitter applied before the cap could allow a huge reset (e.g. 4 years)
+		// to slip through when rand lands below maxRetryAfter.
+		if rawDelay := rateLimitRetryDelay(response); rawDelay > maxRetryAfter {
 			return backoff.Permanent(errRetryDelayMaxExceeded)
 		}
 
+		timeToWait := rateLimitRetryDelayJittered(response)
```
PeterSchafer: Question: Why not just add the jitter to `rawDelay`? this avoids duplicating code and re-reading the same values from the headers

pkg/networking/middleware/retry_middleware.go
nick-y-snyk: The new approach is safer. 

We work with X-RateLimit-Reset, which says when the bucket is fully nuked, not when the next token is available, so it's possible to get a token before the actual reset, but it has some edge cases because of 3 ratelimits (per second/min/hour)

pkg/networking/middleware/retry_middleware.go
nick-y-snyk: good call

nick-y-snyk: Good point — applied in 9bbb8b4. `rateLimitRetryDelayJittered` is gone; jitter is now inlined directly in `shouldRetry` using `rawDelay`. One subtlety: `rateLimitRetryDelay` returns `max(retryAfter, rateLimitReset)`, so blindly adding jitter to it would break the Retry-After-takes-precedence case when reset > retry-after. Fixed by checking `parseRetryDelay(Retry-After)` directly before the jitter branch.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -329,11 +330,22 @@ func shouldRetry(response *http.Response, attempts int, maxAttempts int) error {
...
-		timeToWait := rateLimitRetryDelay(response)
-		if timeToWait > maxRetryAfter {
+		rawDelay := rateLimitRetryDelay(response)
+		if rawDelay > maxRetryAfter {
 			return backoff.Permanent(errRetryDelayMaxExceeded)
 		}
 
+		var timeToWait time.Duration
+		if retryAfter := parseRetryDelay(response.Header.Get("Retry-After")); retryAfter > 0 {
+			// Retry-After is an explicit server directive: honor exactly, no jitter.
```
PeterSchafer: Question: Is this true what the comment says? I mean the header is not called `Retry-In` but `Retry-After`.

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: Yes of course if you get a new token the whole 429 and x-ratelimit-reset constraint may be gone but that may have other implications.

Anyway that complexity is not handled here if I am not mistaken.

Having said that: probably you should make `maxJitterWindow` bigger because depending on the spike size the client risks being rate limited again (i.e. if we split 500 req/sec in two 250 req/sec, the client may be once again throttled)

pkg/networking/middleware/retry_middleware.go
nick-y-snyk: nevermind the comment. Do you think it's worth adding jitter here too? My specific case was to handle ldx-sync api, which uses another header, I can also add this header

pkg/networking/middleware/retry_middleware.go
nick-y-snyk: by token, I meant ratelimit bucket token, not an api token. I see that api-gateway based APIs also send bucket size in the header, I will add some code which tries to use it otherwise having a fallback value

pkg/networking/middleware/retry_middleware.go
PeterSchafer: Got it! I think it should behave the same, I don't see a reason to differentiate the cases. And eventually this will keep the changes small 😄

pkg/networking/middleware/retry_middleware.go
```diff
@@ -394,3 +411,78 @@ func rateLimitRetryDelay(res *http.Response) time.Duration {
...
+type rateLimitPolicy struct {
+	limit         int
+	windowSeconds int
+}
+
+// parseRateLimitPolicies parses the IETF RateLimit-Limit draft format, e.g.
+// `160, 160;w=1;name="...", 1620;w=60;name="...", 97200;w=3600;name="..."`.
+// The bare leading value (no `w=`) has no window attached and is skipped —
+// its unit can't be determined without one.
+func parseRateLimitPolicies(header string) []rateLimitPolicy {
```
PeterSchafer: Question: Do you think there is a big value in scaling the jitter based on the window? We just want to ensure that not all of request retry at the same time.

PeterSchafer: the jitter could also just be scaled based on the actual `rawDelay` value

robertolopezlopez (commented):
it looks good to me - what do you think @PeterSchafer

pkg/networking/middleware/retry_middleware.go
nick-y-snyk: Ai explanation, which seems legit to me:

rawDelay is max(Retry-After, X-RateLimit-Reset) — a countdown to reset, bounded by whichever window the server reports. It's not a capacity signal.

  Per-second buckets always report rawDelay ≤1s whether the bucket is 5 req/s or 160 req/s — no way to tell low-capacity from high-capacity apart. Conversely, an
  hourly bucket can report anywhere from 0 to 3600s depending purely on when in the window you got throttled, independent of how much capacity it has. Same bucket,
  same capacity, wildly different rawDelay on different retries.

  Scaling jitter on rawDelay would inherit that noise — jitter size would track "how far into the window am I" rather than "how many concurrent retriers can this
  bucket safely absorb," which is the actual thing we're trying to size for. That's why this reads X-RateLimit-Limit for the rate instead: it's the only signal here
  that reflects capacity independent of reset timing.

pkg/networking/middleware/retry_middleware.go
nick-y-snyk: applied to both headers

pkg/networking/middleware/retry_middleware.go
robertolopezlopez: Reviewing this thread, second thoughts

I’m concerned that the capacity based jitter adds too much complexity to the retry middleware. The parser and bucket-selection heuristic are difficult to understand, verify, and maintain - while they still cannot estimate the number of concurrent clients. Thus they may infer the wrong exhausted bucket from rawDelay.

Why not using a fixed jitter window? Something closer to your initial approach. For example, wait the full Retry-After (or X-RateLimit-Reset-Delay as we have already) + random(0, fixedWindow)? That would shrink the diff plus we could choose the window from real evidence and revisit it later if necessary.

nick-y-snyk: I don't mind returning a static number. Still better than nothing

We can keep current thing and add more testing too

PeterSchafer: This sounds like a great proposal to me, following the KISS principle (Keep it simple stupid).

robertolopezlopez: > I don't mind returning a static number. Still better than nothing

You got my vote

PeterSchafer (approved):

danskmt:
@nick-y-snyk @PeterSchafer this PR got merged but the CLI did not receive its changes. It seems we are experiencing lots of timeouts for all platforms. I'll revert changes of this PR. Please reopen and ensure it is tested with CLI pipelines.

PR #651, opened 2026-07-08
feat(extension): [CLI-1653] dynamic, out-of-process extensions over gRPC

mattdolan-snyk:
### Description

Adds a way to load CLI **extensions dynamically** — from a prebuilt binary discovered at runtime — instead of only compiling them into the CLI. This unblocks developing an extension without rebuilding the host on every change, and lays the groundwork for downloadable / third-party extensions.

Extensions run **out of process** and talk to the host over gRPC via [`hashicorp/go-plugin`](https://github.com/hashicorp/go-plugin) (the Terraform/Vault model), so a crashing extension can't take down the CLI and no extension code is loaded into the host's address space. Despite the boundary, an extension is written like an in-process workflow: `extension.Handler` is a type alias for `workflow.Callback`, and it receives a real `workflow.InvocationContext`.

**What an extension gets:**
- **Config** — its declared flags, plus settings the host resolves.
- **Authenticated network** — `GetNetworkAccess().GetHttpClient()` returns a client the host authenticates for it, via a per-invocation loopback proxy. The user's credentials stay in the host process; the extension never holds them, and they can only reach the configured Snyk API (not arbitrary hosts).
- **Calling other workflows** — `GetEngine().Invoke(id, …)` runs a built-in or another extension's workflow on the host, in its full context.
- **Analytics** — `AddExtension*Value` / `AddError` feed the host's batch.

Developer loop: `go build` your extension → point a host at it (`app.WithExtensionPaths(...)`, or a `--plugin-path` flag / `internal_extension_paths` config key) → iterate with no host rebuild. Getting-started guide: `docs/dynamic-extensions.md`. Two runnable examples: `examples/extensions/whoami` (minimal, calls a host workflow) and `pkg/extension/testdata/exampleplugin` (config, authenticated API call, sibling invoke + analytics).

**Safety / scope:** this is the *trusted-code* posture — isolation for robustness, not a sandbox (a loaded binary runs with the user's OS privileges). A dynamic extension cannot silently override a bundled workflow (opt in with `internal_extension_allow_override`). The enterprise trust layer (signature/registry verification + a default-deny allowlist) is intentionally a later phase; the loader's dialer is the single chokepoint where verify-before-launch slots in.

**Blast radius:** the `workflow.Engine`, `Register`/`Invoke`, `InvocationContext`, and content-typed `Data` contracts are unchanged; built-in (compile-time) extensions behave exactly as before. The only core addition is `workflow.ResolveInvokeOptions`, a helper that lets an `Invoke` be forwarded across the boundary. Protobuf codegen is wired into `go generate` / `make generate`.

Design + internals: `docs/dynamic-extensions-design.md` · diagrams: `docs/dynamic-extensions-architecture.md`.

#### Architecture (blue = pre-existing, green = added here)

```mermaid
flowchart TB
    classDef existing fill:#dbeafe,stroke:#3b82f6,color:#0b1f44;
    classDef new fill:#d4f8d4,stroke:#2ea043,color:#06210d;

    subgraph HOST["Host CLI process"]
        APP["app.CreateAppEngineWithOptions"]:::existing
        ENG["workflow.Engine<br/>registry + Invoke dispatch"]:::existing
        LWF["localworkflows.Init<br/>built-in extensions (compile-time import)"]:::existing
        BW["built-in workflow callback<br/>e.g. flw://output"]:::existing
        IC["InvocationContext<br/>Config · NetworkAccess · Analytics · UI · Logger"]:::existing

        LOADER["extension.Loader<br/>(a workflow.ExtensionInit)"]:::new
        PROXY["proxy callback<br/>1 per discovered workflow"]:::new
        AP["auth proxy<br/>loopback, option C"]:::new
        HCB["HostCallback server<br/>sibling Invoke · analytics"]:::new
    end

    subgraph EXT["Extension process — separate binary (subprocess)"]
        SERVE["extension.Serve"]:::new
        EH["Handler = workflow.Callback"]:::new
        PIC["plugin InvocationContext<br/>remoteEngine · remoteAnalytics · NetworkAccess"]:::new
    end

    API["Snyk API"]:::existing

    APP --> ENG
    APP --> LWF
    LWF -. "AddExtensionInitializer / Register" .-> ENG
    ENG --> BW
    BW --> IC

    APP --> LOADER
    LOADER -. "launch + Discover, then Register proxy" .-> ENG
    ENG -->|"Invoke(id)"| PROXY
    PROXY -->|"gRPC: Execute<br/>(config snapshot, input,<br/>proxy URL+secret, broker id)"| SERVE
    SERVE --> EH --> PIC

    PIC -->|"plain HTTP + secret"| AP
    AP -->|"authenticated HTTPS<br/>(host injects credentials)"| API

    PIC -->|"gRPC broker (bidirectional)"| HCB
    HCB -->|"Engine.Invoke(sibling)"| ENG
    HCB -->|"record"| IC
```

### Checklist

- [x] Tests added and all succeed — TDD throughout; `go test ./pkg/extension/... ./pkg/app/... ./pkg/workflow/...` passes, including subprocess end-to-end tests (load-without-rebuild, authenticated API call with host-injected token, sibling invoke + analytics over the broker) and a 24-way concurrency test that passes under `-race`. (Full-repo `make test` not run here.)
- [x] Regenerated mocks, etc. (`make generate`) — protobuf now regenerates via `go generate` (buf + protoc plugins pinned in the `tool` block); committed files were regenerated through it.
- [ ] Linted (`make lint`) — `gofmt`, `goimports`, and `go vet` are clean; `golangci-lint` needs a CI run (its binary targeted an older Go than this module locally).
- [ ] Test your changes work for the CLI — not yet exercised against the CLI repo.

---

> [!NOTE]
> **High Risk**
> Touches authentication (loopback proxy, credential export rules) and cross-process invocation of host workflows—security-sensitive paths with broad new surface area, though heavily tested.
> 
> **Overview**
> Adds **dynamic extensions**: standalone binaries loaded at runtime via `hashicorp/go-plugin` and gRPC, registered on the existing `workflow.Engine` like built-ins—no host rebuild for extension iteration.
> 
> New **`pkg/extension`** covers discovery/`Execute`, a per-invocation **loopback auth proxy** so extensions call the Snyk API with host-injected credentials (not raw tokens), and **HostCallback** bridging sibling `Invoke`, analytics, and config overrides with guards (declared-flag-only overrides, credential key blocking, invocation depth cap).
> 
> **`app`** wires `WithExtensionPaths`, reads paths at `Init()` (post-flag-parse), dedupes paths, and **`CreateAppEngineWithCloser`** to tear down subprocesses. **`workflow.ResolveInvokeOptions`** is the small core helper for forwarding `Invoke` options across the boundary.
> 
> Docs (`dynamic-extensions*.md`), **`make build-examples`**, and the **`whoami`** example round out the developer loop; dependencies add gRPC/protobuf/buf tooling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c80700511d4c03aec2a0b9958f20df811c7884ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/extension/plugin.go
```diff
@@ -0,0 +1,134 @@
...
+		NetworkProxyUrl:   req.networkProxyURL,
+		NetworkProxyToken: req.networkProxyToken,
+	}
+
+	// Serve the HostCallback service for this invocation on a broker stream the
+	// extension can dial back into.
+	if c.broker != nil && req.invocation != nil {
+		brokerID := c.broker.NextId()
+		invocation := req.invocation
+		go c.broker.AcceptAndServe(brokerID, func(opts []grpc.ServerOption) *grpc.Server {
```
basti-snyk, line 118: **Should-fix (concurrency / resource leak):** The host launches \`go c.broker.AcceptAndServe(brokerID, …)\` on every invocation with a non-nil `invocation` (and `makeProxy` always sets it), but the plugin only dials the broker at `serve.go:124` — *after* three early returns (`NotFound`, input-decode `InvalidArgument` at `serve.go:109`, identifier-parse `InvalidArgument` at `serve.go:114`). On any of those error paths the plugin returns without dialing, so this `AcceptAndServe` goroutine + its `grpc.Server` block until the plugin client is killed. Ordinary bad input reaches the decode path, so a long-lived host (snyk-ls) leaks one goroutine + listener per errored Execute for the session lifetime. Bound the accept to the Execute RPC (per-call context / stop the broker server when `Execute` returns) rather than relying on the plugin to always dial.\n\n— AI review

pkg/app/app.go
```diff
@@ -425,6 +426,21 @@ func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
 	}
 
 	engine.AddExtensionInitializer(localworkflows.Init)
+
+	// Load any out-of-process extension binaries configured for this engine.
+	// This replaces the previous "scan here for extension binaries" placeholder:
+	// extensions run as gRPC subprocesses and are registered like built-ins.
+	if config != nil {
+		if paths := config.GetStringSlice(extension.ConfigurationKeyPaths); len(paths) > 0 {
+			loader := extension.NewLoader(
```
basti-snyk: **Should-fix (lifecycle / subprocess leak):** `CreateAppEngineWithOptions` builds `loader := extension.NewLoader(...)`, registers `loader.Init`, then discards the reference. `Loader.Close()` (loader.go:231) — the only path that runs the accumulated `client.Kill` cleanups — is unreachable in production (no non-test caller). A long-lived host that recreates engines per session spawns one `exec.Command` subprocess per extension at `Init()` and never reaps them; orphaned processes + gRPC conns accumulate for the host lifetime. Either return/store the loader and call `Close()` on shutdown (or register `plugin.CleanupClients`), or delete `Close()` if you truly rely only on go-plugin parent-death detection so it is not a false promise.\n\n— AI review

pkg/app/app.go
```diff
@@ -425,6 +426,21 @@ func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
 	}
 
 	engine.AddExtensionInitializer(localworkflows.Init)
+
+	// Load any out-of-process extension binaries configured for this engine.
+	// This replaces the previous "scan here for extension binaries" placeholder:
+	// extensions run as gRPC subprocesses and are registered like built-ins.
+	if config != nil {
+		if paths := config.GetStringSlice(extension.ConfigurationKeyPaths); len(paths) > 0 {
```
basti-snyk: **Should-fix (regression — CLI flag silently ignored):** Loader registration is gated on `config.GetStringSlice(ConfigurationKeyPaths)` evaluated at engine *construction*. The documented usage (loader.go:20-24) is a repeatable `--plugin-path` flag bound to this key, but CLIs normally create the engine, then parse flags, then `Init()`. At construction the flag is unparsed → slice empty → `loader.Init` is never registered, and even after flags parse nothing loads. The previous placeholder lived inside `Init()` (late); this moved the decision early. Register a loader initializer unconditionally and read the paths when it runs, so flag-supplied paths take effect. (Only env/opt values present at creation currently work.)\n\n— AI review

pkg/extension/loader.go
```diff
@@ -0,0 +1,275 @@
...
+	return func(invocation workflow.InvocationContext, input []workflow.Data) ([]workflow.Data, error) {
+		config := invocation.GetConfiguration()
+
+		req := executeRequest{
+			identifier: spec.GetIdentifier(),
+			config:     make(map[string]string, len(spec.GetFlags())),
+			invocation: invocation,
+		}
+		for _, flag := range spec.GetFlags() {
+			req.config[flag.GetName()] = config.GetString(flag.GetName())
```
basti-snyk: **Should-fix (credential-isolation bypass — undermines the core premise):** The host forwards config to the extension by iterating over flag names the *extension itself declared* via `Discover`, with no denylist: `req.config[flag.GetName()] = config.GetString(flag.GetName())`. Sensitive keys live in the same config namespace (`snyk_token`, `api`, `snyk_cfg_api`, `snyk_oauth_token` — see pkg/configuration/constants.go). A malicious/compromised extension that declares a flag named `snyk_token`/`api`/`snyk_oauth_token` makes the host read the real credential and serialize it into the `ExecuteRequest` sent to the extension process — defeating the "extension never holds credentials" guarantee the auth-proxy (option C) exists to provide. Add a reserved/sensitive-key denylist (or forbid extension flag names that collide with host config keys) on this loop.\n\n— AI review

pkg/extension/hostcallback.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "parsing identifier %q: %v", req.GetIdentifier(), err)
+	}
+
+	input, err := msgsToDataSlice(req.GetInput(), s.config)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "decoding input: %v", err)
+	}
+
+	output, err := s.engine.Invoke(id, workflow.WithInput(input))
```
basti-snyk: **Should-fix (unbounded cross-process recursion + trust boundary):** `Invoke` calls `s.engine.Invoke(id, …)` for any identifier the extension supplies, with no allowlist and no self-invocation/depth guard. Two issues: (1) an extension can drive *any* registered host workflow in the host's full context (in-process host workflows use the real credentials directly), not just siblings as the comments imply; (2) if the extension invokes its own `flw://x`, the host re-enters the same proxy → new auth proxy + broker stream + Execute → recurses, amplifying subprocesses/goroutines/listeners across the boundary until FD/goroutine/OOM exhaustion (vs a bounded in-process stack). Scope to an allowlist/capability grant and add a depth or cycle guard. Separately, this host-side broker gRPC server has no panic-recovery interceptor, so a host workflow that panics on plugin-supplied input crashes the whole host — add a `recover()` unary interceptor.\n\n— AI review

pkg/extension/loader.go
```diff
@@ -0,0 +1,275 @@
...
+func (l *Loader) loadOne(engine workflow.Engine, path string) error {
+	conn, cleanup, err := l.dialer(context.Background(), path)
+	if err != nil {
+		return fmt.Errorf("launching extension: %w", err)
+	}
+	l.mu.Lock()
+	l.cleanups = append(l.cleanups, cleanup)
+	l.mu.Unlock()
+
+	specs, err := conn.Discover(context.Background())
```
basti-snyk: **Should-fix (violates the "broken extension must never block CLI startup" invariant, loader.go:105-107):** `loadOne` dials and discovers with `context.Background()` (`l.dialer(context.Background(), …)` at :118 and `conn.Discover(context.Background())` here), and `grpcDialer` ignores the passed context. go-plugin bounds only the *handshake*; once a plugin handshakes and then hangs inside `Discover`, this call has no deadline and `Loader.Init` blocks the CLI indefinitely — exactly the failure the stated invariant forbids. Add a bounded context and propagate a real ctx into the dialer.\n\n— AI review

pkg/extension/invocationcontext.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *pluginInvocationContext) GetLogger() *log.Logger                     { return c.logger }
+func (c *pluginInvocationContext) GetEnhancedLogger() *zerolog.Logger         { return c.zlogger }
+func (c *pluginInvocationContext) GetUserInterface() ui.UserInterface         { return c.ui }
+
+// GetEngine returns an engine whose Invoke runs sibling workflows on the host,
+// or nil when the host did not offer callbacks for this invocation.
+func (c *pluginInvocationContext) GetEngine() workflow.Engine { return c.engine }
+
+// GetRuntimeInfo returns nil in this phase; runtime info is not yet bridged.
+func (c *pluginInvocationContext) GetRuntimeInfo() runtimeinfo.RuntimeInfo { return nil }
```
basti-snyk: **Should-fix (nil deref / interface contract violation):** `GetRuntimeInfo()` unconditionally returns `nil`. Host consumers deref it without a nil check — e.g. report_analytics_workflow.go does `invocationCtx.GetRuntimeInfo().GetName()`/`.GetVersion()`. In-process this is populated; run the same workflow as an extension and it panics inside the subprocess. Populate RuntimeInfo from the request (or from host-forwarded values) instead of returning nil.\n\n— AI review

pkg/extension/proxies.go
```diff
@@ -0,0 +1,135 @@
...
+	return e.invoke(id, input)
+}
+
+func (e *remoteEngine) InvokeWithInput(id workflow.Identifier, input []workflow.Data) ([]workflow.Data, error) {
+	return e.invoke(id, input)
+}
+
+// InvokeWithConfig forwards the invocation; per-invocation config overrides are
+// not propagated across the boundary (the sibling runs with the host's config).
+func (e *remoteEngine) InvokeWithConfig(id workflow.Identifier, _ configuration.Configuration) ([]workflow.Data, error) {
```
basti-snyk: **Should-fix (silent wrong behaviour for sibling invokes):** `InvokeWithConfig(id, _ configuration.Configuration)` drops its config argument and calls `invoke(id, nil)`; `InvokeRequest` carries no config, so any per-invocation override is silently discarded. An extension mirroring the in-repo `cfg.Set(FLAG_EXPERIMENTAL, true); GetEngine().InvokeWithConfig(WORKFLOWID_WHOAMI, cfg)` pattern gets the override dropped → sibling fails ("experimental feature") or returns wrong results with no signal. Relatedly `GetWorkflow` always returns `(nil,false)` and `GetWorkflows` returns `nil` (proxies.go:120-121), so guard-then-invoke patterns silently skip valid host workflows. Forward config across the boundary and back the lookup methods with the host engine.\n\n— AI review

pkg/extension/serve.go
```diff
@@ -0,0 +1,208 @@
...
+
+	input, err := msgsToDataSlice(req.GetInput(), config)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "decoding input: %v", err)
+	}
+
+	id, err := url.Parse(req.GetIdentifier())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "parsing identifier: %v", err)
+	}
```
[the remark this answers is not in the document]
mattdolan-snyk: Fixed in a84dd20f1ace59f6ee2af9ea51ff4b94fb935b21: the broker dial now happens first thing in `Execute`, before any error path (unknown workflow, bad input, bad identifier). Every invocation dials back regardless of outcome, so the host's `AcceptAndServe` goroutine always terminates.

pkg/app/options.go
```diff
@@ -49,3 +51,18 @@ func WithRuntimeInfo(ri runtimeinfo.RuntimeInfo) Opts {
...
+// CLI flag or environment variable bound to that key.
+func WithExtensionPaths(paths ...string) Opts {
+	return func(engine workflow.Engine) {
+		config := engine.GetConfiguration()
+		if config == nil {
+			return
+		}
+		existing := config.GetStringSlice(extension.ConfigurationKeyPaths)
+		config.Set(extension.ConfigurationKeyPaths, append(existing, paths...))
+	}
```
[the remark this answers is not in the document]
mattdolan-snyk: Fixed in d005d9d588866776bfe00bb3aac40244e4c7c6db: `WithConfiguration` now merges forward any extension paths already recorded on the engine's prior configuration before swapping it in, so `WithExtensionPaths` survives regardless of option order.

pkg/extension/plugin.go
mattdolan-snyk: Good catch on the mechanism, though the specific leak path here is already closed: the extension's own `Execute` (serve.go) now dials the host broker as the very first thing it does, before any error return (unknown workflow, bad input, bad identifier) — see the earlier fix in this PR at d674285. So a plugin that reaches `Execute` at all will always dial back.

The residual gap you're pointing at — the RPC to the extension failing/timing out *before* it ever reaches `Execute` (so nothing on the plugin side can dial) — is real and not addressed. Bounding `AcceptAndServe` to the surrounding `Execute` call's own lifetime (stop the broker server unconditionally once `c.client.Execute` returns, rather than trusting the plugin) would close that too, but touches go-plugin broker internals I'd want to test more carefully before landing. Leaving as a known residual gap for now rather than rushing it — happy to pick up as a follow-up if this needs closing sooner.

pkg/app/app.go
mattdolan-snyk: Fixed in 165f2fa8702d220cb99eaee7db49fa4010450973: the loader initializer now closes over its `*extension.Loader` instance, and `CreateAppEngineWithCloser` returns a close func wired to it so callers that create engines dynamically can reap the subprocesses on shutdown. `CreateAppEngineWithOptions` is unchanged for existing callers that don't need it.

pkg/app/app.go
mattdolan-snyk: Fixed in 165f2fa8702d220cb99eaee7db49fa4010450973 (same commit as the Close() fix, since both were the same root cause): the loader initializer is now registered unconditionally and reads `extension.ConfigurationKeyPaths` when it actually runs at `Init()` time, not at engine construction. Added a regression test (`TestCreateAppEngine_ExtensionPathSetAfterConstruction_StillLoads`) that sets the path after `CreateAppEngineWithOptions` returns, mirroring the real CLI startup order.

pkg/extension/loader.go
mattdolan-snyk: Confirmed and fixed in aab491a427bf6fa69ff82f4f29465764b6815145: `makeProxy` now checks each declared flag name against a denylist of reserved host credential keys (`snyk_token`, `snyk_oauth_token`, the legacy `api` token alias, OAuth `client-id`/`client-secret`) before exporting its value, and skips (with a warning log) any collision instead. Added `TestLoader_DoesNotExportSensitiveConfigValue` covering the exact scenario you described.

pkg/extension/hostcallback.go
mattdolan-snyk: Split into two fixes in 5ea06d2085f4fc8cc9a2a6d1f8f4b2a7e6827bd5:

1. **Recursion/amplification**: added a round-trip depth guard. Each hop is tracked via a context value threaded through `workflow.WithContext` on the host side (no proto change needed — every hop in the chain is host-process-local except the two gRPC calls, which already carry a context both ways). `Invoke` now rejects the call once depth exceeds a fixed max (10) instead of letting a self-referential chain run until the host exhausts file descriptors/goroutines/memory.
2. **Panic isolation**: added a `recover()` unary interceptor on the host-side HostCallback gRPC server, so a host workflow panicking on plugin-supplied input returns a gRPC error instead of crashing the host process.

On the broader point — extension can drive any registered host workflow, not just declared siblings — leaving that as-is rather than adding an allowlist: it matches how in-process workflows already call each other via `engine.Invoke` (no allowlist there either), and constraining it further felt like a separate design decision worth its own discussion rather than bundling into this fix. Happy to pick that up if you think it needs to be scoped down.

pkg/extension/loader.go
mattdolan-snyk: Fixed in aab491a427bf6fa69ff82f4f29465764b6815145 (same commit as the credential-denylist fix): `loadOne` now bounds dial + `Discover` to a 10s timeout instead of `context.Background()`, and `grpcDialer` derives go-plugin's `StartTimeout` from that same deadline instead of ignoring the passed context.

pkg/extension/invocationcontext.go
mattdolan-snyk: Fixed in c92c5b84ad966267f360ada66eb9e9bdfa2a2fc1: `ExecuteRequest` now carries the host's runtime info name/version, and the plugin side wraps it into a real (possibly empty, but never nil) `runtimeinfo.RuntimeInfo` — matching in-process behavior exactly when the host has one set, and failing safe instead of panicking when it doesn't.

pkg/extension/proxies.go
mattdolan-snyk: The `InvokeWithConfig` dropped-config half was already fixed earlier in this PR (commit a84dd20 — config overrides are now diffed and forwarded across the boundary via `InvokeRequest.config`, applied by `hostCallbackServer.Invoke` on top of the host's own config).

The `GetWorkflow`/`GetWorkflows` stub half is still open — deliberately not doing it here. Backing it truthfully needs a new HostCallback RPC (there's no existing way to ask the host "is this identifier registered" or "list your workflows" across the boundary), which is a bigger, separate change. Tracking it as follow-up rather than rushing a proto addition into this pass.

pkg/extension/proxies.go
```diff
@@ -0,0 +1,163 @@
...
+// Invoke forwards the invocation to the host so the sibling workflow runs in
+// the host's full context. Config overrides supplied via workflow.WithConfig
+// are propagated (the host applies them on top of its own configuration).
+// workflow.WithContext and workflow.WithInstrumentationCollector are not
+// propagated: a Go context and an analytics.InstrumentationCollector cannot
+// cross the process boundary, so the sibling always runs with the host's own
+// context and reports through the host's analytics.
+func (e *remoteEngine) Invoke(id workflow.Identifier, opts ...workflow.EngineInvokeOption) ([]workflow.Data, error) {
+	config, input := workflow.ResolveInvokeOptions(e.config, opts...)
+	return e.invoke(id, input, configOverrides(e.config, config))
```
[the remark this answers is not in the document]
mattdolan-snyk: Good catch, this was a real regression from my own earlier config-forwarding fix in this PR. Fixed in dcb12823ac6dcf1592f1fbff5836b2e912ddb23c: `remoteEngine` now snapshots config's values once, before the extension's handler runs, and `configOverrides` diffs against that snapshot instead of the live (mutable) object — so mutating `invocation.GetConfiguration()` in place and then `Invoke`-ing with that same object is detected correctly. Added `TestRemoteEngine_InvokeForwardsInPlaceConfigMutation` covering exactly this pattern.

pkg/extension/proxies.go
```diff
@@ -0,0 +1,163 @@
...
+func (e *remoteEngine) InvokeWithInputAndConfig(id workflow.Identifier, input []workflow.Data, config configuration.Configuration) ([]workflow.Data, error) {
+	return e.invoke(id, input, configOverrides(e.config, config))
+}
+
+func (e *remoteEngine) GetConfiguration() configuration.Configuration { return e.config }
+func (e *remoteEngine) GetNetworkAccess() networking.NetworkAccess    { return e.network }
+func (e *remoteEngine) GetAnalytics() analytics.Analytics             { return e.stats }
+func (e *remoteEngine) GetLogger() *zerolog.Logger                    { return e.logger }
+func (e *remoteEngine) GetUserInterface() ui.UserInterface            { return e.ui }
+func (e *remoteEngine) GetRuntimeInfo() runtimeinfo.RuntimeInfo       { return nil }
```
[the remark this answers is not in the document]
mattdolan-snyk: Fixed in dcb12823ac6dcf1592f1fbff5836b2e912ddb23c — the earlier RuntimeInfo fix only covered `pluginInvocationContext.GetRuntimeInfo()`; `remoteEngine` (reached via `GetEngine()`) now carries the same forwarded `runtimeinfo.RuntimeInfo` and returns it instead of nil.

pkg/app/app.go
```diff
@@ -425,6 +426,21 @@ func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
...
+			loader := extension.NewLoader(
+				extension.WithPaths(paths...),
+				extension.WithLogger(engine.GetLogger()),
+				extension.WithAllowOverride(config.GetBool(extension.ConfigurationKeyAllowOverride)),
+			)
+			engine.AddExtensionInitializer(loader.Init)
+		}
```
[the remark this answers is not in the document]
mattdolan-snyk: Already fixed — this re-scanned the pre-fix commit. See 165f2fa8702d220cb99eaee7db49fa4010450973, which added `CreateAppEngineWithCloser`.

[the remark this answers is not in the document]
mattdolan-snyk: Already fixed in the same commit as above (165f2fa) — the loader initializer is now registered unconditionally and reads the paths lazily at `Init()` time.

pkg/extension/loader.go
```diff
@@ -0,0 +1,290 @@
...
+		config := invocation.GetConfiguration()
+
+		req := executeRequest{
+			identifier: spec.GetIdentifier(),
+			config:     make(map[string]string, len(spec.GetFlags())),
+			invocation: invocation,
+		}
+		for _, flag := range spec.GetFlags() {
+			req.config[flag.GetName()] = config.GetString(flag.GetName())
+		}
```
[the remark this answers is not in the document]
mattdolan-snyk: Already fixed — see aab491a427bf6fa69ff82f4f29465764b6815145, which denies extension flags colliding with reserved host credential keys.

pkg/extension/loader.go
```diff
@@ -0,0 +1,290 @@
...
+func (l *Loader) loadOne(engine workflow.Engine, path string) error {
+	conn, cleanup, err := l.dialer(context.Background(), path)
+	if err != nil {
+		return fmt.Errorf("launching extension: %w", err)
+	}
+	l.mu.Lock()
+	l.cleanups = append(l.cleanups, cleanup)
+	l.mu.Unlock()
+
+	specs, err := conn.Discover(context.Background())
```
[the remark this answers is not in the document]
mattdolan-snyk: Already fixed in the same commit as above (aab491a) — dial/Discover are now bounded to a 10s timeout instead of `context.Background()`.

pkg/extension/invocationcontext.go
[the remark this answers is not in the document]
mattdolan-snyk: Already fixed — see c92c5b84ad966267f360ada66eb9e9bdfa2a2fc1, RuntimeInfo is now mirrored across the boundary and never nil.

pkg/extension/plugin.go
```diff
@@ -0,0 +1,134 @@
...
+			s := grpc.NewServer(opts...)
+			extensionpb.RegisterHostCallbackServer(s, newHostCallbackServer(
+				invocation.GetEngine(),
+				invocation.GetAnalytics(),
+				invocation.GetConfiguration(),
+			))
+			return s
+		})
+		pbReq.BrokerId = brokerID
+	}
```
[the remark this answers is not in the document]
mattdolan-snyk: Same residual gap I flagged on the earlier version of this comment (d674285): the plugin-side dial-order fix means the extension always dials back once it reaches `Execute`, but if the RPC never reaches `Execute` at all (fails/times out earlier), nothing dials and this goroutine can still leak. Bounding `AcceptAndServe`'s server lifetime to the surrounding `Execute` call itself would close it, but I want to verify that against go-plugin's broker internals (does `Stop()` cleanly unblock an `Accept()` that never got a connection, without racing a legitimate late dial) before landing it. Tracking as a known follow-up rather than rushing it.

pkg/extension/hostcallback.go
```diff
@@ -0,0 +1,89 @@
...
+	invokeOpts := []workflow.EngineInvokeOption{workflow.WithInput(input)}
+	if overrides := req.GetConfig(); len(overrides) > 0 && s.config != nil {
+		cfg := s.config.Clone()
+		for key, value := range overrides {
+			cfg.Set(key, value)
+		}
+		invokeOpts = append(invokeOpts, workflow.WithConfig(cfg))
+	}
+
+	output, err := s.engine.Invoke(id, invokeOpts...)
```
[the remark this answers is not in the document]
mattdolan-snyk: Same as my reply on the earlier version of this comment (5ea06d2): the depth/recursion half is fixed (bounded to 10 round trips). The "can invoke any host workflow, not just siblings" half is left as-is deliberately — it matches how in-process workflows already call each other via `engine.Invoke` with no allowlist, so scoping extensions down further is a product/security policy decision I'd rather get explicit sign-off on than bake in unilaterally. Happy to pick it up once there's a decision on what the allowed surface should be.

pkg/extension/loader.go
```diff
@@ -0,0 +1,336 @@
...
+// again -- workflow registration skips the resulting identifier collisions,
+// but the orphaned processes keep running.
+func (l *Loader) Init(engine workflow.Engine) error {
+	l.mu.Lock()
+	if l.loaded {
+		l.mu.Unlock()
+		return nil
+	}
+	l.loaded = true
+	l.mu.Unlock()
```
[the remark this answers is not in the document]
mattdolan-snyk: Real gap, fixed in 549c76d: Close is now documented and enforced as terminal — a subsequent Init() returns an error instead of silently no-op'ing. Relaunching properly wasn't a safe option here: the engine may already hold proxy workflows registered under identifiers a second load would collide with (skipped by default unless allow-override is set), so a silent no-op was strictly worse than failing loudly.

pkg/app/options.go
```diff
@@ -24,8 +26,18 @@ func WithLogger(logger *log.Logger) Opts {
...
+// WithConfiguration replaces the engine's configuration. If an earlier option
+// (e.g. WithExtensionPaths) already recorded extension paths on the engine's
+// current configuration, they are carried forward onto the new one -- otherwise
+// they would be silently discarded when the configuration object is swapped.
 func WithConfiguration(config configuration.Configuration) Opts {
 	return func(engine workflow.Engine) {
+		if existing := engine.GetConfiguration(); existing != nil {
+			if paths := existing.GetStringSlice(extension.ConfigurationKeyPaths); len(paths) > 0 {
+				merged := append(config.GetStringSlice(extension.ConfigurationKeyPaths), paths...)
+				config.Set(extension.ConfigurationKeyPaths, merged)
```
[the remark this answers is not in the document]
mattdolan-snyk: Fixed in af50a53 — both WithExtensionPaths and WithConfiguration's carry-forward merge now dedupe via the existing utils.Dedupe helper.

pkg/extension/loader.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+		for _, flag := range spec.GetFlags() {
+			name := flag.GetName()
+			if sensitiveConfigKeys[strings.ToLower(name)] {
+				l.logger.Warn().Str("identifier", spec.GetIdentifier()).Str("flag", name).
+					Msg("extension declared a flag colliding with a reserved credential key; refusing to export its value")
+				continue
+			}
+			req.config[name] = config.GetString(name)
+		}
```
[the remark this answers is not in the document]
mattdolan-snyk: Real bug, confirmed by reading configuration.go's GetStringWithError: it only handles string/[]string underlying values, returns an error (swallowed to "") for everything else. Fixed in d61fdae — both call sites now dispatch on the actual value type instead of blind GetString: makeProxy uses the flag's declared type (bool/int/string), and configSnapshot/configOverrides use the raw runtime type of the config value. Added TestLoader_ExportsBoolAndIntFlagValues and TestConfigOverrides_DetectsBoolAndIntChanges.

pkg/extension/hostcallback.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	invokeOpts := []workflow.EngineInvokeOption{
+		workflow.WithInput(input),
+		workflow.WithContext(withInvocationDepth(ctx, nextDepth)),
+	}
+	if overrides := req.GetConfig(); len(overrides) > 0 && s.config != nil {
+		cfg := s.config.Clone()
+		for key, value := range overrides {
+			cfg.Set(key, value)
```
basti-snyk, line 115: **Critical (credential exfiltration — new since last review):** The inbound `HostCallback.Invoke` path applies extension-supplied config overrides onto a clone of the host's *real, token-bearing* configuration with **no key filtering** — `cfg := s.config.Clone()` then `cfg.Set(key, value)` for every entry in `req.GetConfig()`. This is the reverse of the outbound path (`loader.go` `sensitiveConfigKeys`), which deliberately refuses to export credential-colliding keys so the extension never holds the token.

The gap is exploitable: the auth middleware decides whether to attach the token by prefix-matching the request URL against `config.GetString(API_URL)` from *this same cloned config* (`pkg/networking/middleware/auth_header.go` `ShouldRequireAuthentication`). A malicious/compromised extension can call the callback with `config[API_URL]="https://attacker.example"` (or `AUTHENTICATION_ADDITIONAL_URLS`) and then invoke any host workflow that makes an authenticated API call — the host runs the sibling with the real credentials and sends them to the attacker's host. That defeats the auth-proxy isolation this whole feature is built around. Verified independently by three reviewers and traced to the middleware source.

Compounding: `s.engine.Invoke(id, ...)` (line 104) runs *any* workflow the extension names — there is still no allow-list restricting a plugin to a declared/sibling set (raised in the prior review).

Fix: host-side, reject/strip override keys in `sensitiveConfigKeys` **and** the network-routing keys (`configuration.API_URL`, `AUTHENTICATION_ADDITIONAL_URLS`, proxy / TLS-skip keys) before `cfg.Set`; ideally allow only keys declared by the invoked workflow's flags. Add a test asserting an extension-supplied `API_URL` override does not change the host sibling's effective endpoint/credentials.

— AI review

pkg/app/app.go
```diff
@@ -413,6 +414,21 @@ func CreateAppEngine() workflow.Engine {
 }
 
 func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
+	engine, _ := createAppEngine(opts...)
```
basti-snyk, line 418: **Should-fix (residual subprocess leak + unsynchronized loader):** The prior review flagged that `Loader.Close()` was unreachable; this revision adds a `closer` and `CreateAppEngineWithCloser`, but the most-used constructor `CreateAppEngineWithOptions` still discards it (`engine, _ := createAppEngine(opts...)`). Any existing caller (CLI, snyk-ls) that adopts `--plugin-path` through this constructor and creates engines per-session / repeatedly leaks the launched extension subprocesses, their gRPC conns, goroutines, and per-invocation auth-proxy listeners for the host lifetime, with no signal. Consider registering the loader's `Close` on an engine shutdown hook so the default path is safe, rather than relying on callers to switch to the closer variant.

Separately, the `loader` variable in `createAppEngine` is written by the extension-initializer closure (runs inside `engine.Init()`) and read by the returned `closer` with no synchronization — `Loader.mu` guards the struct internals but not this app-level pointer. A daemon that runs `closer` on a shutdown goroutine while `Init()` is still executing races and may observe a nil/partial `loader`, skipping `Close()` and leaking subprocesses.

— AI review

pkg/extension/hostcallback.go
mattdolan-snyk: Confirmed and fixed in 28f2a67 — you're right, this defeated the whole isolation design. hostCallbackServer.Invoke now only applies an override for a key the *invoked workflow itself* declared as configurable (via its ConfigurationOptions/flags at Register time), mirroring how the outbound (host-to-extension) direction already only exports a workflow's own declared FlagSpec keys. An override for API_URL, AUTHENTICATION_ADDITIONAL_URLS, or any other key the target workflow didn't register is refused outright — no fallback, no partial application. Added TestHostCallbackServer_RejectsConfigOverrideForUndeclaredKey covering exactly the API_URL exfiltration path you traced.

On the allow-list point (any workflow, no sibling restriction): still the same open design question from the prior review round — left as-is pending a product/security decision on the intended blast radius, not re-litigating that here.

pkg/extension/loader.go
```diff
@@ -0,0 +1,369 @@
...
+		return fmt.Errorf("launching extension: %w", err)
+	}
+	l.mu.Lock()
+	l.cleanups = append(l.cleanups, cleanup)
+	l.mu.Unlock()
+
+	specs, err := conn.Discover(ctx)
+	if err != nil {
+		return fmt.Errorf("discovering workflows: %w", err)
+	}
```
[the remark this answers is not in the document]
mattdolan-snyk: Fixed in deb4200 — the process cleanup now runs immediately when Discover fails, instead of waiting for Loader.Close(). Added TestLoader_KillsProcessWhenDiscoverFails.

pkg/app/app.go
mattdolan-snyk: Two separate points here:

- **CreateAppEngineWithOptions not getting the closer** — same as my reply on the previous round: by design, additive API, existing callers keep today's behavior. Not re-litigating.
- **Unsynchronized loader variable — real bug, fixed in c807005.** The extension initializer writes loader from inside engine.Init(); the closer reads it. Both were plain closured-variable accesses with no synchronization, so a shutdown handler calling closer() concurrently with a still-running Init() (e.g. slow-loading extension) could observe a torn/nil read and skip Close(), leaking the subprocess. Now guarded by a mutex. Added a race-detector regression test (TestCreateAppEngineWithCloser_ConcurrentInitAndCloseIsRaceFree) exercising Init() and closer() concurrently — clean under `go test -race`.

pkg/extension/hostcallback.go
```diff
@@ -0,0 +1,176 @@
...
+		// Only apply overrides for keys the invoked workflow itself declared
+		// as configurable (its own registered flags) -- never arbitrary host
+		// configuration keys. Without this, an extension could set
+		// configuration.API_URL or AUTHENTICATION_ADDITIONAL_URLS on the
+		// clone below: the auth middleware decides whether to attach the
+		// real credential by matching a request's URL against exactly those
+		// keys (see networking/middleware.ShouldRequireAuthentication), so an
+		// override reaching them would make the host attach the user's real
+		// token to a request destined for a host the extension chose.
+		allowed := invokableConfigKeys(s.engine, id)
```
basti-snyk, line 108: **Critical (credential redirect — the new inbound override filter is bypassable):** This revision adds `invokableConfigKeys` to filter extension-supplied config overrides, but it keys the defense on an **attacker-controlled allowlist**: the permitted keys are exactly the flags the *invoked workflow declared*, and a dynamically-loaded extension controls its own workflows' declared flags. `Discover` returns `FlagSpec`s → `specsToFlagSet` (serve.go) turns every spec into a pflag with **no rejection of reserved key names** → `registerWorkflow` registers it as-is. So a malicious extension can populate the allowlist with a sensitive routing key.

Exploit (self-invoke): the extension declares a workflow flag named `snyk_api` (= `configuration.API_URL`) and/or `internal_additional_auth_urls` (= `AUTHENTICATION_ADDITIONAL_URLS`), then calls `HostCallback.Invoke(id=<its own wf>, config={"snyk_api":"https://attacker.example"})`. `invokableConfigKeys` returns `{"snyk_api":true}`, the check here passes, `cfg.Set("snyk_api", …)` + `WithConfig(cfg)` poison the invocation, and `startAuthProxy` reads `API_URL` and forwards the request — with the host's **real token injected host-side** — to the attacker's host. This is exactly the "attach the user's real token to a host the extension chose" threat the comment above describes; the allowlist does not stop it.

Fix: gate the inbound path on an **unconditional denylist** (credentials *and* auth-routing keys: `snyk_token`, `snyk_oauth_token`, `snyk_api`/`API_URL`, `internal_additional_auth_urls`, `internal_auth_subdomain`, plus the full outbound `sensitiveConfigKeys` set), rejecting them regardless of what the workflow declared — and reject those reserved names in `specsToFlagSet`/`registerWorkflow` so they never enter any workflow's `ConfigurationOptions`. Note: the outbound `sensitiveConfigKeys` denylist (loader.go) currently omits the URL-routing keys, so it is no backstop. (The case-folding path is fine — both sides lowercase and viper normalizes keys — so consistency there is not the issue; the allowlist-poisoning is.)

— AI review

pkg/app/app.go
```diff
@@ -425,7 +442,61 @@ func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
...
+	// loaderMu guards loader itself: the initializer below writes it from
+	// inside engine.Init(), and the returned closer reads it -- potentially
+	// from a different goroutine (e.g. a shutdown handler racing a still-
+	// running Init()) -- so a plain closured variable isn't safe here.
+	var (
+		loaderMu sync.Mutex
+		loader   *extension.Loader
+	)
+	engine.AddExtensionInitializer(func(engine workflow.Engine) error {
+		loaderMu.Lock()
```
basti-snyk, line 466: **Should-fix (mutex makes the pointer race-free but does not serialize the Init-vs-close lifecycle it claims to guard):** The comment above says `loaderMu` defends against "a shutdown handler racing a still-running Init()", but the lock is dropped between the read at :467 and the store at :486, so check-then-act is not atomic and the leak windows remain:

- **Init-after-close leak:** `closer()` (:491) takes the lock, reads `loader == nil` (initializer hasn't stored yet), unlocks, closes nothing. The initializer then stores `newLoader` and runs `newLoader.Init(engine)` (:488), spawning extension subprocess(es) *after* shutdown — nothing ever calls `Close()` on them.
- **Double-init leak:** two concurrent initializer runs both read `existing == nil` before either stores, both `NewLoader` + `Init` (double-spawn), and only the last-stored loader is ever `Close()`d — the earlier one's subprocesses leak.

`go test -race` (and `TestCreateAppEngineWithCloser_ConcurrentInitAndCloseIsRaceFree`) verifies data-race freedom but cannot catch either logical leak, so the new test gives false confidence. Fix: hold `loaderMu` across the whole check-create-store, and add a `closed bool` under the same mutex so the initializer refuses to start a loader once `closer` has run (and `closer` closes whichever loader is set). The root-cause-correct home for the shutdown guard is `Loader.Init` itself. If Init and shutdown are in fact always single-threaded, drop the mutex and document that contract instead of half-implementing serialization. (Separately, the default constructor `CreateAppEngineWithOptions` still discards this closer — raised in the prior review — so the loader is never closed on that path regardless.)

— AI review

pkg/app/app.go
```diff
@@ -425,7 +442,61 @@ func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
...
+	//
+	// loaderMu guards loader itself: the initializer below writes it from
+	// inside engine.Init(), and the returned closer reads it -- potentially
+	// from a different goroutine (e.g. a shutdown handler racing a still-
+	// running Init()) -- so a plain closured variable isn't safe here.
+	var (
+		loaderMu sync.Mutex
+		loader   *extension.Loader
+	)
+	engine.AddExtensionInitializer(func(engine workflow.Engine) error {
```
danskmt, line 465: The inner function can be moved to a separate function, so this current method (createAppEngine) doesn't grow too much and keeps it easy to read

pkg/app/app.go
```diff
@@ -413,6 +415,21 @@ func CreateAppEngine() workflow.Engine {
 }
 
 func CreateAppEngineWithOptions(opts ...Opts) workflow.Engine {
+	engine, _ := createAppEngine(opts...)
+	return engine
+}
```
danskmt, line 420: Note that **createAppEngine** always creates that close function - you probably expect to use the close function here or return it in same way you do for **CreateAppEngineWithCloser**

PR #652, opened 2026-07-08
fix: escape regex and glob metacharacters in ignore path globs

danskmt:
### Description

Snyk Code file exclusions from `.snyk`, `.gitignore`, and `.dcignore` were silently ignored when the project's absolute path contained regex/glob metacharacters — most commonly parentheses, e.g. `C:\Users\...\OneDrive - Foobar (Team1)\...` or `C:\Program Files (x86)\...`.

`parseIgnoreRuleToGlobs` prepends the absolute base path to every rule, and the resulting glob is compiled into a regex by `sabhiram/go-gitignore`. The base path is not pattern data, but it was treated as such: `escapeSpecialGlobChars` only escaped `$`, so characters like `(`, `)`, `+`, `{`, `}`, `^`, `|`, and the glob wildcards `*`, `[`, `]` in the path were interpreted specially and matching silently failed — so nothing was excluded and directories like `node_modules` got scanned in full.

This change escapes the literal base path separately from the glob parts:

- `buildGlob` splits the joined path into the literal base and the glob remainder, escaping the base with the full set (including `*`, `[`, `]`) so it always matches literally, while the rule/`**` parts keep their wildcard meaning.
- `?` and `.` are intentionally not escaped: `go-gitignore` already treats `?` as literal and escapes `.` itself, so escaping them here would double-escape and break matching.

Reproduced independently of OS: a project on a plain path excludes `node_modules` (16 files bundled), while the same project under a `(...)` path did not (~28.9k files, matching a no-exclusion scan). After the fix, both exclude correctly.

### Checklist

- [x] Linted (`make lint`)

---

> [!NOTE]
> **Medium Risk**
> Changes core ignore-glob construction used by Snyk Code file filtering; incorrect escaping could over- or under-exclude files, though behavior is covered by new path metacharacter tests.
> 
> **Overview**
> Fixes **silent failure of `.gitignore` / `.snyk` / `.dcignore` exclusions** when the scan root’s absolute path contains characters that `go-gitignore` treats as regex or glob syntax (e.g. `Program Files (x86)`, `OneDrive - Foobar (Team1)`).
> 
> `parseIgnoreRuleToGlobs` now builds patterns via **`buildGlob`**, which escapes the **literal base directory** (including `*`, `[`, `]` in path segments) separately from the **rule / `**` suffix**, using generalized **`escapeChars`** instead of only escaping `$`. Wildcards in ignore rules are unchanged; `?` and `.` are still not escaped on the path to avoid double-escaping in the matcher.
> 
> Adds **`TestFileFilter_GetFilteredFiles_pathWithRegexMetaChars`** to assert `node_modules` stays excluded across many metacharacter-heavy parent folder names (CLI-1515).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfe1280270cb92f846b1f626d27e22dede635c2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -326,21 +326,60 @@ func parseIgnoreFile(content []byte, filePath string) []string {
 	return ignores
 }
 
-// escapeSpecialGlobChars escapes special characters that should be treated literally in glob patterns.
-// Special Characters to escape: $
-func escapeSpecialGlobChars(rule string) string {
+// regexMetaCharsToEscape are regex metacharacters that are not glob syntax; escaped in rules
+// so they match literally (e.g. "$" in a rule).
+var regexMetaCharsToEscape = map[byte]bool{
+	'$': true,
+	'(': true,
+	')': true,
+	'+': true,
+	'^': true,
+	'|': true,
+	'{': true,
+	'}': true,
+}
+
+// pathMetaCharsToEscape extends regexMetaCharsToEscape with the glob wildcards "*", "[" and
+// "]" for escaping the literal base path (which is not a pattern), e.g. "Program Files (x86)"
+// or a folder named "a*b". "?" and "." are omitted: the ignore matcher already treats them
+// literally / escapes them itself.
+var pathMetaCharsToEscape = map[byte]bool{
+	'$': true,
+	'(': true,
+	')': true,
+	'+': true,
+	'^': true,
+	'|': true,
+	'{': true,
+	'}': true,
+	'*': true,
+	'[': true,
+	']': true,
+}
+
+func escapeChars(s string, charsToEscape map[byte]bool) string {
 	var result strings.Builder
-	for i := 0; i < len(rule); i++ {
-		ch := rule[i]
-		switch ch {
-		case '$':
+	for i := 0; i < len(s); i++ {
+		ch := s[i]
+		if charsToEscape[ch] {
 			result.WriteByte('\\')
-			result.WriteByte(ch)
-		default:
-			result.WriteByte(ch)
 		}
+		result.WriteByte(ch)
 	}
 	return result.String()
+}
+
+// buildGlob joins the base path with glob parts, escaping the base path literally (so path
+// characters are never interpreted as patterns) while the glob parts keep their wildcards.
+func buildGlob(prefix, baseDir string, parts ...string) string {
+	base := filepath.ToSlash(filepath.Clean(baseDir))
+	joined := filepath.ToSlash(filepath.Join(append([]string{baseDir}, parts...)...))
+	if rest, ok := strings.CutPrefix(joined, base); ok {
+		return prefix + escapeChars(base, pathMetaCharsToEscape) + escapeChars(rest, regexMetaCharsToEscape)
+	}
+	// Fallback: base is not a prefix of the joined path (e.g. a rule with "../" escaped above
+	// it). Escape conservatively as a glob so at least regex metacharacters are handled.
+	return escapeChars(prefix+joined, regexMetaCharsToEscape)
 }
 
 // parseIgnoreRuleToGlobs contains the business logic to build glob patterns from a given ignore file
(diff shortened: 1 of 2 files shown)


pkg/utils/file_filter.go
```diff
@@ -326,23 +326,62 @@ func parseIgnoreFile(content []byte, filePath string) []string {
...
 	return result.String()
 }
 
+// buildGlob joins the base path with glob parts, escaping the base path literally (so path
+// characters are never interpreted as patterns) while the glob parts keep their wildcards.
+func buildGlob(prefix, baseDir string, parts ...string) string {
+	base := filepath.ToSlash(filepath.Clean(baseDir))
+	joined := filepath.ToSlash(filepath.Join(append([]string{baseDir}, parts...)...))
+	if rest, ok := strings.CutPrefix(joined, base); ok {
+		return prefix + escapeChars(base, pathMetaCharsToEscape) + escapeChars(rest, regexMetaCharsToEscape)
```
basti-snyk, line 378: **Should-fix — regression: `[^...]` negated character classes in ignore rules now under-match.**

`escapeChars(rest, regexMetaCharsToEscape)` escapes `^` everywhere in the rule portion, but `^` is only regex-metacharacter-meaningful at the start of a `[...]` class. A gitignore rule `foo[^x]` is compiled by `go-gitignore` into `.../foo[\^x]`, flipping a *negated* class (any char except `x`) into a *literal* class (`^` or `x`).

Confirmed empirically against the real matcher: with rule `foo[^x]`, path `.../fooa/file.js` was ignored on `main` and is **no longer ignored** after this change (`OLD MatchesPath=true`, `NEW=false`). Files the user intended to exclude get scanned. The old code (escaping only `$`) did not regress this.

Root cause is context-insensitive escaping of `^` in the pattern half. Consider not escaping `[ ] ^` inside the rule portion (they are valid gitignore/fnmatch class syntax), or handling `^` only outside character classes.

— AI review

pkg/utils/file_filter.go
```diff
@@ -326,23 +326,62 @@ func parseIgnoreFile(content []byte, filePath string) []string {
...
+	'|': true,
+	'{': true,
+	'}': true,
+}
+
+// pathMetaCharsToEscape extends regexMetaCharsToEscape with the glob wildcards "*", "[" and
+// "]" for escaping the literal base path (which is not a pattern), e.g. "Program Files (x86)"
+// or a folder named "a*b". "?" and "." are omitted: the ignore matcher already treats them
+// literally / escapes them itself.
+var pathMetaCharsToEscape = map[byte]bool{
```
basti-snyk, line 346: **Should-fix — backslash (`\`) is the one regex metacharacter still not escaped in the base path.**

Neither escape map contains `\`. A literal `\` in a POSIX base path (legal in a Unix dir name; on Windows `filepath.ToSlash` converts separators so this is Unix-only) flows unescaped into the matcher's regex. E.g. a base segment `a\b` compiles to regex `a\b`, where `\b` is a word boundary → the directory never matches → ignore rules under it silently stop excluding (the exact CLI-1515 failure class this PR targets). Worse, `a\$b` emits `a\\$b`, defeating the very `$` escaping added here. Fix is a safe one-liner: add `'\\': true` to `pathMetaCharsToEscape` (base only, not the rule map — a `\` in a rule is an author escape).

**Suggestion (DRY):** `pathMetaCharsToEscape` hand-duplicates all eight entries of `regexMetaCharsToEscape` plus three extras. Per the repo's DRY value, derive it from `regexMetaCharsToEscape` (copy + add `* [ ]`, and `\` per above) so the shared set has one source of truth.

— AI review

pkg/utils/file_filter.go
```diff
@@ -326,23 +326,62 @@ func parseIgnoreFile(content []byte, filePath string) []string {
...
+// characters are never interpreted as patterns) while the glob parts keep their wildcards.
+func buildGlob(prefix, baseDir string, parts ...string) string {
+	base := filepath.ToSlash(filepath.Clean(baseDir))
+	joined := filepath.ToSlash(filepath.Join(append([]string{baseDir}, parts...)...))
+	if rest, ok := strings.CutPrefix(joined, base); ok {
+		return prefix + escapeChars(base, pathMetaCharsToEscape) + escapeChars(rest, regexMetaCharsToEscape)
+	}
+	// Fallback: base is not a prefix of the joined path (e.g. a rule with "../" escaped above
+	// it). Escape conservatively as a glob so at least regex metacharacters are handled.
+	return escapeChars(prefix+joined, regexMetaCharsToEscape)
```
basti-snyk, line 382: **Suggestion — fallback branch under-escapes the base path.**

When `base` is not a prefix of `joined` (a rule with enough `../` to climb above `baseDir`, or a relative `baseDir` that `Clean` reduces to `.`), the whole string is escaped with `regexMetaCharsToEscape` only, which omits `* [ ]`. If such a base contained `*`, `[`, or `]`, those would still be treated as wildcards/char-classes — the very CLI-1515 bug re-appearing on the fallback path. It's acknowledged in the comment and the shape is rare; escaping the fallback with `pathMetaCharsToEscape` (the fallback string is dominated by base-path content) would close it at no cost.

— AI review

danskmt:
Closing this PR in favor of https://github.com/snyk/go-application-framework/pull/653
(Using a bug task instead)

PR #653, opened 2026-07-09
fix: respect Snyk Code exclusions on paths with special characters

danskmt:
### Description

Snyk Code file exclusions from `.snyk`, `.gitignore`, and `.dcignore` were silently ignored when the project's absolute path contained regex/glob metacharacters — most commonly parentheses, e.g. `C:\Users\...\OneDrive - Company (Tenant)\...` or `C:\Program Files (x86)\...`.

`parseIgnoreRuleToGlobs` prepends the absolute base path to every rule, and the resulting glob is compiled into a regex by `sabhiram/go-gitignore`. The base path is not pattern data, but it was treated as such: `escapeSpecialGlobChars` only escaped `$`, so characters like `(`, `)`, `+`, `{`, `}`, `^`, `|`, and the glob wildcards `*`, `[`, `]` in the path were interpreted specially and matching silently failed — so nothing was excluded and directories like `node_modules` got scanned in full.

`buildGlob` now escapes the literal base path separately from the glob parts, so path characters always match literally while the rule/`**` parts keep their wildcard meaning. `?` and `.` are intentionally not escaped: `go-gitignore` already treats `?` as literal and escapes `.` itself.

Reproduced independently of OS: a project on a plain path excludes `node_modules` (~16 files bundled), while the same project under a `(...)` path did not (~28.9k files, matching a no-exclusion scan). After the fix, both exclude correctly.

Ticket: CLI-1648

### Checklist

- [x] Linted (`make lint`)

https://github.com/snyk/cli/pull/6980

---

> [!NOTE]
> **Medium Risk**
> Changes which files are included in Snyk Code scans; incorrect matching could over- or under-scan, though behavior is covered by new scenario tests.
> 
> **Overview**
> Fixes **silent failure of Snyk Code file exclusions** (`.gitignore`, `.snyk`, `.dcignore`) when the scan root’s absolute path contains characters that `go-gitignore` treats as regex/glob syntax—common on Windows paths like `Program Files (x86)` or OneDrive folder names with parentheses.
> 
> `parseIgnoreRuleToGlobs` no longer only escapes `$` on the full joined pattern. It uses **`buildGlob`** to escape the **literal base directory** aggressively (`*`, `[`, `]`, `^`, `\`, plus regex metacharacters) while still allowing **gitignore rule syntax** such as negated classes (`cache[^S]`) on the rule suffix. A conservative fallback applies when a rule’s `../` climbs above `baseDir`.
> 
> Adds broad tests for meta-character paths, nested ignore files, directory rules, and the `../` glob edge case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07a51b4f473f5c5bde3fe4bdb7bea48d2ade7167. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"maps"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -326,36 +327,30 @@ func parseIgnoreFile(content []byte, filePath string) []string {
 	return ignores
 }
 
-// regexMetaCharsToEscape are regex metacharacters that are not glob syntax; escaped in rules
-// so they match literally (e.g. "$" in a rule).
+// regexMetaCharsToEscape are regex metacharacters that are not glob syntax; escaped in ignore
+// rules so they match literally (e.g. "$" in a rule). "^", "[" and "]" are deliberately not
+// included: they are valid character-class syntax in gitignore rules (e.g. "foo[^x]").
 var regexMetaCharsToEscape = map[byte]bool{
 	'$': true,
 	'(': true,
 	')': true,
 	'+': true,
-	'^': true,
 	'|': true,
 	'{': true,
 	'}': true,
 }
 
-// pathMetaCharsToEscape extends regexMetaCharsToEscape with the glob wildcards "*", "[" and
-// "]" for escaping the literal base path (which is not a pattern), e.g. "Program Files (x86)"
-// or a folder named "a*b". "?" and "." are omitted: the ignore matcher already treats them
-// literally / escapes them itself.
-var pathMetaCharsToEscape = map[byte]bool{
-	'$': true,
-	'(': true,
-	')': true,
-	'+': true,
-	'^': true,
-	'|': true,
-	'{': true,
-	'}': true,
-	'*': true,
-	'[': true,
-	']': true,
-}
+// pathMetaCharsToEscape is used for the literal base path (which is not a pattern), so on top
+// of regexMetaCharsToEscape it also escapes the glob wildcards "*", "[", "]", the class anchor
+// "^", and "\". "?" and "." are omitted: the ignore matcher already treats them literally /
+// escapes them itself.
+var pathMetaCharsToEscape = func() map[byte]bool {
+	m := maps.Clone(regexMetaCharsToEscape)
+	for _, c := range []byte{'^', '*', '[', ']', '\\'} {
+		m[c] = true
+	}
+	return m
+}()
 
 func escapeChars(s string, charsToEscape map[byte]bool) string {
 	var result strings.Builder
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -326,21 +326,60 @@ func parseIgnoreFile(content []byte, filePath string) []string {
 	return ignores
 }
 
-// escapeSpecialGlobChars escapes special characters that should be treated literally in glob patterns.
-// Special Characters to escape: $
-func escapeSpecialGlobChars(rule string) string {
+// regexMetaCharsToEscape are regex metacharacters that are not glob syntax; escaped in rules
+// so they match literally (e.g. "$" in a rule).
+var regexMetaCharsToEscape = map[byte]bool{
+	'$': true,
+	'(': true,
+	')': true,
+	'+': true,
+	'^': true,
+	'|': true,
+	'{': true,
+	'}': true,
+}
+
+// pathMetaCharsToEscape extends regexMetaCharsToEscape with the glob wildcards "*", "[" and
+// "]" for escaping the literal base path (which is not a pattern), e.g. "Program Files (x86)"
+// or a folder named "a*b". "?" and "." are omitted: the ignore matcher already treats them
+// literally / escapes them itself.
+var pathMetaCharsToEscape = map[byte]bool{
+	'$': true,
+	'(': true,
+	')': true,
+	'+': true,
+	'^': true,
+	'|': true,
+	'{': true,
+	'}': true,
+	'*': true,
+	'[': true,
+	']': true,
+}
+
+func escapeChars(s string, charsToEscape map[byte]bool) string {
 	var result strings.Builder
-	for i := 0; i < len(rule); i++ {
-		ch := rule[i]
-		switch ch {
-		case '$':
+	for i := 0; i < len(s); i++ {
+		ch := s[i]
+		if charsToEscape[ch] {
 			result.WriteByte('\\')
-			result.WriteByte(ch)
-		default:
-			result.WriteByte(ch)
 		}
+		result.WriteByte(ch)
 	}
 	return result.String()
+}
+
+// buildGlob joins the base path with glob parts, escaping the base path literally (so path
+// characters are never interpreted as patterns) while the glob parts keep their wildcards.
+func buildGlob(prefix, baseDir string, parts ...string) string {
+	base := filepath.ToSlash(filepath.Clean(baseDir))
+	joined := filepath.ToSlash(filepath.Join(append([]string{baseDir}, parts...)...))
+	if rest, ok := strings.CutPrefix(joined, base); ok {
+		return prefix + escapeChars(base, pathMetaCharsToEscape) + escapeChars(rest, regexMetaCharsToEscape)
+	}
+	// Fallback: base is not a prefix of the joined path (e.g. a rule with "../" escaped above
+	// it). Escape conservatively as a glob so at least regex metacharacters are handled.
+	return escapeChars(prefix+joined, regexMetaCharsToEscape)
 }
 
 // parseIgnoreRuleToGlobs contains the business logic to build glob patterns from a given ignore file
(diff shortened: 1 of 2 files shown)


pkg/utils/file_filter.go
```diff
@@ -326,23 +326,62 @@ func parseIgnoreFile(content []byte, filePath string) []string {
...
+	'|': true,
+	'{': true,
+	'}': true,
+}
+
+// pathMetaCharsToEscape extends regexMetaCharsToEscape with the glob wildcards "*", "[" and
+// "]" for escaping the literal base path (which is not a pattern), e.g. "Program Files (x86)"
+// or a folder named "a*b". "?" and "." are omitted: the ignore matcher already treats them
+// literally / escapes them itself.
+var pathMetaCharsToEscape = map[byte]bool{
```
basti-snyk: Suggestion: `pathMetaCharsToEscape` duplicates all 8 entries of `regexMetaCharsToEscape` verbatim, then adds `* [ ]`. The two maps can drift — a future regex metachar added to one but not the other would silently change base-path escaping. Consider deriving `pathMetaCharsToEscape` from `regexMetaCharsToEscape` (copy, then add the three glob chars) so the shared set lives in one place.

— AI review

pkg/utils/file_filter.go
```diff
@@ -326,23 +326,62 @@ func parseIgnoreFile(content []byte, filePath string) []string {
...
+	'^': true,
+	'|': true,
+	'{': true,
+	'}': true,
+	'*': true,
+	'[': true,
+	']': true,
+}
+
+func escapeChars(s string, charsToEscape map[byte]bool) string {
```
basti-snyk, line 355: Should fix: a literal backslash `\` in a directory name is not escaped. On Linux/macOS `\` is a legal filename char and `filepath.ToSlash` is a no-op, so a dir named `a\b` reaches the matcher unescaped; `sabhiram/go-gitignore` + Go RE2 then read `\b` as a word boundary (or an invalid `\c` fails to compile → the rule is silently dropped), losing the exclusion. This is the same bug class this PR fixes, left uncovered — empirically confirmed that `MatchesPath` returns false for a path under `a\b`. Consider adding backslash to `pathMetaCharsToEscape` (ordering is safe since `escapeChars` only prepends). Narrow non-Windows edge, noting because it belongs in this function.

— AI review

pkg/utils/file_filter.go
```diff
@@ -326,23 +326,62 @@ func parseIgnoreFile(content []byte, filePath string) []string {
...
+// characters are never interpreted as patterns) while the glob parts keep their wildcards.
+func buildGlob(prefix, baseDir string, parts ...string) string {
+	base := filepath.ToSlash(filepath.Clean(baseDir))
+	joined := filepath.ToSlash(filepath.Join(append([]string{baseDir}, parts...)...))
+	if rest, ok := strings.CutPrefix(joined, base); ok {
+		return prefix + escapeChars(base, pathMetaCharsToEscape) + escapeChars(rest, regexMetaCharsToEscape)
+	}
+	// Fallback: base is not a prefix of the joined path (e.g. a rule with "../" escaped above
+	// it). Escape conservatively as a glob so at least regex metacharacters are handled.
+	return escapeChars(prefix+joined, regexMetaCharsToEscape)
```
basti-snyk, line 377: Suggestion: the fallback branch escapes with `regexMetaCharsToEscape`, which omits `*`, `[`, `]`. A base path containing `a*b` or `a[b]c` combined with a rule that climbs above `baseDir` via `../` would re-expose the exact metachar-in-path bug this PR fixes. Reachability is very low (needs `../`-heavy rules surviving `filepath.Join` cleaning) and the branch has no test coverage. Consider escaping the fallback with `pathMetaCharsToEscape` for consistency, or add a climbing-rule test case.

— AI review

robertolopezlopez:
suggestion (long term fix): you may leverage `utils.ToRelativeUnixPath()` (from `path.go`) . This way you can drop the `go-gitignore` requirement, which is enforcing regex (the root cause of this bug). 

For example in `GetFilteredFiles()`:

change:

`if !globPatternMatcher.MatchesPath(f) {` (line 131)

to:

```
rel, err := ToRelativeUnixPath(fw.path, f)
if err != nil { /* handle the error */ }
if !globPatternMatcher.MatchesPath(rel) {
```

danskmt:
> suggestion (long term fix): you may leverage `utils.ToRelativeUnixPath()` (from `path.go`) . This way you can drop the `go-gitignore` requirement, which is enforcing regex (the root cause of this bug).
> 
> For example in `GetFilteredFiles()`:
> 
> change:
> 
> `if !globPatternMatcher.MatchesPath(f) {` (line 131)
> 
> to:
> 
> ```
> rel, err := ToRelativeUnixPath(fw.path, f)
> if err != nil { /* handle the error */ }
> if !globPatternMatcher.MatchesPath(rel) {
> ```

I added some more test cases where using this strategy would fail (nested patterns)

pkg/utils/file_filter.go
```diff
@@ -326,23 +327,56 @@ func parseIgnoreFile(content []byte, filePath string) []string {
 	return ignores
 }
 
-// escapeSpecialGlobChars escapes special characters that should be treated literally in glob patterns.
-// Special Characters to escape: $
-func escapeSpecialGlobChars(rule string) string {
+// regexMetaCharsToEscape are regex metacharacters that are not glob syntax; escaped in ignore
+// rules so they match literally (e.g. "$" in a rule). "^", "[" and "]" are deliberately not
+// included: they are valid character-class syntax in gitignore rules (e.g. "foo[^x]").
+var regexMetaCharsToEscape = map[byte]bool{
```
PeterSchafer, line 333: Suggestion: Can we use this to safely escape the string?
[regexp package - regexp - Go Packages](https://pkg.go.dev/regexp#QuoteMeta)

robertolopezlopez (commented):
## Review summary

**Verdict: Approve** — correct root-cause fix with strong test coverage; one low-reachability edge case worth acknowledging.

### Root cause & approach
The diagnosis is right: `parseIgnoreRuleToGlobs` prepends the absolute scan root into every glob, and `sabhiram/go-gitignore` compiles that string to regex. Path characters like `( ) * [ ] ^` were treated as pattern syntax, so exclusions silently failed — matching CLI-1648 / customer paths (`OneDrive - Foobar (Team1)`, `Program Files (x86)`).

`buildGlob` is the right split:
- **Base path** → `pathMetaCharsToEscape` (literal filesystem segment)
- **Rule suffix** → `regexMetaCharsToEscape` only, deliberately **not** escaping `^`, `[`, `]` so gitignore character classes (`cache[^S]`) keep working

The `da4ae565` follow-up commit that preserved negated classes was necessary — good catch.

### What looks good
- `maps.Clone(regexMetaCharsToEscape)` for the path map avoids drift between the two sets (addresses earlier review feedback).
- Backslash included in `pathMetaCharsToEscape` — relevant on Unix paths like `a\b`.
- Broad tests: per-metachar path matrix (with Windows-illegal skips), table-driven rule scenarios (nested ignores, scoped rules, directory rules), and a unit test for the `../` fallback glob shape.
- CI green including **Windows tests** — important for this bug class.
- CLI companion PR linked ([snyk/cli#6980](https://github.com/snyk/cli/pull/6980)).

### Non-blocking nits

1. **Fallback branch** (`buildGlob` when `CutPrefix` fails): still escapes with `regexMetaCharsToEscape`, so `*`, `[`, `]` in ancestor path segments could theoretically slip through if a `../`-heavy rule climbs above `baseDir` while the resolved path still contains filesystem metachars (e.g. `/tmp/a*b/...`). Reachability is very low and the primary customer path is covered; acceptable to ship, but a one-line comment documenting the limitation would help future readers.

2. **`regexp.QuoteMeta`** (per Peter's comment): not suitable for the full joined string — it would break rule character classes. The per-segment split here is the correct trade-off.

### Post-merge
- Refresh [CLI #6980](https://github.com/snyk/cli/pull/6980) to the merged GAF commit (currently pins `2b4c6a0dbdab`, head is `07a51b4f473f`).

### Risk
Medium impact surface (which files enter the Code bundle) but change is narrowly scoped and fail-safe direction is *more* exclusions applied, not fewer.

pkg/utils/file_filter.go
danskmt: It seems we cannot - the LLM response here is a good one to see why not:
image

robertolopezlopez: I asked cursor to craft this table:

| Target | QuoteMeta | Result |
| :--- | :--- | :--- |
| Path segment my (dir) | my \(dir\) | OK-ish |
| Path segment v1.0 | v1\.0 | Breaks — go-gitignore re-escapes . |
| Path segment a?b | a\?b | Breaks — double-escape on ? |
| Rule suffix /node_modules/ | \\/node_modules/\\ | Wildcards dead |
| Rule cache[^S] | cache\[\^S\] | Class dead |
| Rule price$tag | price\$tag | Works (same as escapeSpecialGlobChars) |

danskmt: @PeterSchafer and @robertolopezlopez it seems we have many gaps coming from go-gitignore, I added more info in a comment in [CLI-1410](https://snyksec.atlassian.net/browse/CLI-1410)

robertolopezlopez:
> > suggestion (long term fix): you may leverage `utils.ToRelativeUnixPath()` (from `path.go`) . This way you can drop the `go-gitignore` requirement, which is enforcing regex (the root cause of this bug).
> > For example in `GetFilteredFiles()`:
> > change:
> > `if !globPatternMatcher.MatchesPath(f) {` (line 131)
> > to:
> > ```
> > rel, err := ToRelativeUnixPath(fw.path, f)
> > if err != nil { /* handle the error */ }
> > if !globPatternMatcher.MatchesPath(rel) {
> > ```
> 
> I added some more test cases where using this strategy would fail (nested patterns)

@danskmt 

In my PR those tests pass ;-) https://github.com/snyk/go-application-framework/pull/654

pkg/utils/file_filter.go
PeterSchafer: @danskmt sorry this might be a misunderstanding, I'm not asking to escape all characters from the whole pattern, just from the base path. This is what is not a pattern

PeterSchafer: and sorry AI, I don't agree 😊

PeterSchafer (changes requested):

pkg/utils/file_filter.go
danskmt: @PeterSchafer you mean in line 373 to have this?
```go
return prefix + regexp.QuoteMeta(base) + escapeChars(rest, regexMetaCharsToEscape)
```

danskmt:
Closing this PR in favor of https://github.com/snyk/go-application-framework/pull/656

PR #656, opened 2026-07-13
fix: Code scan exclusions ignored when paths or ignore rules contain special characters

danskmt:
### Description

Code SAST scanning (via CLI and IDE) silently stops honoring `.gitignore` and `.snyk` exclude
rules when the project lives under a path that contains regex metacharacters — parentheses,
plus signs, dollar signs, braces, etc.

The root cause: `parseIgnoreRuleToGlobs` builds absolute glob patterns by concatenating the
base directory path with the ignore rule. The `go-gitignore` library internally compiles these
patterns to regexes, so characters like `(`, `)`, `$`, `+` in the path are interpreted as regex
syntax instead of literals, causing all exclusion rules to silently fail.

The fix escapes the base path with `regexp.QuoteMeta` before embedding it in patterns, and
escapes only the regex-significant characters in the rule portion (preserving glob wildcards).

**Affected paths example:** `C:\Users\…\OneDrive - Foo (789EBar)\Documents\Projects\…`

**Risk: Low** — the change only affects how the base directory is embedded in ignore patterns;
glob/gitignore syntax in the rule portion is preserved. All existing tests pass, plus new tests
covering 11 metacharacter variants and 5 ignore-rule edge cases.

### Checklist

- [X] Regenerated mocks, etc. (`make generate`)
- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI: https://github.com/snyk/cli/pull/6980

---

> [!NOTE]
> **Low Risk**
> Changes are confined to ignore glob construction in file filtering; scope is narrow though incorrect escaping could still affect which files are scanned.
> 
> **Overview**
> Fixes **CLI-1648**, where `.gitignore`, `.dcignore`, and `.snyk` excludes could be ignored when the project or ignore-file directory path contains characters that `go-gitignore` treats as regex syntax (e.g. parentheses in OneDrive folder names).
> 
> `parseIgnoreRuleToGlobs` now **quotes the base directory** with `regexp.QuoteMeta` (with selective un-escaping so dots and `?` in paths still match literally), **escapes additional regex metacharacters in the rule text** while keeping glob wildcards like `*` and `[^S]`, and builds patterns with **`joinGlob`** so Windows **UNC** and extended-length `//` prefixes are not collapsed by `path.Join`.
> 
> Adds a large regression suite: metacharacter scan roots, ignore-rule scenarios (including `.snyk`), UNC integration tests on Windows, and expanded `TestParseIgnoreRuleToGlobs` cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 790cd588fe030b2bee7491ca6afd96140fb5d4b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -328,19 +328,30 @@ func parseIgnoreFile(content []byte, filePath string) []string {
 	return ignores
 }
 
-// escapeSpecialGlobChars escapes special characters that should be treated literally in glob patterns.
-// Special Characters to escape: $
+// ruleRegexMetaChars are regex metacharacters that gitignore treats as literal, so they must be
+// escaped before the rule reaches the regex-based go-gitignore matcher (e.g. a folder literally
+// named "build (old)"). Glob syntax the matcher relies on is deliberately excluded: "*", "?",
+// "[", "]" (wildcards / character classes) and "^" (character-class negation, e.g. "cache[^S]").
+var ruleRegexMetaChars = map[byte]bool{
+	'$': true,
+	'(': true,
+	')': true,
+	'+': true,
+	'|': true,
+	'{': true,
+	'}': true,
+}
+
+// escapeSpecialGlobChars escapes regex metacharacters in an ignore rule that gitignore treats as
+// literal, so they match literally instead of being interpreted by go-gitignore's regex engine.
 func escapeSpecialGlobChars(rule string) string {
 	var result strings.Builder
 	for i := 0; i < len(rule); i++ {
 		ch := rule[i]
-		switch ch {
-		case '$':
+		if ruleRegexMetaChars[ch] {
 			result.WriteByte('\\')
-			result.WriteByte(ch)
-		default:
-			result.WriteByte(ch)
 		}
+		result.WriteByte(ch)
 	}
 	return result.String()
 }
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -6,6 +6,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/utils/file_filter.go
```diff
@@ -364,6 +365,7 @@ func parseIgnoreRuleToGlobs(rule string, filePath string, invalidRules []string)
 	const slash = "/"
 	const all = "**"
 	baseDir := filepath.ToSlash(filePath)
+	baseDir = regexp.QuoteMeta(baseDir)
```
danskmt, line 390: @PeterSchafer @robertolopezlopez on Windows this approach fails: https://app.circleci.com/pipelines/gh/snyk/go-application-framework/4169/workflows/cd1254c6-a719-464c-acda-9353afe95985/jobs/18548

regexp.QuoteMeta adds `\` before metacharacters (e.g. `\(`, `\$`). Then `filepath.Join` on Windows treats those `\` as path separators, and `filepath.ToSlash` converts them to `/`, destroying the escaping.

pkg/utils/file_filter_test.go
```diff
@@ -245,6 +245,155 @@ func TestFileFilter_GetFilteredFiles(t *testing.T) {
...
+		"c{d}",                      // braces (not a valid quantifier)
+		"backup{2}",                 // braces forming a valid regex quantifier
+		"a^b",                       // caret
+		"a|b",                       // pipe / alternation
+		"a$b",                       // dollar
+		"a*b",                       // glob wildcard in the path
+		"a[b]c",                     // glob character class in the path
+		// "a?b",                       // glob single-char wildcard in the path
+		"a\\b", // literal backslash (legal on unix, illegal on windows)
+	}
```
danskmt, line 267: @PeterSchafer @robertolopezlopez for the case being skipped: `a?b`
**regexp.QuoteMeta** escapes `?` to `\?`, then **go-gitignore** also processes `?`, so `\?` becomes `\\?` in the final regex, matching literal `\?` instead of literal `?`. This is the same case of double-escaping that we saw, but this time made by the library.

Image

robertolopezlopez: please add a note to the commented `a?b`, stating why (for the future dev)

danskmt: actually pushed a commit undoing the escaping so we avoid double escaping

robertolopezlopez (approved):

PR #659, opened 2026-07-17
fix: load snyk.json only from configstore path

octavian-snyk:
### Description

Noted that if a `snyk.json` file is present in the working directory and absent from the user’s `~/.config/configstore/` , snyk will use the file from the working directory to authenticate requests.

This PR ensures we have a single source of truth for the `snyk.json`.

Relevant ticket: [CLI-1124](https://snyksec.atlassian.net/browse/CLI-1124)
CLI PR: https://github.com/snyk/cli/pull/7008

### Checklist

- [x] Linted (`make lint`)

---

> [!NOTE]
> **Medium Risk**
> Changes where authentication and endpoint settings are sourced from disk; correct for security but may surprise users who relied on a project-local snyk.json.
> 
> **Overview**
> Stops the CLI from treating a **`snyk.json` in the current working directory** as configuration when loading defaults. **`readConfigFilesIntoViper`** no longer adds `"."` as a Viper config path, so file-based settings (including **`api`**, **`org`**, **`endpoint`**) come only from **`~/.config/configstore`** via **`determineBasePath()`**.
> 
> This closes the case where a repo-local **`snyk.json`** could supply credentials when nothing was in the user configstore. New tests cover **cwd-only** (values stay empty) and **configstore + cwd** (home configstore wins).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d9e0f2df7ee9172339ee1c210842886264bd56a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[CLI-1124]: https://snyksec.atlassian.net/browse/CLI-1124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -276,7 +276,6 @@ func readConfigFilesIntoViper(files []string, config *extendedViper) {
 	}
 
 	config.viper.AddConfigPath(configPath)
-	config.viper.AddConfigPath(".")
 
 	// read config files
 	//nolint:errcheck // breaking api change needed to fix this
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -1178,6 +1178,65 @@ func Test_ReloadConfig(t *testing.T) {
 	cleanupConfigstore(t)
 }
 
+func Test_Configuration_doesNotLoadSnykJsonFromWorkingDirectory(t *testing.T) {
+	fakeHome := t.TempDir()
+	projectDir := t.TempDir()
+
+	t.Setenv("HOME", fakeHome)
+	t.Setenv("USERPROFILE", fakeHome)
+
+	cwdConfig := filepath.Join(projectDir, "snyk.json")
+	err := os.WriteFile(cwdConfig, []byte(`{
+		"api": "local-dir-token",
+		"org": "dir-org",
+		"endpoint": "https://api.invalid/api"
+	}`), 0o600)
+	assert.NoError(t, err)
+
+	oldWd, err := os.Getwd()
+	assert.NoError(t, err)
+	assert.NoError(t, os.Chdir(projectDir))
+	t.Cleanup(func() {
+		assert.NoError(t, os.Chdir(oldWd))
+	})
+
+	config := New()
+
+	assert.Empty(t, config.GetString("api"))
+	assert.Empty(t, config.GetString("org"))
+	assert.Empty(t, config.GetString("endpoint"))
+}
+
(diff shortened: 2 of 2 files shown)


pkg/configuration/configuration.go
```diff
@@ -276,7 +276,6 @@ func readConfigFilesIntoViper(files []string, config *extendedViper) {
 	}
 
 	config.viper.AddConfigPath(configPath)
-	config.viper.AddConfigPath(".")
```
danskmt, line 279: what other implications this removal can have?

octavian-snyk: I am not noticing any other files besides `snyk` referenced for the config (https://github.com/snyk/go-application-framework/blob/d1936e1132a202f90c45c2b1272d3cb83bad5d96/pkg/configuration/configuration.go#L213) 

Also, because of https://github.com/snyk/go-application-framework/blob/d1936e1132a202f90c45c2b1272d3cb83bad5d96/pkg/configuration/configuration.go#L275 this function always seems to load only the last file from the array. I believe we could also remove the for loop.

danskmt: does this affect `.snyk` files? Should the behavior consider the current directory, same behavior as happens for `.gitignore`?

danskmt: also need to think:
1. What if there are customers using the current approach? This will be a breaking change for them.

I also think should contain more tests, i.e. not only look for snyk.json, but for others (.snyk, .gitignore, ...)

danskmt: I see the ticket is a bit old, so I wonder if the understanding there remains the same

octavian-snyk: .snyk and .gitignore shouldn't be affected. 😀

The config is managed by Viper while .snyk and .gitignore seem to be loaded here: https://github.com/snyk/go-application-framework/blob/d1936e1132a202f90c45c2b1272d3cb83bad5d96/pkg/utils/file_filter.go#L158
Tests regarding their loading seems to be performed here: https://github.com/snyk/go-application-framework/blob/main/pkg/utils/file_filter_test.go

I am not sure if we should consider this a breaking change or simply a fix. 😔
TS CLI commands (e.g.: test, container, iac) never loaded the local directory snyk.json config file. The behavior of snyk commands regarding it is currently inconsistent.

PR #662, opened 2026-07-21
fix(auth): validate auth host via domain allowlist [IDE-1896]

rrama:
### Description

Replaces the brittle allowed-host regex (`SNYK_DEFAULT_ALLOWED_HOST_REGEXP` = `^(https?://)?api(\.(.+))?\.(snyk|snykgov)\.io$`) used to validate the OAuth instance-redirect host, with parsed-URL validation.

New public API in `pkg/auth/authHost.go`:

- `IsValidSnykHost(conf configuration.Configuration, input string) (bool, error)` — orchestrates four single-purpose checks, each performed twice with independent techniques (parsed `*url.URL` **and** raw normalized string) that must agree, else fail closed:
  - `parseHostURL` — trim, prepend `https://` if schemeless, `url.Parse`, reject empty/whitespace/unparseable/non-http(s).
  - `urlIsHostOnly` / `stringIsHostOnly` — no userinfo/opaque/port/path/query/fragment.
  - `urlHasApiSubdomain` / `stringHasApiSubdomain` — leading label is `api`.
  - `urlHasAllowedDomain` / `stringHasAllowedDomain` — DNS-label-boundary match against new config key `CONFIG_KEY_ALLOWED_HOSTS` (default `{snyk.io, snykgov.io}`, configurable for private cloud), fail-closed on empty allowlist.
- New config key `auth.CONFIG_KEY_ALLOWED_HOSTS = "INTERNAL_OAUTH_ALLOWED_HOST_DOMAINS"` (`[]string`).
- `oauth2authenticator.go` `modifyTokenUrl` switched to `IsValidSnykHost`.

**Additive / non-breaking** (see [RFC: Breaking changes in GAF](https://snyksec.atlassian.net/wiki/spaces/IDE/pages/4835934272/RFC+Breaking+changes+in+GAF)): the old `CONFIG_KEY_ALLOWED_HOST_REGEXP`, `IsValidAuthHost`, and `SNYK_DEFAULT_ALLOWED_HOST_REGEXP` are retained, with plain "superseded / retained for backwards compatibility" comments — deliberately not the Go `// Deprecated:` marker, since that trips staticcheck SA1019 and could block a consumer's routine GAF bump. Real deprecation/removal is a follow-up ticket once consumers have migrated.

### Stack

Part of the IDE-1896 PR stack (base of stack):

1. **This PR** — go-application-framework
2. snyk-ls — "Depends on GAF #<this PR>" (points at this branch's pseudo-version; will be repointed to a GAF release tag before merge, per RFC merge order)
3. [cli](https://github.com/snyk/cli/pull/) — independent tidy-up, mergeable any time

### Checklist

- [x] Linted (`make lint`)
- [ ] Test your changes work for the CLI (n/a — additive, no CLI-facing change required for this PR)

Jira: [IDE-1896](https://snyksec.atlassian.net/browse/IDE-1896)

[IDE-1896]: https://snyksec.atlassian.net/browse/IDE-1896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Changes OAuth callback instance validation and token URL host rewriting—security-critical auth redirect logic—with new fail-closed rules that could block legitimate private-cloud hosts if allowlists are misconfigured.
> 
> **Overview**
> OAuth **instance redirect** validation no longer uses the regex config key **`INTERNAL_OAUTH_ALLOWED_HOSTS`** inside GAF. **`modifyTokenUrl`** now calls **`IsValidSnykHost`**, which only accepts hosts that are **`api.*`** subdomains of domains in **`CONFIG_KEY_ALLOWED_HOSTS`** (`INTERNAL_OAUTH_ALLOWED_HOST_DOMAINS`, default **`snyk.io`** / **`snykgov.io`**).
> 
> The new validator parses http(s) URLs, rejects ports/paths/userinfo/smuggling, matches allowlisted domains on **DNS label boundaries**, and **fails closed** on an empty allowlist. Each check is duplicated on parsed `*url.URL` and the normalized string so parser quirks cannot bypass rules.
> 
> **`redirectAuthHost`** now treats schemeless hosts via **`://`** instead of a **`http`** prefix, so names like **`httpfoo.example.com`** are not mis-parsed. App init registers defaults for both the legacy regex key (for external **`IsValidAuthHost`** callers) and the new domain list. Regex-based **`IsValidAuthHost`** remains exported but deprecated; broad tests cover OAuth flows and **`modifyTokenUrl`** rejection of attacker hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a23b7612120268c299ab98ad62f628fa23e784a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/auth/authHost.go
+++ b/pkg/auth/authHost.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/snyk/go-application-framework/internal/api"
+	"github.com/snyk/go-application-framework/pkg/configuration"
 	"github.com/snyk/go-application-framework/pkg/utils"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/auth/oauth2authenticator.go
```diff
@@ -473,9 +477,8 @@ func (o *oAuth2Authenticator) modifyTokenUrl(responseInstance string) error {
 		return err
 	}
 
-	redirectAuthHostRE := o.config.GetString(CONFIG_KEY_ALLOWED_HOST_REGEXP)
-	o.logger.Info().Msgf("Validating with regexp: \"%s\"", redirectAuthHostRE)
-	isValidHost, err := IsValidAuthHost(authHost, redirectAuthHostRE)
+	o.logger.Info().Msgf("Validating host: \"%s\"", authHost)
+	isValidHost, err := IsValidSnykHost(o.config, authHost)
```
basti-snyk, line 495: **Should Fix — legacy `CONFIG_KEY_ALLOWED_HOST_REGEXP` is now inert; "backwards compatibility" is only skin-deep.**

This call now routes host validation exclusively through `IsValidSnykHost`, which reads *only* `CONFIG_KEY_ALLOWED_HOSTS` (`INTERNAL_OAUTH_ALLOWED_HOST_DOMAINS`). The old key `CONFIG_KEY_ALLOWED_HOST_REGEXP` (`INTERNAL_OAUTH_ALLOWED_HOSTS`) has no remaining production reader — `IsValidAuthHost` is dead in-tree, and its registered default at `pkg/app/app.go` is never consulted.

Any downstream consumer (CLI/IDE, enterprise/self-hosted) that set `INTERNAL_OAUTH_ALLOWED_HOSTS` to permit a custom instance host (e.g. `api.snyk.mycorp.io`) will have that override **silently ignored** and OAuth to that instance will fail closed with "invalid host" — no warning, no fallback. The comments on the retained symbols say "Retained for backwards compatibility," but only the *symbol* is retained, not the behavior; per GAF's stability rules, silently changing default behavior a consumer relies on is a breaking change.

Recommend one of: (a) fall back to the regexp key when it is set to a non-default value, (b) log a warning when the old key is set-but-ignored, or (c) fix the comments to state the key is inert and flag this as a `BREAKING CHANGE:` in the changelog. Default snyk.io/snykgov.io users are unaffected — this is fail-closed, so security posture is not weakened.

— AI review

pkg/auth/authHost.go
```diff
@@ -27,10 +28,173 @@ func redirectAuthHost(instance string) (string, error) {
...
+// confirm it carries nothing but a host, confirm the host's leading DNS label
+// is "api", and confirm an allowlisted domain matches on DNS-label boundaries.
+//
+// Each step is checked twice with two independent techniques — a url* helper
+// that inspects the parsed *url.URL, and a string* helper that inspects the raw
+// normalized string — and both must agree. If the parsed URL and the raw string
+// disagree (e.g. a parser quirk that hides a smuggled component), validation
+// fails closed rather than slipping through.
+func IsValidSnykHost(conf configuration.Configuration, input string) (bool, error) {
+	allowedDomains := conf.GetStringSlice(CONFIG_KEY_ALLOWED_HOSTS)
```
basti-snyk, line 66: **Should Fix — a comma-separated env override of the allowlist matches nothing (silently allows no host).**

`conf.GetStringSlice` returns a single-element slice `[]string{v}` for a string value and does **not** split on commas. So an env override `INTERNAL_OAUTH_ALLOWED_HOST_DOMAINS="snyk.io,snykgov.io"` yields the one bogus entry `"snyk.io,snykgov.io"`, which `domainToLabels` then splits on `.` into `["snyk","io,snykgov","io"]` — never matching a real host. Only the programmatic `[]string` default and a *single-domain* env value work; the multi-domain env path silently validates nothing.

If env override of multiple domains is intended, split the entry on `,` here (e.g. `strings.Split`) before `domainToLabels`; otherwise document that the allowlist is programmatic-only / single-domain via env.

— AI review

rrama:
Re: https://github.com/snyk/go-application-framework/pull/662#issuecomment-5044484640

Won't fix — intentional defense-in-depth; if the allowlist is configured to something as broad as a single TLD label, that's an operator misconfiguration outside GAF's threat model here.

pkg/auth/authHost.go
rrama: Acknowledged, won't fix as a code change — this is intentional: the env-var override of `CONFIG_KEY_ALLOWED_HOSTS` is single-domain only by design. Multi-domain allowlists are configured programmatically (`[]string` default), not via environment variable. Documented explicitly in 17b3af6 (config key declaration + `GetStringSlice` call site in `IsValidSnykHost`).

pkg/auth/authHost.go
```diff
@@ -27,10 +28,182 @@ func redirectAuthHost(instance string) (string, error) {
...
+// normalized string — and both must agree; this is intentional
+// parser-differential defense, so a discrepancy between the two
+// interpretations (e.g. a parser quirk that hides a smuggled component) fails
+// closed rather than trusting either one alone.
+func IsValidSnykHost(conf configuration.Configuration, input string) (bool, error) {
+	// GetStringSlice does not split a plain string value on commas: an env
+	// var override of CONFIG_KEY_ALLOWED_HOSTS yields a single-element slice
+	// containing the whole value. See the CONFIG_KEY_ALLOWED_HOSTS doc
+	// comment for the single-domain-only env var caveat.
+	allowedDomains := conf.GetStringSlice(CONFIG_KEY_ALLOWED_HOSTS)
```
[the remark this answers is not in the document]
rrama: Acknowledged, won't fix as a code change — this is intentional: the env-var override of `CONFIG_KEY_ALLOWED_HOSTS` is single-domain only by design. Multi-domain allowlists are configured programmatically (`[]string` default), not via environment variable. Documented explicitly in 17b3af6 (config key declaration + `GetStringSlice` call site in `IsValidSnykHost`).

nick-y-snyk (approved):

PR #664, opened 2026-07-22
feat: add internal contributorbilling emit client for CLI billing [IANDT-237]

brettgurman-snyk:
### Description

Adds **CLI Active Contributor billing emit client** in GAF (IANDT-237). After a successful in-scope command, hosts can fire-and-forget JSON:API POSTs to entitlements-service `contributing_devs` ingest (Kafka billing downstream).

**Packaging:** core logic lives under `internal/contributorbilling/` only — no public orchestration facade in this PR. Capture middleware is **IANDT-238** (#671); command lifecycle + cliv2 wiring is **IANDT-240** (GAF #681 + CLI #7067).

#### What it does

- **`EmitContributorBilling`** — fire-and-forget async POST to the entitlements-service ingest endpoint. Returns immediately; outcomes are reported via `OnResult`. Never surfaces errors that should change the caller's exit code.
- **`BillingEmitter` + `Wait` / `WaitWithTimeout` / `WaitBudget`** — CLI billing uses one shared process-wide emitter; capture, emit, and teardown wait all use the same instance so a short-lived CLI can drain in-flight POSTs before exit. `NewEmitter()` exists for long-lived hosts (IDE/MCP) that need isolated wait scopes per workflow.
- **Git contributor collection** — when `CollectContributors` is true, scans git log (90-day author window, max 500 commits from HEAD) via `pkg/utils/git.ListContributors`. Non-git/empty repos return empty contributors without error.
- **Emit-time dedupe** — contributors deduplicated by exact email (case-sensitive, as git returns), keeping the latest commit date before each POST.
- **Ingest payload (ES contract)** — one POST per `BillingItem` with separate `contributors_entity_type` (`project`, `target`, or `revision`) and bare UUID `contributors_entity_id`. Golden fixture in `testdata/golden_ingest_payload.json`.
- **`ApplyFromConfiguration`** — fills unset `IngestURL`, auth header, and HTTP client from GAF `configuration.Configuration` + `workflow.Engine`.
- **Ingest client** — vendored oapi-codegen client under `internal/contributorbilling/client/entitlements_service` for `POST /hidden/orgs/{org_id}/contributing_devs?version=2024-10-15` (201 Created).

#### What this PR does **not** include

- HTTP capture middleware, response parsing, or `EmitFromCapture` (**IANDT-238**, #671)
- Lazy-open capture session, middleware gating, or analytics command resolver (**IANDT-238**, #671)
- `pkg/clibilling` host helpers or cliv2 wiring (**IANDT-240**, #681 + #7067)

### Architecture context (downstream stack)

The merged stack uses **lazy-open capture**: the CLI does not open a capture session at command start. Middleware opens the session on the first billable product API request; `FinishCommand` at CLI teardown closes the session and emits via `EmitFromCapture`. This PR supplies only the emit/ingest layer that downstream code calls.

### Checklist

- [x] Tests added and all succeed
- [x] Linted
- [x] GAF CI green
- [ ] E2E after #671 + #681 + #7067 merge (lazy-open stack validated locally)

### Merge order

**IANDT-237 (#664)** → **IANDT-238 (#671)** → **IANDT-240 (GAF #681 + CLI #7067)** → E2E validation

---

> [!NOTE]
> **Medium Risk**
> Introduces new billing data emission logic and network calls, but operates as best-effort fire-and-forget with validation, timeouts, and panic recovery. The workflow engine refactor is low-risk and only changes internal timeout storage.
> 
> **Overview**
> Adds a new `internal/contributorbilling` package that provides a fire-and-forget client for posting CLI Active Contributor billing events to entitlements-service. Callers invoke `EmitContributorBilling` (or `Emitter.EmitContributorBilling`) to emit one JSON:API POST per `BillingItem`; outcomes are delivered asynchronously via `OnResult` without affecting command exit codes.
> 
> The package includes contributor collection via `pkg/utils/git.ListContributors` (90-day window, 500-commit cap), deduplication by email, payload building, and a vendored oapi-codegen ingest client. It also provides `ApplyFromConfiguration` to fill network/auth defaults from GAF configuration and engine, plus `Wait`/`WaitWithTimeout`/`WaitBudget` for draining in-flight POSTs.
> 
> Separately refactors `pkg/workflow/engineimpl.go` to move the post-invoke hook timeout from a package-level variable to an instance field with a safe default, and updates `CODEOWNERS` for the new billing directory.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bceaccfd08979257bc1fd61107b2cf276189cae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/contributorbilling/wait.go
```diff
@@ -0,0 +1,36 @@
...
+// A non-positive d waits indefinitely, matching Wait().
+func WaitWithTimeout(d time.Duration) bool {
+	if d <= 0 {
+		pending.Wait()
+		return true
+	}
+
+	done := make(chan struct{})
+	go func() {
+		pending.Wait()
```
basti-snyk: **Critical — data race on the package-global `pending` WaitGroup (this is why CI is red).**

`go test -race ./pkg/contributorbilling/...` fails with `WARNING: DATA RACE`. `WaitWithTimeout` starts this goroutine to call `pending.Wait()`, but on the timeout branch it returns `false` while this goroutine stays blocked in `pending.Wait()`. Because `pending` is a package-global `sync.WaitGroup` shared by every caller, that still-blocked `Wait()` then races with a later `pending.Add(1)` in `EmitContributorBilling` (concurrent read/write of the WaitGroup counter). It is also documented `sync.WaitGroup` misuse: reusing / `Add`-ing a WaitGroup before an in-flight `Wait` has returned can panic in production (`WaitGroup is reused before previous Wait has returned`), not just trip the race detector in tests.

Root cause = global mutable concurrency state combined with a `Wait()` goroutine that outlives its own timeout. Consider making the in-flight tracking instance-scoped (an exported emitter that owns its own WaitGroup) and using a wait that does not leave a `Wait()` goroutine reading the group after the deadline. Note that eliminating the goroutine leak alone will not fully fix the reuse-after-timeout hazard — the reused-global-WaitGroup pattern is the thing to address.

— AI review

pkg/contributorbilling/types.go
```diff
@@ -0,0 +1,82 @@
...
+	SkipReasonMissingScopeID    SkipReason = "missing_scope_id"
+)
+
+// FailReason explains why an emit POST failed.
+type FailReason string
+
+const (
+	FailReasonHTTPError        FailReason = "http_error"
+	FailReasonTimeout          FailReason = "timeout"
+	FailReasonCanceled         FailReason = "canceled"
```
basti-snyk: Suggestion — `FailReasonCanceled` is never assigned anywhere in the package. `postIngest` wraps the request in `context.WithoutCancel(parent)`, so parent cancellation cannot propagate; only `DeadlineExceeded` → `FailReasonTimeout` and other errors → `FailReasonHTTPError` are ever set. Either drop this constant or wire it up, so the outcome taxonomy does not advertise a state that can never occur.

— AI review

pkg/contributorbilling/emit.go
```diff
@@ -0,0 +1,240 @@
...
+
+	result := postIngest(parent, opts, body)
+	if collectionErr != nil {
+		result.ContributorCollectionErr = collectionErr
+	}
+	return result
+}
+
+func validateRequiredFields(opts EmitOptions) SkipReason {
+	if strings.TrimSpace(opts.Capability) == "" {
```
basti-snyk: Suggestion — `validateRequiredFields` only checks that `Capability` is non-empty, not that it is one of the defined `oss | code | iac` values. A typo'd capability would silently POST bad data to the ingest endpoint. Consider validating against the `Capability*` constants and skipping with a clear reason otherwise.

— AI review

PeterSchafer:
Thank you for you contribution @brettgurman-snyk! On a first glance, there are a couple of things we might want to talk about. 

1. We generate api clients from API Specs! This will help the maintenance of the client. There is an existing package `apiclients`
2. There are git related helpers that should probably live in the existing `git` package.
3. The integration of this client still needs to be clarified, which will inform how the public library API will best be designed.

pkg/apiclients/entitlements_service/2026-07-29/spec.yaml
```diff
@@ -0,0 +1,72 @@
...
+          type: string
+        items:
+          type: array
+          items:
+            $ref: '#/components/schemas/ContributorIngestItem'
+    ContributorIngestItem:
+      type: object
+      required:
+        - scope_id
+        - target_id
```
NikkiL-Snyk: I thought we explicitly don't have target id here? so a) for normal CLI this should be project_id? and b) for evo this will be revisionId (we think). Better still, update this to entity_id and entity_type :)

brettgurman-snyk: fixed! contributors_entity_id is <entity_type>:<entity_id>

internal/contributorbilling/client/entitlements_service/client.go
```diff
@@ -0,0 +1,98 @@
...
+// IngestClient posts contributor billing ingest payloads to entitlements-service.
+type IngestClient struct {
+	api     v20260729.ClientWithResponsesInterface
+	version string
+}
+
+// NewIngestClient creates a client for the ingest endpoint at ingestURL.
+// ingestURL must include scheme and host; an optional path prefix is preserved
+// (e.g. https://api.snyk.io or https://gateway.example.com/v1).
+func NewIngestClient(httpClient *http.Client, ingestURL string) (*IngestClient, error) {
```
jonnyowenpowell, line 30: The `*http.Client` is passed in all the way from `EmitContributorBilling` in the `EmitOptions`, but I don't think that's what we want. I think we have the right idea in `internal/contributorbilling/config.go` getting the values from config, but I don't love this `EmitOptions` as a combination of options a DI container paradigm. I would prefer to see us constructing a few things at the top of that, rather than constructing them ad-hoc and using EmitOptions to ship dependencies around. Similar applies to `ingestURL`.

By the way, the client we get from the network access interface is already authenticated, so we don't need to handle auth in here. We should however set a per-request timeout and think about retries in here - I know it costs time, and we should be careful with backoffs, but no retries when we're relying on whatever network the client is on isn't great. I suggest using a wrapper around the incoming http client to implement it.

internal/contributorbilling/client/entitlements_service/client.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return &IngestClient{
+		api:     api,
+		version: DefaultIngestAPIVersion,
+	}, nil
+}
+
+// CreateContributingDevs POSTs contributors for one entity to entitlements-service ingest.
+func (c *IngestClient) CreateContributingDevs(
```
jonnyowenpowell, line 53: Minor point, but if we expect a UUID, this signature should just use `orgID uuid.UUID` and parsing should happen where we receive that value, not right before we use it.

internal/contributorbilling/client/entitlements_service/client.go
```diff
@@ -0,0 +1,98 @@
...
+	ctx context.Context,
+	orgID string,
+	authHeader string,
```
jonnyowenpowell, line 56: With my suggestion above, `authHeader` goes because it comes with the client.

internal/contributorbilling/client/entitlements_service/README.md
```diff
@@ -0,0 +1,32 @@
...
+- `client.go`: Thin wrapper used by `internal/contributorbilling`.
+
+## Usage
+
+```go
+client, err := entitlements_service.NewIngestClient(httpClient, "https://api.snyk.io")
+resp, err := client.CreateContributingDevs(ctx, orgID, authHeader, request)
+```
+
+`ingestURL` must include scheme and host; an optional path prefix is preserved when passed directly to `NewIngestClient`.
```
jonnyowenpowell, line 27: Why - we know the full path, so we can just accept a base URL.

internal/contributorbilling/client/entitlements_service/client.go
```diff
@@ -0,0 +1,98 @@
+package entitlements_service
```
jonnyowenpowell, line 1: entitlements-service is a strange place for this endpoint to live - I know we're not going to change that right now, but we could avoid baking it into this code. This client doesn't need to care, it's just using the public API, so I'd reframe it as a contributors_ingest_api client or similar.

internal/contributorbilling/client/entitlements_service/client.go
```diff
@@ -0,0 +1,98 @@
...
+		version: DefaultIngestAPIVersion,
+	}, nil
+}
+
+// CreateContributingDevs POSTs contributors for one entity to entitlements-service ingest.
+func (c *IngestClient) CreateContributingDevs(
+	ctx context.Context,
+	orgID string,
+	authHeader string,
+	body IngestRequest,
```
jonnyowenpowell, line 57: Taking an `IngestRequest` pushes a lot of request structure concerns onto the caller - I'd suggest making this take more natural parameters, like a list of contributors and and entity type/ID combo, and having the request body be created here. That moves `internal/contributorbilling/payload.go`, the associated test and the "golden_ingest_payload.json" file inside this package, which seems a lot more natural, since it's an API client concern.

The main reason I care about this it that tests in `internal/contributorbilling/emit_test.go` can become a lot simpler if the interface here is simple, focussing on the logic rather than having to deal with the API payload shape.

internal/contributorbilling/payload.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/rs/zerolog"
+	v20260729 "github.com/snyk/go-application-framework/internal/contributorbilling/client/entitlements_service/2026-07-29"
+)
+
+func buildIngestRequest(item BillingItem, logger *zerolog.Logger) v20260729.CreateContributingDevsApplicationVndAPIPlusJSONRequestBody {
+	contributors := make([]v20260729.Contributor, 0, len(item.Contributors))
+	for _, contributor := range item.Contributors {
+		if contributor.LatestCommitDate.IsZero() {
+			if logger != nil {
+				logger.Warn().Msg("contributor billing: skipping contributor with zero latest commit date")
```
jonnyowenpowell, line 15: I think validating the data _as we're building the request_ is a mistake - we should validate it as we originally collect it, and then building the request can focus on shaping the data, not doing two jobs at once.

internal/contributorbilling/emit_test.go
```diff
@@ -0,0 +1,884 @@
...
+	t.Parallel()
+
+	var (
+		mu        sync.Mutex
+		gotBody   map[string]interface{}
+		gotAuth   string
+		gotMethod string
+	)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jonnyowenpowell, line 56: These tests are made a lot harder to work with by using an HTTP server here - I'd use an interface instead of the real HTTP client and a mock (handwritten or generated, I prefer the former, but up to taste). Leave HTTP to the API client, and just check this business logic calls it with the right data - see comments on the client for a discussion of making the `CreateContributingDevs` signature clean so it's easy to test.

internal/contributorbilling/config.go
```diff
@@ -0,0 +1,30 @@
...
+	"strings"
+
+	"github.com/snyk/go-application-framework/pkg/auth"
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+)
+
+// ApplyFromConfiguration returns a copy of opts with unset network fields filled from configuration and engine.
+// Explicit HTTPClient, IngestURL, and AuthHeader values are left unchanged.
+func ApplyFromConfiguration(opts EmitOptions, config configuration.Configuration, engine workflow.Engine) EmitOptions {
```
jonnyowenpowell, line 13: As I discussed somewhat in my comments on the HTTP client, I think we should avoid stuffing everything into `EmitOptions`, and instead just construct the services we need and pass those around, behind interfaces so we can test things in isolation. We don't need a lot of layers here, so passing them around shouldn't be too painful.

internal/contributorbilling/emit_test.go
```diff
@@ -0,0 +1,884 @@
...
+	assert.Equal(t, http.MethodPost, gotMethod)
+	assert.Equal(t, "token test-token", gotAuth)
+
+	attributes := ingestAttributes(gotBody)
+	require.NotNil(t, attributes)
+	assert.Equal(t, "project-1", attributes["contributors_entity_id"])
+	assert.Equal(t, "project", attributes["contributors_entity_type"])
+}
+
+func TestEmitContributorBilling_SkipsEmptyItems(t *testing.T) {
```
jonnyowenpowell, line 108: This seems to be functionally the same as TestEmitContributorBilling_SkipsMissingEntityID

pkg/utils/git/contributors.go
```diff
@@ -0,0 +1,93 @@
+package git
```
jonnyowenpowell, line 1: I'm not convinced this belongs here - it's quite a niche capability tied to counting contributors, and I recommend it moves to internal/contributorbilling to keep it all together.

internal/contributorbilling/types.go
```diff
@@ -0,0 +1,94 @@
...
+	Item       BillingItem
+
+	// Configuration and Engine optionally supply IngestURL, AuthHeader, and HTTPClient
+	// via ApplyFromConfiguration when those fields are unset.
+	Configuration configuration.Configuration
+	Engine        workflow.Engine
+
+	// RepoPath is the default git root when CollectContributors fills empty item slices.
+	RepoPath            string
+	CollectContributors bool
```
jonnyowenpowell, line 50: Why does this exist? Would it not make more sense to presume that the call to `EmitContributorBilling` being made by a consumer should signal they want to do it, rather than providing an option to make that call but have it do nothing?

internal/contributorbilling/emitter.go
```diff
@@ -0,0 +1,177 @@
...
+	"context"
+	"fmt"
+	"sync"
+	"time"
+)
+
+// Emitter coordinates fire-and-forget contributor billing POSTs for one host scope.
+// Each Emitter owns its own in-flight tracking so Wait/WaitWithTimeout only observe
+// emits started on that instance.
+type Emitter struct {
```
jonnyowenpowell, line 13: This is all rather elaborate. I don't mind it too much because it stays out of the way of the business logic, but I would be keen to discuss why we need it - we've got a lot of similar work going on in the post hook mechanism introduced by the base branch here, and if we really do have to add a bunch _more_, it feels like that mechanism is missing some functionality?

internal/contributorbilling/emitter.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// EmitContributorBilling fires an async POST to entitlements-service ingest.
+// It returns immediately and never surfaces an error that should fail the caller's command.
+// Short-lived hosts (e.g. the CLI) must call Wait or WaitWithTimeout on the same Emitter
+// before process exit.
+func (e *Emitter) EmitContributorBilling(ctx context.Context, opts EmitOptions) {
+	opts = opts.withDefaults()
+	opts = ApplyFromConfiguration(opts, opts.Configuration, opts.Engine)
+	opts.Item = cloneItem(opts.Item)
```
jonnyowenpowell, line 29: Why? We should probably have this be a genuine input, rather than a partial object we consume and then enrich as we go along, which would remove the "need" to clone it, although I suspect in practice the caller won't be re-using it later anyway.

internal/contributorbilling/emit.go
```diff
@@ -0,0 +1,184 @@
...
+		cloned.RepoPath = resolveRepoPath(item.RepoPath)
+	}
+	if len(item.Contributors) > 0 {
+		cloned.Contributors = make([]Contributor, len(item.Contributors))
+		copy(cloned.Contributors, item.Contributors)
+	}
+	return cloned
+}
+
+func (opts EmitOptions) withDefaults() EmitOptions {
```
jonnyowenpowell, line 38: There's a lot of uncertainty here - I think we should decide on what we should know, and what our caller should know, and stick to that, rather than having layers of processing options. Additionally, half of this function is santization, which reads as strange in a `withDefaults` function.

internal/contributorbilling/contributors.go
```diff
@@ -0,0 +1,85 @@
+package contributorbilling
```
jonnyowenpowell, line 1: This feel feels like functionality that is missing from the other contributor collection code which is therefore wrapper around it later - I'd prefer to see this folded into the contributor collection code, so we see the full mechanism in one place, and can have sane tests that check the thing works as expected, rather than splitting the work somewhat arbitrarily.

internal/contributorbilling/types.go
```diff
@@ -0,0 +1,94 @@
...
+	RepoPath string
+}
+
+// EmitOptions configures a fire-and-forget contributor billing POST.
+type EmitOptions struct {
+	HTTPClient *http.Client
+	IngestURL  string
+	AuthHeader string
+	// Capability is an optional caller-side telemetry label (e.g. oss, code, iac).
+	// It is not sent on the HTTP payload.
```
jonnyowenpowell, line 38: This seems to be a mischaracterisation - it's actually just used to decide whether to reject the request to emit contributors.

jonnyowenpowell (commented):
I have given this a first pass - I think we can _significantly_ reduce the lines of code and make this a lot easier to read, at which point it should be easier to see the wood from the trees and check it's correct.

jonnyowenpowell:
Replaced by https://github.com/snyk/go-application-framework/pull/721

PR #666, opened 2026-07-24
feat(fileupload): add path encoder and content transcoder options

florinMiro:
### Description

Add `WithPathEncoder` and `WithContentTranscoder` to the high-level file upload client
(`pkg/apiclients/fileupload`), so callers can URI-encode upload paths and transcode content
(e.g. to UTF-8) without touching the internal upload revision client. Both default to no-op.

### Checklist

- [x] Linted (`make lint`)

---

> [!NOTE]
> **Medium Risk**
> Changes the internal upload revision `UploadFile` contract and shifts when/how file content is read and validated during batching and upload, which could affect callers or memory use when transcoding is enabled.
> 
> **Overview**
> Adds **`WithPathEncoder`** and **`WithContentTranscoder`** on the high-level `fileupload` client so upload paths and file bytes can be transformed (e.g. URI-encoding paths, transcoding to UTF-8) before limits and upload run. Encoded paths drive path-length filtering and skipped-file reporting; transcoded byte length drives size limits and what is sent.
> 
> Refactors upload batching so files are not opened for the whole batch: **`UploadFile`** now carries an explicit **`Size`** and **`io.Reader`** instead of **`fs.File`**, and the low-level upload revision client validates and streams from that reader. Batching builds lazy readers (`transcodedFileReader`) and moves regular-file / readability checks into **`createUploadFile`** in the high-level layer.
> 
> Filters now use transcoded content size rather than **`os.FileInfo`**. Tests are updated for the new **`UploadFile`** shape and expanded with option-focused coverage (encoded paths, transcoder failures, symlinks, unreadable files).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b57c739f8675d07d97d9c1bc42acebe4fafb4c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/api/fileupload/uploadrevision/client_test.go
+++ b/internal/api/fileupload/uploadrevision/client_test.go
@@ -1,6 +1,7 @@
 package uploadrevision_test
 
 import (
+	"bytes"
 	"compress/gzip"
 	"context"
 	"errors"
@@ -24,6 +25,15 @@ var (
 	orgID = uuid.MustParse("9102b78b-c28d-4392-a39f-08dd26fd9622")
 	revID = uuid.MustParse("ff1bd2c6-7a5f-48fb-9a5b-52d711c8b47f")
 )
+
+// uploadFile builds an UploadFile whose Size matches the content its Reader produces.
+func uploadFile(path string, content []byte) uploadrevision2.UploadFile {
+	return uploadrevision2.UploadFile{
+		Path:   path,
+		Size:   int64(len(content)),
+		Reader: bytes.NewReader(content),
+	}
+}
 
 func TestClient_CreateRevision(t *testing.T) {
 	srv, c := setupTestServer(t)
@@ -71,7 +81,7 @@ func TestClient_UploadFiles(t *testing.T) {
 		orgID,
 		revID,
 		[]uploadrevision2.UploadFile{
-			{Path: "foo/bar", Content: []byte("asdf")},
+			uploadFile("foo/bar", []byte("asdf")),
 		})
 
 	require.NoError(t, err)
@@ -85,8 +95,8 @@ func TestClient_UploadFiles_MultipleFiles(t *testing.T) {
 		orgID,
 		revID,
 		[]uploadrevision2.UploadFile{
-			{Path: "file1.txt", Content: []byte("content1")},
-			{Path: "file2.json", Content: []byte("content2")},
+			uploadFile("file1.txt", []byte("content1")),
+			uploadFile("file2.json", []byte("content2")),
 		})
 
 	require.NoError(t, err)
@@ -99,7 +109,7 @@ func TestClient_UploadFiles_EmptyOrgID(t *testing.T) {
 		uuid.Nil, // empty orgID
 		revID,
 		[]uploadrevision2.UploadFile{
-			{Path: "test.txt", Content: []byte("content")},
+			uploadFile("test.txt", []byte("content")),
 		})
 
 	assert.Error(t, err)
@@ -113,7 +123,7 @@ func TestClient_UploadFiles_EmptyRevisionID(t *testing.T) {
 		orgID,
 		uuid.Nil, // empty revisionID
 		[]uploadrevision2.UploadFile{
-			{Path: "test.txt", Content: []byte("content")},
+			uploadFile("test.txt", []byte("content")),
 		})
 
 	assert.Error(t, err)
@@ -129,7 +139,7 @@ func TestClient_UploadFiles_FileSizeLimit(t *testing.T) {
 		orgID,
 		revID,
 		[]uploadrevision2.UploadFile{
-			{Path: "large_file.txt", Content: largeContent},
+			uploadFile("large_file.txt", largeContent),
 		})
 
 	assert.Error(t, err)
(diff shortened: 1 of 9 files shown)


pkg/apiclients/fileupload/batch.go
```diff
@@ -73,14 +91,25 @@ func batchPaths(
 				continue
 			}
 
-			f, err := os.Open(path)
+			relPath = pathEncoder(relPath)
+
+			osFile, err := os.Open(path)
 			if err != nil {
 				logger.Debug().Msgf("failed to open file: %s", path)
-				f.Close()
```
PeterSchafer, line 79: Question: Why was this removed here? It is a bind mind bending to understand, maybe we can make this easier?

florinMiro: This was removed because it was redundant.

If we fail to open `f` there is no file to close.

I'll spend a bit of time looking for simplifications on file content transcoding and file closing.

florinMiro: Both you and Cursor were right.
The use of the File interface was overly complicated, and it created a file descriptor leak.

I've refactored the ContentEncoder interface(+Batcher as a side effect) to expect and return byte buffers, and completely removed the stacked file-closing logic.

florinMiro: I'm changing internal interfaces + making the Batcher load the file content into memory.
But I think this is acceptable.

internal/api/fileupload/uploadrevision/types.go
```diff
@@ -110,10 +108,10 @@ type ErrorResponseBody struct {
 	Errors []ResponseError `json:"errors"`
 }
 
-// UploadFile represents a file to be uploaded, containing both the path and file handle.
+// UploadFile represents a file to be uploaded, containing both the path and the content.
 type UploadFile struct {
-	Path string // The path of the uploaded file, relative to the root directory.
-	File fs.File
+	Path    string // The path of the uploaded file, relative to the root directory.
+	Content []byte // The content to upload, already read and transcoded.
```
PeterSchafer: Question: Did we look into the impact on memory consumption? should we maybe prefer a lazy read and ensure that the Content is cleared soon after consumption?

PeterSchafer: I looked a bit more at the code and I'm now more convinced we should read and transform the file content on demand only. Basically introducing a Getter for Content at `UploadFile`. It can use `os.ReadFile()` and apply `contentTranscoder()` as it is currently done in `batchPaths()`.

internal/api/fileupload/uploadrevision/client_test.go
```diff
@@ -92,22 +91,12 @@ func TestClient_UploadFiles_MultipleFiles(t *testing.T) {
...
-		"file2.json": {Data: []byte("content2")},
-	}
-
-	file1, err := mockFS.Open("file1.txt")
-	require.NoError(t, err)
-	file2, err := mockFS.Open("file2.json")
-	require.NoError(t, err)
-
-	err = c.UploadFiles(context.Background(),
+	err := c.UploadFiles(context.Background(),
```
PeterSchafer: nitpick: while here, maybe use t.Context()

internal/api/fileupload/uploadrevision/client_test.go
```diff
@@ -30,6 +26,15 @@ var (
 	revID = uuid.MustParse("ff1bd2c6-7a5f-48fb-9a5b-52d711c8b47f")
 )
 
+// uploadFile builds an UploadFile whose Size matches the content its Reader produces.
+func uploadFile(path string, content []byte) uploadrevision2.UploadFile {
```
PeterSchafer, line 29: Praise: Very nice!

pkg/apiclients/fileupload/batch.go
```diff
@@ -38,22 +39,115 @@ func (b *uploadBatch) isEmpty() bool {
...
+		r.content = bytes.NewReader(content)
+	}
+
+	return r.content.Read(p)
+}
+
+// createUploadFile turns a filesystem path into the file to upload, or reports why it cannot be
+// uploaded. The returned file holds no content and no descriptor; both are produced when it is
+// streamed.
+func createUploadFile(
```
PeterSchafer, line 98: Praise: Nice refactoring!

internal/api/fileupload/uploadrevision/client_test.go
```diff
@@ -332,119 +259,14 @@ func TestClient_UploadFiles_IndividualFileSizeExactlyAtLimit(t *testing.T) {
...
-			assert.Error(t, err)
-
-			var sfe *uploadrevision2.SpecialFileError
-			assert.ErrorAs(t, err, &sfe)
-			assert.Equal(t, filePath, sfe.FilePath)
-		})
-	}
-}
-
-func TestClient_UploadFiles_Symlink(t *testing.T) {
```
PeterSchafer, line 420: Question: It seems that this test case still has value even in the new implementation. Is there a reason we removed it?

florinMiro: Yes, I removed it because this method is no longer concerned with file types.

I've moved this test case to the `CreateRevisionFromChan` suite as a consequence:
https://github.com/snyk/go-application-framework/pull/666/changes/BASE..a6be78563a21bb7f4e011735958c4942b956ae52#diff-fd5d6d7e26d97c1fb5aa38325408e37cd029cc9b5b7904040a68a9ac25ab7f8eR185

internal/api/fileupload/uploadrevision/client_test.go
florinMiro: Fixed https://github.com/snyk/go-application-framework/pull/666/changes/2b57c739f8675d07d97d9c1bc42acebe4fafb4c9

PeterSchafer (approved):
https://github.com/snyk/cli/pull/7059

PR #667, opened 2026-07-24
fix: Gate ignore-rule metacharacter fix behind an opt-in flag

danskmt:
### Description

Adds `NewFileFilterFromConfig`, which resolves feature-flag-gated `FileFilter` behavior (currently the ignore-rule metacharacter fix) directly from a `configuration.Configuration`, so future flags of this kind don't need caller changes. `config` may be `nil`, which keeps such behavior at its default.

`NewFileFilter` is now deprecated in favor of it. Flag keys shared across packages live in a new `pkg/featureflags` package (avoids an import cycle with `pkg/configuration`).

### Checklist

- [x] Linted (`make lint`)

PR in CLI: https://github.com/snyk/cli/pull/7039

---

> [!NOTE]
> **Medium Risk**
> Changes which files are excluded when paths or rules contain regex metacharacters; behavior depends on the flag and which constructor callers use, so incorrect rollout could alter scan scope.
> 
> **Overview**
> Introduces **`NewFileFilterFromConfig`**, which reads feature-flag-gated behavior (starting with the ignore-rule metacharacter fix) from configuration via a small **`pkg/featureflags`** package, avoiding an import cycle with `pkg/configuration`. **`NewFileFilter`** is deprecated but still forces the metacharacter fix on for unmigrated callers.
> 
> When the flag is off (the default for `newFileFilterInternal` / nil config), ignore parsing uses a **legacy** glob path that only escapes `$`; when on, rules and base paths escape the full set of regex metacharacters gitignore treats as literal before matching with go-gitignore.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 820cf89e54a8c202e53b2a669d54d47e319763a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/configuration/constants.go
+++ b/pkg/configuration/constants.go
@@ -1,5 +1,7 @@
 package configuration
 
+import "github.com/snyk/go-application-framework/pkg/featureflags"
+
 const (
 	// ---------
 	// platform related configuration that is normally not required to be accessed directly
@@ -95,12 +97,12 @@ const (
 	// Feature flag to enable native implementation for code, used in code-client-go's code workflow
 	FF_CODE_NATIVE_IMPLEMENTATION string = "internal_snyk_code_native_implementation"
 	// Feature flag to enable the fix for ignore rules/paths containing regex metacharacters
-	FF_FILE_FILTER_METACHARACTER_FIX string = "internal_snyk_file_filter_metacharacter_fix_enabled"
+	FF_FILE_FILTER_METACHARACTER_FIX string = featureflags.FileFilterMetacharacterFix
 
 	// ---------
 	// feature flag names (platform-registered names backing the config keys above)
 	// ---------
 
 	// Feature flag name backing FF_FILE_FILTER_METACHARACTER_FIX
-	SNYK_FILE_FILTER_METACHARACTER_FIX string = "snykFileFilterMetacharacterFix"
+	SNYK_FILE_FILTER_METACHARACTER_FIX string = featureflags.FileFilterMetacharacterFixBackendName
 )
--- a/pkg/configuration/constants.go
+++ b/pkg/configuration/constants.go
@@ -94,4 +94,13 @@ const (
 	FF_CODE_CONSISTENT_IGNORES string = "internal_snyk_code_ignores_enabled"
 	// Feature flag to enable native implementation for code, used in code-client-go's code workflow
 	FF_CODE_NATIVE_IMPLEMENTATION string = "internal_snyk_code_native_implementation"
+	// Feature flag to enable the fix for ignore rules/paths containing regex metacharacters
+	FF_FILE_FILTER_METACHARACTER_FIX string = "internal_snyk_file_filter_metacharacter_fix_enabled"
+
+	// ---------
+	// feature flag names (platform-registered names backing the config keys above)
+	// ---------
+
+	// Feature flag name backing FF_FILE_FILTER_METACHARACTER_FIX
+	SNYK_FILE_FILTER_METACHARACTER_FIX string = "snykFileFilterMetacharacterFix"
 )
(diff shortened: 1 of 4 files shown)


octavian-snyk (approved):

octavian-snyk (approved):

octavian-snyk (approved):

pkg/configuration/constants.go
```diff
@@ -94,4 +96,13 @@ const (
 	FF_CODE_CONSISTENT_IGNORES string = "internal_snyk_code_ignores_enabled"
 	// Feature flag to enable native implementation for code, used in code-client-go's code workflow
 	FF_CODE_NATIVE_IMPLEMENTATION string = "internal_snyk_code_native_implementation"
+	// Feature flag to enable the fix for ignore rules/paths containing regex metacharacters
+	FF_FILE_FILTER_METACHARACTER_FIX string = featureflags.FileFilterMetacharacterFix
```
PeterSchafer, line 100: @danskmt please don't add more feature flags here, instead keep them close to the logic where they are used. In this case file filter.

pkg/featureflags/featureflags.go
```diff
@@ -0,0 +1,18 @@
+// Package featureflags is a minimal, dependency-free home for feature-flag surfaces shared
+// between packages that can't import pkg/configuration directly (pkg/configuration already
+// depends on pkg/utils transitively, via internal/utils, so pkg/utils importing
+// pkg/configuration back would be an import cycle).
+package featureflags
+
+// FeatureFlagReader is satisfied by any configuration.Configuration.
+type FeatureFlagReader interface {
```
PeterSchafer, line 8: Question: Why are you limiting the interface here? Why not just use configuration.Configuration? The existing Configuration interface is much richer and supports for example retrieving values with Errors. Limiting to a narrow interface does create additional cognitive and maintenance complexity.

pkg/utils/file_filter.go
```diff
@@ -102,6 +118,19 @@ func NewFileFilter(path string, logger *zerolog.Logger, options ...FileFilterOpt
 	return filter
 }
 
+// NewFileFilterFromConfig resolves feature-flag-gated behavior (like the ignore-rule
+// metacharacter fix) from config, so future flags of this kind need no caller changes. config
+// may be nil, in which case such behavior stays at its default.
+func NewFileFilterFromConfig(path string, logger *zerolog.Logger, config featureflags.FeatureFlagReader, options ...FileFilterOption) *FileFilter {
```
PeterSchafer, line 124: Issue: Introducing a new NewXyZ function is the opposite of what we have been doing in the GAF interface design, instead we have been shifting to the Functional Options Pattern. Why do we do both here?

pkg/utils/file_filter.go
```diff
@@ -408,6 +438,24 @@ func escapeSpecialGlobChars(rule string) string {
 	return result.String()
 }
 
+// escapeSpecialGlobCharsLegacy escapes special characters that should be treated literally in
+// glob patterns. Special Characters to escape: $
+// This is the legacy behavior, to be removed in future releases.
+func escapeSpecialGlobCharsLegacy(rule string) string {
```
PeterSchafer, line 444: Question: Why is this function duplicated rather then escapeIgnoreRuleMetaChars() parameterized with ruleRegexMetaChars?

pkg/utils/file_filter_test.go
```diff
@@ -36,7 +36,7 @@ func TestFileFilter_GetAllFiles(t *testing.T) {
 		tempFile2 := filepath.Join(tempDir, "test2.ts")
 		createFileInPath(t, tempFile2, []byte{})
 
-		fileFilter := NewFileFilter(tempDir, &log.Logger)
+		fileFilter := newFileFilterInternal(tempDir, &log.Logger)
```
PeterSchafer, line 39: Suggestion: Test should cover the public interface as well not just an internal one

pkg/featureflags/featureflags.go
PeterSchafer: Context: The design goal of the configuration package is to decouple the source and the consumption of configuration values. So that a consumer doesn't know if a value is a feature flag or a command line argument or an environment variable. It is just all configuration accessed via a name.

PR #671, opened 2026-07-28
feat: add internal contributorbilling capture middleware for CLI billing [IANDT-238]

brettgurman-snyk:
## Summary

Adds GAF **HTTP capture middleware** and **`EmitFromCapture`** for CLI Active Contributor billing. Middleware records billing entities from successful product API responses into a **command-scoped capture session**. At CLI teardown (IANDT-240), captured records become IANDT-237 ingest POSTs.

**Depends on IANDT-237** (`internal/contributorbilling` emit/ingest + shared `BillingEmitter`). **Does not go live end-to-end until IANDT-240 lands** (#681 + CLI #7067): `FinishCommand` at tearDown closes the session and emits.

## Lazy-open capture (current design)

The capture session is **not** opened at CLI startup. It opens **lazily on the first billable product API request**:

```text
snyk monitor
  → analytics.SetCommand("monitor")          # normal instrumentation
  → first billable HTTP (e.g. PUT /v1/monitor/...)
       → middleware: URL match → capture flag → billable command?
       → EnsureCaptureSessionForConfig()     # session opens here
       → parse response → bag.Add(project UUID)
  → tearDown → FinishCommand (IANDT-240)     # close + EmitFromCapture
```

**Not in this PR:** `BeginCommand`, `WithContributorBillingCapture()`, or any cliv2 billing call at command start.

### Capture (`internal/contributorbilling/capture`)

- Process-scoped **command session** (`EnsureCommandSession` / `CloseCommandSession` / `ActiveCapture`)
- **`gate.go`** — `IsBillableCommand`, `CaptureEnabledForBillableHTTP`, `EnsureCaptureSessionForConfig`, `CommandNameFromRawArgs`
- Parsers per supported CLI command / API flow:

  **`snyk monitor` (OSS)**
  - Registry `PUT /v1/monitor/...` or `PUT /v1/monitor-dependencies/...`
  - 2xx response: parse `uri` for project public ID
  - Captured entity: project UUID, capability `oss`

  **`snyk iac test --report` (IaC legacy TS path)**
  - Registry `POST /v1/iac-cli-share-results`
  - 2xx response: parse share-result map for project IDs
  - Captured entity: project UUID, capability `iac`

  **`snyk code test --report` (Code native `--report`)**
  - Test API: `POST /orgs/{org}/tests` with `publish_report: true` and SAST scan config
  - Follow-up on 2xx: `GET /orgs/{org}/tests/{test}` or `GET .../components` → project ID
  - Captured entity: project UUID, capability `code`

  **Native OS `--report` (Test API, non-legacy monitor path)**
  - Same Test API create/follow-up flow with `publish_report: true` without SAST config
  - Captured entity: project UUID, capability `oss`

- Gated on `contributor_billing_capture_enabled` (FFS `enable-entity-contributors-publish`) and billable commands

### Middleware (`ContributorCaptureMiddleware`)

- Wired into the default GAF round tripper for **all HTTP paths**
- **URL-first gating** — checks matched product API traffic before reading the capture flag, avoiding config/org/FFS recursion during auth lookups
- **Lazy session open** — `EnsureCaptureSessionForConfig` on first in-scope billable request when flag + command are eligible
- **`networking.SetActiveCommandResolver`** + **`analytics.GetCommand()`** — active CLI command from analytics, with `RAW_CMD_ARGS` fallback
- Guards: known Snyk host, 2xx only, 64 KiB body cap, safe response-body replay, panic recovery

### Emit bridge

- **`EmitFromCapture`** — groups deduped capture records by capability and calls `EmitContributorBilling` with git contributor collection at emit time

## What this PR does **not** include

- `pkg/clibilling` host helpers (`FinishCommand`, post-invoke failure cleanup) — **IANDT-240 (#681)**
- cliv2 wiring — **CLI #7067**
- Ingest POST at command teardown — **IANDT-240** (`FinishCommand` owns close + emit)

Safe to merge without billing going live: middleware only captures when the flag is on and a billable command is active; **emit does not run until IANDT-240 `FinishCommand` at tearDown**.

## Test plan

- [x] Unit + middleware tests (`internal/contributorbilling/capture`, `contributor_capture`, `from_capture`)
- [x] GAF CI green (including `-race` on capture/middleware)
- [x] Lazy-open E2E validated locally with #681 + #7067
- [ ] Production E2E after full stack merges

## Merge order

**IANDT-237 (#664)** → **IANDT-238 (#671)** → **IANDT-240 (GAF #681 + CLI #7067)** → E2E validation

---

> [!NOTE]
> **Medium Risk**
> Adds billing HTTP emits and global network middleware that parses product API traffic; failures are fire-and-forget but mis-capture or flag rollout could post wrong or unexpected contributor data. Scope is feature-flagged and ingest at tearDown remains out of this PR.
> 
> **Overview**
> Introduces **CLI Active Contributor billing** in GAF: HTTP middleware records project/target entities from successful in-scope API calls into a **lazy-opened command capture session**, and **`EmitFromCapture`** / **`EmitContributorBilling`** fire-and-forget POSTs to entitlements-service ingest (with optional **90-day git log** contributor collection). Gated by **`contributor_billing_capture_enabled`** (FFS `enable-entity-contributors-publish`) and billable commands (`monitor`, `iac test --report`, `code test --report`).
> 
> **Capture** parses Registry monitor/IaC share responses and Test API create/follow-up flows (OSS/Code/IaC capabilities). **ContributorCaptureMiddleware** sits on the default round tripper with URL-first gating to avoid FFS/auth recursion, 64 KiB body limits, and safe body replay for downstream consumers.
> 
> Also adds **`workflow.PostInvokeHook`** / **`WithPostInvokeHooks`**, **`analytics.GetCommand()`**, and **`SetActiveCommandResolver`** for command-aware capture. End-to-end billing at CLI teardown is **not wired here** (depends on IANDT-240 `FinishCommand`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 148d2e37fff7a1a1818ea610ce820013473f0693. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/networking/contributorcapture/capture.go
```diff
@@ -0,0 +1,86 @@
...
+	"sync"
+)
+
+type contextKey struct{}
+
+// Capability identifies the CLI product flow that produced a captured project ID.
+type Capability string
+
+const (
+	CapabilityOSS  Capability = "oss"
```
rahultoora: [question]: shouldnt this be:

```go
CapabilityOS  Capability = "os"
```

i.e os not oss

pkg/networking/middleware/contributor_capture.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil || res == nil || req == nil || req.URL == nil {
+		return res, err
+	}
+
+	m.tryCapture(req, res)
+	return res, err
+}
+
+func (m *ContributorCaptureMiddleware) tryCapture(req *http.Request, res *http.Response) {
+	capture := contributorcapture.FromContext(req.Context())
```
rahultoora: [thought/suggestion]: tryCapture (which is invoked in RoundTrip) runs JSON/regex/UUID parsing inline. If any of it ever panics (which I know is not likely with today's code because json.Unmarshal/regexp/uuid.Parse return errors rather than panicking . However a future parser could), the panic propagates up and fails the user's actual CLI request. 

For a passive "spy" whose entire contract is to be invisible, that's the wrong failure mode. 

Should we wrap the capture work so it can never take down the request? 

```go
func (m *ContributorCaptureMiddleware) tryCapture(req *http.Request, res *http.Response) {
    defer func() {
        if r := recover(); r != nil && m.logger != nil {
            m.logger.Debug().Interface("panic", r).Msg("contributor capture: recovered from panic")
        }
    }()
    // ... existing body ...
}

```

pkg/networking/middleware/contributor_capture.go
```diff
@@ -0,0 +1,100 @@
...
+		})
+	}
+}
+
+func readResponseBody(res *http.Response) ([]byte, error) {
+	if res.Body == nil {
+		return nil, nil
+	}
+
+	bodyBytes, err := io.ReadAll(res.Body)
```
rahultoora: [question]: is it worth passing in `io.LimitReader` to set a ceiling on what we read into memory?

pkg/networking/contributorcapture/parse.go
```diff
@@ -0,0 +1,131 @@
...
+func parseProjectID(value string) string {
+	projectID, err := uuid.Parse(strings.TrimSpace(value))
+	if err != nil {
+		return ""
+	}
+
+	return projectID.String()
+}
+
+func parseIaCShareResponse(body []byte) []string {
```
rahultoora: [suggestion]: if my understanding is correct what this is doing is "any value in this object that happens to be a UUID string is a project ID". That's a heuristic tied to today's response shape. 

Is it worth adding a comment here mentioning this or even looking for the specific key instead?

pkg/networking/middleware/contributor_capture_test.go
```diff
@@ -0,0 +1,183 @@
...
+	logger := zerolog.Nop()
+	rt := middleware.NewContributorCaptureMiddleware(http.DefaultTransport, config, &logger)
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodPut, server.URL+"/v1/monitor/npm", http.NoBody)
+	require.NoError(t, err)
+	require.Nil(t, contributorcapture.FromContext(req.Context()))
+	require.Same(t, capture, contributorcapture.FromContext(ctxWithBag))
+
+	res, err := rt.RoundTrip(req)
+	require.NoError(t, err)
```
rahultoora, line 279: [suggestion]: good tests! Can we add a negative test case for a matched matched path (`/v1/monitor/...`) on an unknown host and validate that nothing is captured.

pkg/networking/middleware/contributor_capture.go
```diff
@@ -0,0 +1,100 @@
...
+	apiURL := m.config.GetString(configuration.API_URL)
+	additionalSubdomains := m.config.GetStringSlice(configuration.AUTHENTICATION_SUBDOMAINS)
+	additionalURLs := m.config.GetStringSlice(configuration.AUTHENTICATION_ADDITIONAL_URLS)
+
+	isKnownHost, err := ShouldRequireAuthentication(apiURL, req.URL, additionalSubdomains, additionalURLs)
+	if !isKnownHost || err != nil {
+		return
+	}
+
+	capability, matched := contributorcapture.MatchRequest(req.Method, req.URL.Path)
```
rahultoora: [nit]: my understanding of what the code does now in order:


1. capture bag present?
2. status 2xx?
3. m.config.GetString(...) + two m.config.GetStringSlice(...) ← config reads
4. ShouldRequireAuthentication(...) ← parses the URL
5. MatchRequest(method, path) ← two cheap regex checks
6. read body / parse

MatchRequest in step 5 is a couple of quick regex rejects. If you move MatchRequest before the host/config work, the common "not a monitor/iac request" case bails out immediately and you skip the config reads and URL parsing entirely

pkg/networking/contributorcapture/parse.go
```diff
@@ -0,0 +1,131 @@
...
+
+var (
+	monitorPathPattern = regexp.MustCompile(`(?i)^/v1/monitor(?:/|$)`)
+	monitorDepsPattern = regexp.MustCompile(`(?i)^/v1/monitor-dependencies(?:/|$)`)
+	iacSharePattern    = regexp.MustCompile(`(?i)^/v1/iac-cli-share-results(?:/|$)`)
+	projectIDFromURI   = regexp.MustCompile(`/project/([0-9a-fA-F-]{36})(?:/|$)`)
+)
+
+type monitorResult struct {
+	ID  string `json:"id"`
```
rahultoora: [question]: is the ID used can it be dropped?

pkg/networking/middleware/contributor_capture.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusMultipleChoices {
+		return
+	}
+
+	apiURL := m.config.GetString(configuration.API_URL)
+	additionalSubdomains := m.config.GetStringSlice(configuration.AUTHENTICATION_SUBDOMAINS)
+	additionalURLs := m.config.GetStringSlice(configuration.AUTHENTICATION_ADDITIONAL_URLS)
+
+	isKnownHost, err := ShouldRequireAuthentication(apiURL, req.URL, additionalSubdomains, additionalURLs)
```
rahultoora, line 238: [nit]: A one-line comment on why ShouldRequireAuthentication is used as a "known Snyk host" check would be helpful as it's not obvious that "requires auth" is being repurposed as "safe host to capture from."

rahultoora (approved):

rahultoora:
overall good work have some non blocking comments. Have you got a jira ticket created for the code path?

pkg/networking/contributorcapture/emit.go
```diff
@@ -0,0 +1,28 @@
+package contributorcapture
+
+import (
+	"context"
+)
+
+// EmitOptions configures a future contributor billing POST from captured records.
+// TODO(IANDT-165): wire to contributorbilling.EmitContributorBilling once
```
rahultoora: this is the incorrect ticket you have linked https://snyksec.atlassian.net/browse/IANDT-165 as it  is about dbt

pkg/networking/contributorcapture/emit.go
```diff
@@ -0,0 +1,28 @@
...
+	// RepoPath is the git root used when collecting contributors at emit time.
+	RepoPath string
+}
+
+// EmitCapturedRecords will post contributor billing for project IDs captured during the
+// current CLI interaction. It is intentionally a no-op until pkg/contributorbilling
+// merges to main; callers may invoke it from cliv2 teardown ahead of that wiring.
+// When capture is nil, it returns immediately without side effects.
+//
+// TODO(IANDT-165): implement using contributorbilling.EmitContributorBilling and
```
rahultoora: again this ticket is related to dbt

pkg/networking/contributorcapture/capture.go
NikkiL-Snyk, line 1: Do we want to be codeowners of this?

pkg/networking/contributorcapture/capture.go
```diff
@@ -0,0 +1,86 @@
...
+	mu      sync.Mutex
+	records []Record
+}
+
+// NewCapture returns an empty capture bag.
+func NewCapture() *Capture {
+	return &Capture{}
+}
+
+// Add appends a record when projectID is non-empty. Duplicate project IDs are allowed
```
NikkiL-Snyk: I'm not following this comment? Why would we get duplicate projects?

pkg/networking/contributorcapture/capture.go
brettgurman-snyk: oss matches pre-existing gaf work from 2024, as well as the pitch doc

not sure if it hugely matters as long as we're consistent

pkg/networking/contributorcapture/parse.go
brettgurman-snyk: there isnt a stable key name to match, so going with adding a comment

pkg/networking/contributorcapture/capture.go
brettgurman-snyk: comment was wrong, multiple IaC targets in one share produce different project IDs. duplicates would occur if the same project is captured from multiple monitor/share calls in one CLI interaction.

that being said i dont think that detail even belongs here, so i'm removing it from the comment

pkg/networking/contributorcapture/capture.go
rahultoora: the capability is called open source so would prefer it be called that personally what does the additional s stand for?

jonnyowenpowell:
This is _very_ invisible to the product teams who own these flows/APIs. How do we plan to avoid silent breakages if their APIs change, they migrate onto new APIs, etc. ?

internal/contributorbilling/client/entitlements_service/2026-07-29/spec.yaml
```diff
@@ -0,0 +1,106 @@
...
+                - contributing_devs
+            attributes:
+              $ref: '#/components/schemas/ContributingDevsIngestAttributes'
+    ContributingDevsIngestAttributes:
+      type: object
+      required:
+        - contributors_entity_id
+        - contributors
+      properties:
+        contributors_entity_id:
```
NikkiL-Snyk, line 82: Raised this with @rahultoora - why do we have just an entity id and not an id and type (I think the id is a combination of id and type, but why not keep them separate as they will be in the ongoing events etc? Having it in one field just adds work right?

internal/contributorbilling/client/entitlements_service/2026-07-29/spec.yaml
```diff
@@ -0,0 +1,106 @@
...
+info:
+  title: Contributing Devs Ingest API
+  description: |
+    entitlements-service hidden ingest API for CLI Active Contributor billing.
+    Vendored from contributingDevsIngestApi (2024-10-15); sweater-comb refs inlined for codegen.
+  version: 2024-10-15
+servers:
+  - url: /
+paths:
+  /hidden/orgs/{org_id}/contributing_devs:
```
NikkiL-Snyk, line 11: Does this spec have all the data we need? Do we not want to record which cli command/capability sent the event/AC list? currently I don't think there's a way that the ingest API can distinguish between a code CLI command or an evo one?

NikkiL-Snyk: I can see a world in which it doesn't matter where we get the AC list from, but want to verify if it's something we should be sending

rahultoora: its a good question we previously discussed sending "categories" data to the api similar to how its sent to analytics-service so we can see what cli command was ran based on the categories: code, iac, aibom, secrets, container, monitor, report, upload etc so we could filter on the ES side and not publish AC data to the topic if the category data reveals its a non billable cli command that was ran. 

We decided it makes more sense to do that filter in the cli ro reduce traffic to the es api. 

If we can see an immediate need to send more data to the ingest api im happy to update the spec but we can always update it later on i dont want to send additional data to the ingest api for the sake of it though

NikkiL-Snyk: Yeah ok, given that the events we're producing don't have that info on then it is probably fine, as the info would be lost at that point anyway! I haven't got as far in my review to see if we make metrics with relevant info on this, but we should make sure to include that on a metric at least  for traceabilty:)

internal/contributorbilling/client/entitlements_service/2026-07-29/entitlements_service.go
basti-snyk, line 1: Wondering: why not use / add to https://github.com/snyk/rest-go-libs/blob/29e037aedc89128e0640cd33f1d59ce1068bafff/authz/README.md for entitlements?

pkg/utils/git/contributors.go
```diff
@@ -0,0 +1,94 @@
...
+		DetectDotGit: true,
+	})
+	if errors.Is(err, git.ErrRepositoryNotExists) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("open repository: %w", err)
+	}
+
+	head, err := repo.Head()
```
basti-snyk, line 38: This should probably be cached using something like commit hash as key and contributors as value. Else each call would potentially cause filesystem IOs.

pkg/utils/git/contributors.go
jonnyowenpowell, line 79:
```suggestion
    iter, err := repo.Log(&git.LogOptions{})
	if err != nil {
		return nil, fmt.Errorf("read log: %w", err)
	}
	defer iter.Close()

    authors := make(map[string]time.Time)
    
	iter.ForEach(func(c *object.Commit) error{
		when, email := c.Author.When, c.Author.Email

		if when.Before(since) {
			return nil
		}

		if prev, ok := authors[email]; ok && when.Before(prev) {
			return nil
		}

		authors[email] = when

		return nil
	})
```

jonnyowenpowell: ^ iterate through all commits on all branches, using authored date as cutoff and latest commit date value, and don't limit it to "now", i.e. count commits that are logged as being in the future

jonnyowenpowell:
Replaced by https://github.com/snyk/go-application-framework/pull/721

PR #673, opened 2026-07-29
test(networking): remove live snyk.io call from Test_GetHTTPClient

rrama:
### Description

`Test_GetHTTPClient` made a real request to `https://www.snyk.io` and then asserted on the response with a **non-fatal** `assert.Nil(t, err)`:

```go
response, err := client.Get("https://www.snyk.io")
assert.Nil(t, err)                        // non-fatal: execution continues on error
assert.Equal(t, 200, response.StatusCode) // response is nil on error -> nil deref
```

When the request fails, `response` is `nil` and the next line dereferences it. That panics the entire `pkg/networking` test binary with `invalid memory address or nil pointer dereference`, aborting every remaining test in the package rather than failing this one test.

This reproduces deterministically anywhere `www.snyk.io` is unreachable — restricted CI sandboxes, offline development, or a captive network. It was found by an agent running the suite in a network-restricted environment.

Two problems, both fixed here:

1. **Panic instead of a clean failure.** `require.NoError` now stops the test at the point of failure, so a transport error fails this single test with a useful message and the rest of the package still runs.
2. **A live external dependency in a unit test.** The request now goes to an `httptest` server, matching the ten other HTTP tests in this same file, all of which already use `httptest.NewServer`.

TLS and CA-trust behaviour remain covered by `Test_GetHTTPClient_EmptyCAs`, which stands up its own TLS listener with a self-signed cert, so no coverage is lost. The test now needs no network and completes in 0.00s instead of ~0.4s.

Verified that the failure mode is fixed by temporarily pointing the client at a closed port: the test now reports `FAIL ... connection refused` with no panic, and the remaining tests in the binary still execute.

Bonus: Did some similar housekeeping for other tests in the file.

### Checklist

  - `pkg/networking` passes at 89.7% coverage. One unrelated pre-existing failure exists on my machine: `TestDetectProxyConfig` in `connectivity_check_extension/connectivity` inherits the ambient `NODE_EXTRA_CA_CERTS` instead of isolating it, and passes with that variable unset. Untouched by this PR.
- [ ] Regenerated mocks, etc. (`make generate`) — n/a, no mocks or API clients affected
- [x] Linted (`make lint`) — 0 issues
- [ ] Test your changes work for the CLI — n/a, test-only change; no shipped code or public API is touched, so there is nothing for a consumer to pick up

Made with [Cursor](https://cursor.com)


pkg/networking/networking_test.go
```diff
@@ -55,9 +56,12 @@ func Test_HttpClient_CallingApiUrl_Unauthorized(t *testing.T) {
 		}
 	})
 	server := httptest.NewServer(handler)
+	t.Cleanup(server.Close)
```
rrama: I acknowledge that in Go 1.27 we can move to `httptest.NewTestServer(t, handler)` and this cleanup will be handled automatically. (Along with the other advantages `NewTestServer` brings)
But we aren't there just yet.

basti-snyk (approved):

PR #679, opened 2026-07-31
feat: Keep git-tracked files that a .gitignore rule would exclude

danskmt:
### Description

Git only ignores files it does not track, but `FileFilter` excluded any file matching a `.gitignore` rule. A file committed before the rule existed was therefore hidden from scans. This aligns the filter with git's behavior.

`GetRules` now records the git-tracked files that a `.gitignore` rule excludes, and `GetFilteredFiles` keeps them. Exclusions from `.snyk`, `.dcignore` and the built-in defaults still apply to tracked files, since those are Snyk's own exclusions rather than git's. That is resolved when the set is built, so filtering still works from the flat rule list it is handed.

Gated behind `FF_GITIGNORE_RESPECT_TRACKED_FILES`, default off. With the flag off, no config, or no readable git repository, behavior is unchanged. `GetRules` returns byte-identical rules either way, which a test asserts.

### Checklist

- [x] Linted (`make lint`)


danskmt:
Closing this PR in favor of https://github.com/snyk/go-application-framework/pull/680

PR #682, opened 2026-08-03
chore: report file filter duration metrics via analytics [CLI-1740]

octavian-snyk:
### Description

Report file filter duration metrics via analytics.

### Checklist

- [x] Linted (`make lint`)
  
  CLI PR: https://github.com/snyk/cli/pull/7076

---

> [!NOTE]
> **Low Risk**
> Changes are limited to optional analytics recording; filtering logic is unchanged when no recorder is configured, and behavior is heavily covered by new tests including concurrency.
> 
> **Overview**
> **FileFilter** now records analytics extension metrics when `WithMetrics` is set (invocation context passes **Analytics** automatically). Previously this was only stubbed plumbing.
> 
> Each **GetRules** and **GetFilteredFiles** run gets its own scope ID (`file-filter.<scopeID>.<metric>`) so repeated or concurrent runs do not overwrite keys. **GetRules** reports `rulesBuildDurationMs` plus metacharacter-fix and respect-tracked-files flags. **GetFilteredFiles** reports `durationMs`, `survivingFileCount`, `trackedFilesKeptCount`, and the same flags. Tracked-file preservation increments `trackedFilesKeptCount` only when a git-tracked file would be dropped by `.gitignore` but is kept and not excluded by other rules.
> 
> `RecorderFake` is mutex-protected for concurrent metric writes in tests. Tests cover scoped keys, concurrent runs, tracked-file counting, and that invocation analytics receives the expected extension keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d198f992a5450ce4b867f9d639ec1588114f4f97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -75,12 +75,14 @@ type DotSnykRule struct {
 }
 
 const (
-	metricFileFilterPrefix                = "file-filter"
-	metricFileFilterDurationMs            = "durationMs"
-	metricFileFilterFileCount             = "fileCount"
-	metricFileFilterMetacharacterFix      = "metacharacterFix"
-	metricFileFilterRespectTrackedFiles   = "respectTrackedFiles"
-	metricFileFilterTrackedFilesKeptCount = "trackedFilesKeptCount"
+	metricFileFilterPrefix = "file-filter" // prefix for all file-filter analytics keys
+
+	metricFileFilterDurationMs            = "durationMs"            // elapsed time for GetFilteredFiles: exclusion-predicate build (incl. git index read) and filtering, including caller drain of the result channel
+	metricFileFilterSurvivingFileCount    = "survivingFileCount"    // number of files that passed exclusion in this filter run
+	metricFileFilterRulesBuildDurationMs  = "rulesBuildDurationMs"  // elapsed time for GetRules: directory walk, ignore discovery, and buildGlobs
+	metricFileFilterMetacharacterFix      = "metacharacterFix"      // whether FF_FILE_FILTER_METACHARACTER_FIX was enabled for this run
+	metricFileFilterRespectTrackedFiles   = "respectTrackedFiles"   // whether FF_GITIGNORE_RESPECT_TRACKED_FILES was enabled for this run
+	metricFileFilterTrackedFilesKeptCount = "trackedFilesKeptCount" // tracked files kept despite a matching .gitignore rule
 )
 
 type FileFilterOption func(*FileFilter) error
@@ -206,6 +208,21 @@ func (fw *FileFilter) GetAllFiles() chan string {
 
 // GetRules builds a list of glob patterns that can be used to filter filesToFilter
 func (fw *FileFilter) GetRules(ruleFiles []string) ([]string, error) {
+	// Both feature flags change how rules are built (see buildGlobs), so the rules scope reports
+	// them alongside the build duration rather than relying on a GetFilteredFiles run that may
+	// never happen or may read a different flag value.
+	scopeID := newFileFilterMetricScopeID()
+	start := time.Now()
+	defer fw.recordMetricLazy(scopeID, metricFileFilterRulesBuildDurationMs, func() int {
+		return int(time.Since(start).Milliseconds())
+	})
+	fw.recordBoolMetricLazy(scopeID, metricFileFilterMetacharacterFix, func() bool {
+		return fw.config.GetBool(FF_FILE_FILTER_METACHARACTER_FIX)
+	})
+	fw.recordBoolMetricLazy(scopeID, metricFileFilterRespectTrackedFiles, func() bool {
+		return fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)
+	})
+
 	files := fw.GetAllFiles()
 
 	// iterate filesToFilter channel and find ignore filesToFilter
@@ -236,17 +253,6 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 
 	var filteredFilesCh = make(chan string)
 
-	var isFileExcluded func(string) bool
-	var trackedFilesKeptCount atomic.Int64
-	if fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES) {
-		isFileExcluded = fw.trackedFileExclusionPredicate(globs, &trackedFilesKeptCount)
-	} else {
-		globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
-		isFileExcluded = func(filePath string) bool {
-			return globPatternMatcher.MatchesPath(filePath)
-		}
-	}
-
 	go func() {
 		ctx := context.Background()
 		availableThreads := semaphore.NewWeighted(fw.max_threads)
@@ -254,6 +260,19 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 		var resolvedFileCount atomic.Int64
 
 		defer close(filteredFilesCh)
+
+		// Building the predicate is part of the run: with FF_GITIGNORE_RESPECT_TRACKED_FILES it
+		// opens the repository and reads the git index, so it is timed alongside the filtering.
+		var isFileExcluded func(string) bool
+		var trackedFilesKeptCount atomic.Int64
+		if fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES) {
+			isFileExcluded = fw.trackedFileExclusionPredicate(globs, &trackedFilesKeptCount)
+		} else {
+			globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
+			isFileExcluded = func(filePath string) bool {
+				return globPatternMatcher.MatchesPath(filePath)
+			}
+		}
 
 		// iterate the filesToFilter channel
 		for file := range filesCh {
@@ -286,7 +305,7 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 		fw.recordMetricLazy(runScopeID, metricFileFilterDurationMs, func() int {
 			return int(time.Since(start).Milliseconds())
 		})
-		fw.recordMetricLazy(runScopeID, metricFileFilterFileCount, func() int {
+		fw.recordMetricLazy(runScopeID, metricFileFilterSurvivingFileCount, func() int {
 			return int(resolvedFileCount.Load())
 		})
 		fw.recordMetricLazy(runScopeID, metricFileFilterTrackedFilesKeptCount, func() int {
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -44,7 +44,6 @@ type FileFilter struct {
 	dotSnykSections []DotSnykExcludeSectionName
 	config          configuration.Configuration
 	metricsRecorder metrics.Recorder
-	metricScopeID   string
 }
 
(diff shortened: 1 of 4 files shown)


danskmt (approved):

pkg/utils/file_filter.go
```diff
@@ -104,6 +105,33 @@ func WithConfig(config configuration.Configuration) FileFilterOption {
...
+	metricFileFilterRulesBuildDurationMs       = "file-filter.rulesBuildDurationMs"
+	metricFileFilterTrackedFilesKept           = "file-filter.trackedFilesKept"
+	metricFileFilterGitIndexReadDurationMs     = "file-filter.gitIndexReadDurationMs"
+	metricFileFilterTrackedFilesMapMemoryBytes = "file-filter.trackedFilesMapMemoryBytes"
+)
+
+// MetricsRecorder is satisfied by analytics.Analytics.
+type MetricsRecorder interface {
+	AddExtensionIntegerValue(key string, value int)
+}
```
danskmt: Just move the type and constants up, I think it is good. Some of these constants are still as placeholders for what is to come I see

pkg/utils/file_filter.go
```diff
@@ -66,6 +67,19 @@ type DotSnykRule struct {
...
```
danskmt: here we need to consider the other ones as well such as **AddExtensionStringValue**, etc.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,7 @@
+package metrics
+
+// MetricsRecorder is satisfied by analytics.Analytics.
+type MetricsRecorder interface {
+	AddExtensionIntegerValue(key string, value int)
+	AddExtensionStringValue(key string, value string)
+}
```
danskmt: I think there is also **AddExtensionBoolValue**

pkg/utils/file_filter.go
```diff
@@ -66,6 +70,16 @@ type DotSnykRule struct {
 	Exclude map[DotSnykExcludeSectionName]yaml.Node `yaml:"exclude"`
 }
 
+const (
+	metricFileFilterDurationMs                 = "file-filter.durationMs"
+	metricFileFilterRulesBuildDurationMs       = "file-filter.rulesBuildDurationMs"
```
PeterSchafer: Suggestion: Let's keep it simple for the start and add things while we go. This task doesn't need to add everything ever possible.

IMO File count and Duration is enough for the start.

PeterSchafer:
I think this will conflict quite a lot with https://github.com/snyk/go-application-framework/pull/685, let's separate the changes here to provide the infrastructure for adding metrics and later based on the linked PR add the actual metrics.

pkg/utils/file_filter.go
```diff
@@ -187,43 +234,63 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
-				fw.logger.Err(err).Msg("failed to limit threads")
-			}
-			go func(f string) {
-				defer availableThreads.Release(1)
-				// filesToFilter that do not match the glob pattern are filtered
-				if !globPatternMatcher.MatchesPath(f) {
-					filteredFilesCh <- f
-				}
-			}(file)
+		if fw.metrics == nil {
```
PeterSchafer: Suggestion: Let's avoid changing the business logic for metric collection. Maybe use a lazy resolution of metric values to avoid checking if fw.metrics is nil.

```
func (fw *FileFilter) recordMetricLazy(key string, getMetric func() int) {
	if fw.metrics != nil {
		fw.metrics.AddExtensionIntegerValue(fw.metricKey(key), getMetric())
	}
}
```

pkg/utils/file_filter.go
```diff
@@ -104,6 +119,31 @@ func WithConfig(config configuration.Configuration) FileFilterOption {
...
+	}
+}
+
+func (fw *FileFilter) recordMetric(key string, value int) {
+	if fw.metrics != nil {
+		fw.metrics.AddExtensionIntegerValue(fw.metricKey(key), value)
+	}
+}
+
+func (fw *FileFilter) metricKey(key string) string {
```
PeterSchafer: Based on our conversation, we might need to make this just some unique key actually rather then deriving it from path

pkg/utils/file_filter.go
```diff
@@ -272,6 +276,22 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+
+		fw.recordBoolMetricLazy(runScopeID, metricFileFilterMetacharacterFix, func() bool {
+			return fw.config.GetBool(FF_FILE_FILTER_METACHARACTER_FIX)
+		})
+		fw.recordBoolMetricLazy(runScopeID, metricFileFilterRespectTrackedFiles, func() bool {
+			return fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)
+		})
+		fw.recordMetricLazy(runScopeID, metricFileFilterDurationMs, func() int {
+			return int(time.Since(start).Milliseconds())
+		})
```
danskmt, line 307: This metric is not counting what happens in **GetRules** function. From my agent:

" **It excludes a large, genuinely-attributable part of the filter's cost.** GetRules() runs its own full filepath.WalkDir of the tree via GetAllFiles() (file_filter.go:226) plus all ignore-file reads and parsing, and trackedFileExclusionPredicate (git index open + match every index entry against the gitignore matcher) runs synchronously before start is taken. On this repo:

GetRules=90ms   GetFilteredFiles wall=70ms   reported durationMs=69

So ~55% of the real cost is invisible, and the part that is excluded — the git-index work — is precisely the code CLI-1411 added and that CLI-1740 presumably exists to evaluate. The tracked-files path costs ~2 ms here on a 440-entry index; on a monorepo index that is unbounded and unmeasured."

pkg/utils/file_filter.go
```diff
@@ -249,6 +250,8 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 	go func() {
 		ctx := context.Background()
 		availableThreads := semaphore.NewWeighted(fw.max_threads)
+		start := time.Now()
+		var resolvedFileCount atomic.Int64
 
 		defer close(filteredFilesCh)
```
danskmt, line 262: It seems we have to review what is this time we are computing, because from my agent:

"**durationMs** includes the downstream consumer's time. filteredFilesCh is unbuffered, so filteredFilesCh <- f (file_filter.go:268) blocks until the caller reads. start is set at the top of that same goroutine and time.Since(start) is taken after all workers drain. I verified this directly — a consumer that sleeps 100 ms per file:

files=6  durationMs=504   (consumer slept 600ms total)

The filter did ~0 ms of work; it reported 504 ms."

pkg/utils/file_filter.go
```diff
@@ -75,14 +74,13 @@ type DotSnykRule struct {
...
-//
-//nolint:unused // plumbing for a follow-up PR that wires these into recordMetricLazy/recordBoolMetricLazy call sites [CLI-1740]
 const (
-	metricFileFilterPrefix           = "file-filter"
-	metricFileFilterDurationMs       = "durationMs"
-	metricFileFilterFileCount        = "fileCount"
-	metricFileFilterMetacharacterFix = "metacharacterFix"
+	metricFileFilterPrefix                = "file-filter"
+	metricFileFilterDurationMs            = "durationMs"
+	metricFileFilterFileCount             = "fileCount"
```
danskmt: hmmm maybe we want better naming, what is this **fileCount**, i.e. we might want to be more explicit of what it means, i.e. "is it overall file count?", "is it only files that were tracked but being ignored?"..

pkg/utils/file_filter.go
```diff
@@ -272,6 +276,22 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+		})
+		fw.recordMetricLazy(runScopeID, metricFileFilterDurationMs, func() int {
+			return int(time.Since(start).Milliseconds())
+		})
+		fw.recordMetricLazy(runScopeID, metricFileFilterFileCount, func() int {
+			return int(resolvedFileCount.Load())
+		})
+		fw.recordMetricLazy(runScopeID, metricFileFilterTrackedFilesKeptCount, func() int {
+			return int(trackedFilesKeptCount.Load())
+		})
```
danskmt, line 313: each call emits for each scope ID. Is there any impact we should consider here?

pkg/utils/file_filter.go
```diff
@@ -272,6 +276,22 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 		if err != nil {
 			fw.logger.Err(err).Msg("failed to wait for all threads")
 		}
+
+		fw.recordBoolMetricLazy(runScopeID, metricFileFilterMetacharacterFix, func() bool {
+			return fw.config.GetBool(FF_FILE_FILTER_METACHARACTER_FIX)
+		})
+		fw.recordBoolMetricLazy(runScopeID, metricFileFilterRespectTrackedFiles, func() bool {
+			return fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)
+		})
```
danskmt, line 304: I believe we also might want to have metrics in **GetRules** as it calls:
```go
globs, err := fw.buildGlobs(ignoreFiles)
```
and **buildGlobs** checks the:
```go
enableMetacharacterFix := fw.config.GetBool(FF_FILE_FILTER_METACHARACTER_FIX)
respectGitIgnoreTrackedFiles := fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)
```

pkg/utils/file_filter.go
```diff
@@ -75,14 +74,13 @@ type DotSnykRule struct {
...
-	metricFileFilterDurationMs       = "durationMs"
-	metricFileFilterFileCount        = "fileCount"
-	metricFileFilterMetacharacterFix = "metacharacterFix"
+	metricFileFilterPrefix                = "file-filter"
+	metricFileFilterDurationMs            = "durationMs"
+	metricFileFilterFileCount             = "fileCount"
+	metricFileFilterMetacharacterFix      = "metacharacterFix"
+	metricFileFilterRespectTrackedFiles   = "respectTrackedFiles"
+	metricFileFilterTrackedFilesKeptCount = "trackedFilesKeptCount"
 )
```
danskmt, line 86: I wonder if we have to document these metrics in some way that explains what they are used for etc.. Or if just leaving by the file they are in, should be enough?

pkg/utils/file_filter.go
octavian-snyk: I'll add godoc for each metric.

pkg/utils/file_filter.go
octavian-snyk: It seems our options are:

Option 1 (Internal Time): Measure strictly the filter's own processing time.
Option 2 (End-to-End Time): Measure the full pass, including the time spent waiting while the caller drains results.

I'd lean toward Option 2. It keeps things simpler and reflects the real-world latency impact of our changes, rather than just an isolated metric.

What are your thoughts on this?

pkg/utils/file_filter.go
octavian-snyk: Each run gets a unique scope ID so concurrent or repeated runs don't overwrite each other in the analytics. I am not sure if we have some plugins which could generate a lot of noise here.

pkg/utils/file_filter.go
danskmt: I guess Option 2 works

danskmt (approved):

PR #685, opened 2026-08-04
feat: file filter doesn't exclude tracked files though .gitignore rules

CatalinSnyk:
### Description

### Checklist

- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI - https://github.com/snyk/cli/pull/7085

---

> [!NOTE]
> **Medium Risk**
> This changes which files are included in scans (core CLI behavior) when the flag is enabled, though it defaults off and falls back when Git/index access fails. Cross-source ignore ordering for tracked files is new and could surprise consumers if misconfigured.
> 
> **Overview**
> Adds **tracked-file-aware `.gitignore` handling** behind `FF_GITIGNORE_RESPECT_TRACKED_FILES` (CLI-1411). When enabled, rules from `.gitignore` are tagged with `#gitignore:` in `GetRules`; `GetFilteredFiles` uses a new exclusion predicate that reads the Git index and **does not apply `.gitignore` rules to indexed (tracked) paths**, while untracked paths still follow the full rule list.
> 
> **`.dcignore` and `.snyk` exclusions are unchanged** and still exclude tracked files, including cases where `.gitignore` would negate an ignore. If Git metadata is missing or the index cannot be opened, behavior falls back to the previous all-rules matcher.
> 
> Also clones the glob list before filtering (safe reuse of rules) and adds subdirectory scan-root support via `openGitIndex`, plus broad tests and benchmarks for the new path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9b1d30707cd64c10076edc9bab47172d08874c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

___

### **PR Type**
Enhancement, Bug fix

___

### **Description**
- Introduce tracked-file-aware .gitignore filtering.

- .gitignore rules now only exclude untracked files.

- Tracked files are always scanned unless explicitly excluded.

- Adds comprehensive tests for the new logic.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[FF_GITIGNORE_RESPECT_TRACKED_FILES enabled?] -->|Yes| B(Check Git Index);
  A -->|No| D(Apply all ignore rules);
  B -->|Untracked File| C(.gitignore filters untracked files);
  B -->|Tracked File| C;
  C --> E(Result: Filtered/Scanned);
  D --> E;
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>file_filter.go</strong><dd><code>Implement tracked-file-aware .gitignore filtering</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/utils/file_filter.go

<ul><li>Implements logic to distinguish between tracked and untracked files.<br> <li> Introduces <code>GitignoreGlobPrefix</code> for tagging gitignore rules.<br> <li> Adds functions <code>openGitIndex</code>, <code>findTrackedFilesToKeep</code>, <br><code>newFileExclusionCheck</code>.<br> <li> Modifies <code>GetFilteredFiles</code> and <code>buildGlobs</code> to incorporate new logic and <br>tagging.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/685/changes#diff-bf9226d9c8e73c07414a93ee5a315c13c9ad2be1c6ead9066e92b1adf6b296cd">+126/-3</a>&nbsp; </td>

</tr>
<tr>
  <td>
    <details>
      <summary><strong>file_filter_test.go</strong><dd><code>Add tests for tracked-file-aware gitignore filtering</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

pkg/utils/file_filter_test.go

<ul><li>Adds helper functions to initialize Git repos and create filters.<br> <li> Includes extensive test cases for the new .gitignore filtering logic,<br>  <br> covering flag states, file tracking, and different ignore file types.<br> <li> Tests <code>openGitIndex</code> fallback behavior and subdirectory scanning.</ul>

</details>

  </td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/685/changes#diff-0525df151b63831e042ffd8a2c7cab466a29f8d6f710cb726c359b6becffa89c">+370/-0</a>&nbsp; </td>

</tr>

</details>

___

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-git/go-git/v5"
+	gitindex "github.com/go-git/go-git/v5/plumbing/format/index"
 	"github.com/rs/zerolog"
 	gitignore "github.com/sabhiram/go-gitignore"
 	"golang.org/x/sync/semaphore"
@@ -25,6 +27,14 @@ const (
 	FF_FILE_FILTER_METACHARACTER_FIX   string = "internal_snyk_file_filter_metacharacter_fix_enabled"   // FF_FILE_FILTER_METACHARACTER_FIX (boolean) enables the fix for ignore rules and paths containing regex metacharacters
 	FF_GITIGNORE_RESPECT_TRACKED_FILES string = "internal_snyk_gitignore_respect_tracked_files_enabled" // FF_GITIGNORE_RESPECT_TRACKED_FILES (boolean) enables tracked-file-aware .gitignore filtering (CLI-1411)
 )
+
+// GitignoreGlobPrefix tags a glob returned by GetRules as sourced from a .gitignore file, so the
+// rule list carries its own provenance and git's exclusions can be told apart from other rules. The
+// leading "#" makes the line a comment to gitignore.CompileIgnoreLines, so the tag is inert for any
+// matcher that does not look for it.
+//
+// Tags are only emitted while FF_GITIGNORE_RESPECT_TRACKED_FILES is enabled.
+const GitignoreGlobPrefix = "#gitignore:"
 
 // by default, all rules are valid
 var defaultInvalidRules = []string{}
@@ -183,14 +193,15 @@ func (fw *FileFilter) GetRules(ruleFiles []string) ([]string, error) {
 // GetFilteredFiles returns a filtered channel of filepaths from a given channel of filespaths and glob patterns to filter on
 func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan string {
 	var filteredFilesCh = make(chan string)
+	globs = slices.Clone(globs)
 
-	// create pattern matcher used to match filesToFilter to glob patterns
-	globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
 	go func() {
 		ctx := context.Background()
 		availableThreads := semaphore.NewWeighted(fw.max_threads)
 
 		defer close(filteredFilesCh)
+
+		isExcluded := fw.newFileExclusionCheck(globs)
 
 		// iterate the filesToFilter channel
 		for file := range filesCh {
@@ -201,7 +212,7 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 			go func(f string) {
 				defer availableThreads.Release(1)
 				// filesToFilter that do not match the glob pattern are filtered
-				if !globPatternMatcher.MatchesPath(f) {
+				if !isExcluded(f) {
 					filteredFilesCh <- f
 				}
 			}(file)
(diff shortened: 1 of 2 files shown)


pkg/utils/file_filter.go
```diff
@@ -238,6 +356,11 @@ func (fw *FileFilter) buildGlobs(ignoreFiles []string) ([]string, error) {
 			globs = append(globs, parsedRules...)
 		} else { // .gitignore, .dcignore, etc. are just a list of ignore rules
 			parsedRules := parseIgnoreFile(content, filepath.Dir(ignoreFile), enableMetacharacterFix)
+			if tagGitignoreGlobs && filepath.Base(ignoreFile) == ".gitignore" {
+				for i, rule := range parsedRules {
+					parsedRules[i] = GitignoreGlobPrefix + rule
+				}
+			}
 			globs = append(globs, parsedRules...)
 		}
```
danskmt, line 412: I'm not sure if it is a good idea to rewrite rules... please double check if that doesn't cause any real issues.
Also, you might want to create a separate `if` clause since you are checking for .gitignore? or at least move `filepath.Base(ignoreFile)` to a variable

pkg/utils/file_filter.go
```diff
@@ -217,13 +228,120 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+func (fw *FileFilter) findTrackedFilesToKeep(gitignoreMatcher, nonGitignoreMatcher *gitignore.GitIgnore) map[string]struct{} {
+	index, scanRootPrefix, ok := fw.openGitIndex()
+	if !ok {
+		return nil
+	}
+
+	trackedFilesToKeep := map[string]struct{}{}
+	for _, entry := range index.Entries {
+		relPath, inScanRoot := strings.CutPrefix(entry.Name, scanRootPrefix)
+		if !inScanRoot {
```
PeterSchafer: Question: what is the example where files in the index are not in the scanroot?

pkg/utils/file_filter.go
```diff
@@ -217,13 +228,120 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+			continue
+		}
+
+		filePath := filepath.Join(fw.path, filepath.FromSlash(relPath))
```
PeterSchafer: Question: while I'm not sure yet why we have to now go back to absolute paths, I wonder if the Separator convention here is correct. Isn't matching happening on Slash paths?

pkg/utils/file_filter.go
```diff
@@ -217,13 +228,120 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+		if gitignoreMatcher.MatchesPath(filePath) && !nonGitignoreMatcher.MatchesPath(filePath) {
+			trackedFilesToKeep[relPath] = struct{}{}
+		}
+	}
+
+	return trackedFilesToKeep
+}
+
+func (fw *FileFilter) isTrackedFileToKeep(filePath string, trackedFilesToKeep map[string]struct{}) bool {
+	relPath, err := filepath.Rel(fw.path, filepath.FromSlash(filePath))
```
PeterSchafer, line 330: Question: a lot of back and forth with slashes, is there a potential to make this more consistent and save some operations?

PeterSchafer:
Let's investigate the performance here a bit more as well. Let's add a benchmark.

pkg/utils/file_filter.go
CatalinSnyk: The existing glob parser anchors every generated rule to the directory containing the ignore file. Since `GetAllFiles` and the ignore-file paths are filesystem paths, the candidate passed to the matcher must use the same rooted path form; matching only the repo-relative index name would not match those generated globs. `go-gitignore` normalizes the OS separator to `/` internally. The retained-set identity stays Git-style and slash-relative; native paths are used only at the filesystem/matcher boundary.

pkg/utils/file_filter.go
CatalinSnyk: A concrete example is scanning `/repo/packages/app` while the Git repository root is `/repo`. `DetectDotGit` opens `/repo/.git/index`, whose entries also include `/repo/packages/other`, repository-level files, and every other tracked path. `scanRootPrefix` limits the index entries to `packages/app/` before matching or retaining anything.

pkg/utils/file_filter.go
CatalinSnyk: Thanks. The rewrite is intentional and follows the pitch: `#gitignore:` is provenance metadata, and `go-gitignore` treats the leading `#` as a comment, so it is inert for ordinary consumers. Tags are emitted only while the feature flag is enabled; with the flag off, `GetRules` remains byte-identical. `GetFilteredFiles` detects and removes the reserved prefix for the Git-only matcher. Tests assert that only `.gitignore` rules are tagged and that stripping the tags reproduces the previous rules. I agree that the source distinction should be visually clearer; I will follow up by making the `.gitignore` handling explicit rather than hiding it in the generic branch.

pkg/utils/file_filter.go
CatalinSnyk: There are two intentional representations: Git index entries and retained-set keys are scan-root-relative with `/`, while filesystem paths are native and rooted because that is what the existing generated globs match against. `FromSlash` is used once when building the matcher candidate; `Rel` plus `ToSlash` converts walker output back to the Git key for lookup. The new benchmark shows the steady-state path check is effectively unchanged by the flag (about 72-73 ms per 10k files); the material cost is initial index decoding and matching. I would keep these boundary conversions explicit unless we also change the existing glob representation.

pkg/utils/file_filter.go
```diff
@@ -217,13 +228,120 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+	trackedFilesToKeep := fw.findTrackedFilesToKeep(gitignoreMatcher, nonGitignoreMatcher)
+
+	return func(filePath string) bool {
+		if nonGitignoreMatcher.MatchesPath(filePath) {
+			return true
+		}
+
+		return gitignoreMatcher.MatchesPath(filePath) &&
+			!fw.isTrackedFileToKeep(filePath, trackedFilesToKeep)
+	}
```
danskmt, line 308: @CatalinSnyk Please check this scenario, I believe it currently fails:

**Scenario 15**: A `.gitignore` negation of a rule from another ignore source ⚠️

**Triggered by actions**:
  1. `.dcignore` contains `gen/*.js`; `.gitignore` contains `!gen/keep.js`.
  2. Create `gen/keep.js` and `gen/drop.js`; **track nothing**.
  3. `snyk code test <root>`.

**Expected results**: identical in both flag states. Nothing is tracked, so the feature must be
a complete no-op.

**Cause**: the implementation splits one matcher into two and treats a non-`.gitignore` match as
final, so a `.gitignore` negation can no longer rescue a `.dcignore` match. This is a
behaviour change unrelated to tracked files.

pkg/utils/file_filter_test.go
```diff
@@ -1697,3 +1701,471 @@ func TestDotSnykExclude_isExpired(t *testing.T) {
...
+				trackedPaths := benchmarkTrackedPaths(indexSize)
+				initGitRepoWithTrackedFiles(b, root, trackedPaths)
+
+				fileFilter := newTrackedFilesFilter(root, enabled)
+				rules, err := fileFilter.GetRules([]string{".gitignore"})
+				if err != nil {
+					b.Fatal(err)
+				}
+
+				check := fileFilter.newFileExclusionCheck(rules)
```
danskmt: these names are not great.
- variable name **check** and 
- the function name **newFileExclusionCheck**

Please give them better names of what they are actually evaluating / returning

pkg/utils/file_filter.go
```diff
@@ -217,13 +228,120 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+}
+
 // buildGlobs iterates a list of ignore filesToFilter and returns a list of glob patterns that can be used to test for ignored filesToFilter
 func (fw *FileFilter) buildGlobs(ignoreFiles []string) ([]string, error) {
 	if len(ignoreFiles) == 0 {
 		return nil, nil
 	}
 
 	enableMetacharacterFix := fw.config.GetBool(FF_FILE_FILTER_METACHARACTER_FIX)
+	tagGitignoreGlobs := fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)
```
danskmt: change the variable name to reflect better the feature flag name. Here you are using for example 'globs' but I would suggest to stick closer to the feature flag name, e.g. `respectGitignoreTrackedFiles` or something similar

pkg/utils/file_filter.go
```diff
@@ -183,15 +193,16 @@ func (fw *FileFilter) GetRules(ruleFiles []string) ([]string, error) {
...
 
-	// create pattern matcher used to match filesToFilter to glob patterns
-	globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
 	go func() {
 		ctx := context.Background()
 		availableThreads := semaphore.NewWeighted(fw.max_threads)
 
 		defer close(filteredFilesCh)
 
+		isExcluded := fw.newFileExclusionCheck(globs)
```
danskmt: I wonder if instead of always computing this, should check the value of:
`fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)`

and if is false, to use the original function as before, so behavior unchanged `globPatternMatcher.MatchesPath`.
Right now it seems the flow is going through the new implementation even with the flag off?

Also please change variable name and this function name to be more consistent of what it does. I see three different variable names for this function in the code: `isExcluded`, `check` and `timedCheck`

pkg/utils/file_filter.go
CatalinSnyk: Should be fixed (and tested) in [f4beeeb](https://github.com/snyk/go-application-framework/pull/685/commits/f4beeebc99c30568c86598096758b379a12ec7cd)

pkg/utils/file_filter_test.go
CatalinSnyk: Done!

pkg/utils/file_filter.go
CatalinSnyk: done

pkg/utils/file_filter.go
```diff
@@ -26,6 +28,14 @@ const (
 	FF_GITIGNORE_RESPECT_TRACKED_FILES string = "internal_snyk_gitignore_respect_tracked_files_enabled" // FF_GITIGNORE_RESPECT_TRACKED_FILES (boolean) enables tracked-file-aware .gitignore filtering (CLI-1411)
 )
 
+// GitignoreGlobPrefix tags a glob returned by GetRules as sourced from a .gitignore file, so the
+// rule list carries its own provenance and git's exclusions can be told apart from other rules. The
+// leading "#" makes the line a comment to gitignore.CompileIgnoreLines, so the tag is inert for any
+// matcher that does not look for it.
+//
+// Tags are only emitted while FF_GITIGNORE_RESPECT_TRACKED_FILES is enabled.
+const GitignoreGlobPrefix = "#gitignore:"
```
PeterSchafer, line 39: please unexport

pkg/utils/file_filter.go
```diff
@@ -217,13 +229,117 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
...
+
+	trackedFilePaths := make([]string, 0, len(index.Entries))
+	for _, entry := range index.Entries {
+		relPath, inScanRoot := strings.CutPrefix(entry.Name, scanRootPrefix)
+		if inScanRoot {
+			trackedFilePaths = append(trackedFilePaths, relPath)
+		}
+	}
+
+	return trackedFilePaths
```
PeterSchafer, line 326: Question: From my understanding trackedFilePaths contains all tracked file paths which is unbound, but we actually only care about the tracked files that are also in the gitignore file mentioned. so there could be a possibiity to improve memory performance here and later when matching files a performance improvement as we only go through a subset of files.

danskmt:
@CatalinSnyk closing this PR in favor of the other one you created & merged

PR #686, opened 2026-08-04
feat(networking): opt-in network retry resilience [IDE-1890]

basti-snyk:
### Description

Everything for [IDE-1890](https://snyksec.atlassian.net/browse/IDE-1890), consolidated into one PR. All of it sits behind a single opt-in — `configuration.NETWORK_REQUEST_RETRIES_ENABLED`. **With that unset, behaviour is byte-identical to today**, pinned by regression tests at every layer.

Four pieces, in dependency order:

**1. The opt-in itself** ([IDE-2412](https://snyksec.atlassian.net/browse/IDE-2412)) — new `NETWORK_REQUEST_RETRIES_ENABLED` config key. When set, the retry middleware's default attempt count becomes 3 without also requiring `PREVIEW_FEATURES_ENABLED`. An explicitly configured attempt count still wins over both.

**2. Transport-error retry** ([IDE-2419](https://snyksec.atlassian.net/browse/IDE-2419)) — today every transport error is wrapped in `backoff.Permanent` and never retried, whatever the config says. Now `ECONNRESET` and network timeouts (including DNS timeouts) are retried; DNS `NotFound`, context cancellation/deadline, and TLS/auth failures are not.

Two details worth a reviewer's attention: the deny-check runs *before* the timeout allow-check, because `context.DeadlineExceeded` also satisfies `net.Error.Timeout()` and a caller-abandoned request must never be retried. And Windows `WSAECONNRESET` (10054) is matched explicitly — Go does not map it onto `syscall.ECONNRESET`, so a Unix-only check would pass CI on Linux while silently never retrying on Windows.

Retry eligibility is method- and configuration-based. Safe methods retry by default; mutating methods retry only when their path matches `NETWORK_REQUEST_RETRY_ALLOWED_PATHS`. The default includes the monitor endpoint paths needed by current consumers, and applications such as the CLI can override the list through Configuration. Matching uses contiguous full path segments, so similarly named segments are unaffected.

**3. Truncated-response recovery** — the part that fixes the reported customer. Behind an SSL-inspecting corporate proxy, the upstream HTTP/2 stream is reset mid-body *after* `RoundTrip` has already returned `200 OK`. The truncated body is then parsed as an empty result and the scan reports zero vulnerabilities on a project that has them, roughly 1 session in 5. Because `RoundTrip` returns before the body is read, no amount of retry *configuration* reaches this — which is why pieces 1 and 2 alone did not fix it.

The fix reads retryable 2xx JSON bodies inside the existing retry loop; a truncated read then becomes an ordinary retryable error the existing `backoff.Retry` already handles. The integrity-check buffer has a dedicated 4 MiB cap. Larger JSON bodies preserve the inspected prefix and continue streaming from the original body, while non-JSON responses remain untouched. Three attempts take session success from ~80% to ~99%.

**4. Review fixes** — the opt-in predicate is extracted to `internal/utils.NetworkRetriesEnabled` so the two call sites cannot drift, and a nil-guarded `drainAndClose` closes a response-body leak on the retried transport-error path.

Design context: [IDE-2050](https://snyksec.atlassian.net/wiki/spaces/IDE/pages/4888789043). CLI flag wiring is deferred to [IDE-2415](https://snyksec.atlassian.net/browse/IDE-2415).

### Automated review findings

The accumulated review findings have been addressed or evaluated against the final design:

| Finding | Disposition |
|---|---|
| DRY: opt-in predicate duplicated across two files | **Fixed** — extracted to `internal/utils`, not exported, to avoid growing GAF's public API |
| Response body leak on retried transport error | **Fixed** — note the suggested patch used a bare `drainAndClose(response.Body)`, which panics on the common nil-response transport-error path; the guard checks the response, not just the body |
| Retry safety for mutating HTTP methods | **Fixed** — safe methods retry by default; mutating methods require a configured allowed path, which preserves the customer POST use case without replaying arbitrary writes |
| Test constant duplicates the retry-after key | **Fixed** — the middleware owns an exported key alongside its existing exported attempt-count key; tests now use that single definition |

### Known limits, stated deliberately

- With the opt-in unset, a truncated response still silently reports zero vulnerabilities. Required, and accepted.
- `PREVIEW_FEATURES_ENABLED` also activates all of this, matching the pre-existing attempt-count default. Intentional and tested.
- Opted-in consumers inspect up to 4 MiB of retryable 2xx JSON bodies inside the retry loop. Larger bodies continue streaming after the buffered prefix, so truncation beyond the cap cannot be retried.
- Mutating requests are retried only when their paths match the configured allowlist; consumers can override the defaults through Configuration. Blank allowlist entries are ignored so persisted JSON values cannot match every absolute path.
- `getErrorList` swallows the identical read error on non-2xx responses. A real adjacent gap, left alone because changing it would alter flag-off behaviour.

### Size

~1380 added lines across 9 files, well over the 700-line guideline. Deliberate, and consolidated on request: this is one coherent opt-in feature whose pieces do not stand alone — 2 and 3 are inert without 1, and 3 is the only one that fixes the customer. The large majority is a three-layer test pyramid for correctness-sensitive retry-safety; production code is a small fraction of the diff.

### Test plan

- [x] Acceptance: real engine + real TCP server forcing RST — opted-in caller recovers, non-opted-in caller sees today's behaviour exactly
- [x] Acceptance: truncated-body recovery, with the flag-off guard confirmed GREEN on baseline *before* any production code was written
- [x] Integration: real wiring, including a test that fails if the opt-in OR-condition regresses; monitor path never buffered or retried; non-JSON streams unbuffered; exhausted attempts yield a non-nil error *and* a non-nil response
- [x] Unit: error-axis matrix, deny-before-allow ordering, Windows errno, content-type helper, opt-in truth table
- [x] `make format && make lint && make test && make generate`; `snyk code test` — no new findings

### Checklist

- [ ] Full `make test` currently reports two unrelated repository/environment failures (`Test_UfmPresenter_HumanReadable` fixture drift and `TestDetectProxyConfig` due the sandbox CA environment); the changed acceptance, integration, and unit paths pass
- [x] Linted (`make lint`)

[IDE-1890]: https://snyksec.atlassian.net/browse/IDE-1890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDE-2412]: https://snyksec.atlassian.net/browse/IDE-2412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDE-2419]: https://snyksec.atlassian.net/browse/IDE-2419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDE-2050]: https://snyksec.atlassian.net/browse/IDE-2050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes shared HTTP client retry semantics and can re-send POSTs on allowlisted paths; incorrect allowlist or buffering could affect scan results or duplicate side effects, though monitor paths are excluded and the feature is opt-in.
> 
> **Overview**
> Adds an **opt-in resilient network retry policy** behind `NETWORK_REQUEST_RETRIES_ENABLED` (or existing preview features), with **flag-off behavior unchanged** from today.
> 
> **Configuration:** New keys for the opt-in flag and a **path allowlist** for retries on unsafe HTTP methods (defaults like `test-dep-graph`, `verify/token`, `feature_flags/evaluation`). Default max attempts (3 vs 1) now follows shared `NetworkRetriesEnabled`; env CSV and persisted JSON arrays are normalized for the allowlist. `GetStringSlice` now handles JSON `[]interface{}` from config files.
> 
> **Retry middleware (when opted in):** Retries **transient transport errors** (e.g. connection reset, network timeouts) instead of always treating RoundTripper errors as permanent. **Status-code retries** on POST/etc. are limited to allowlisted path segments; safe methods still retry broadly. **Successful JSON responses** up to ~4MB are buffered in the retry loop so truncated bodies after `200 OK` can trigger another attempt; monitor-style paths stay excluded.
> 
> **Tests:** Large integration/unit coverage for opt-in gates, path rules, transport vs truncation recovery, and config wiring through the real app engine.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbf8bd1fb57c2c41a6a6f5935e4e7c87dcc72079. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

___

### **PR Type**
Enhancement

___

### **Description**
- Introduce opt-in network retry mechanism.

- Handle transient transport errors (resets, timeouts).

- Retry on truncated JSON responses.

- Configure retryable paths via configuration.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A[New Opt-in Flag] --> B{Retry Logic};
  C[Transport Errors (resets/timeouts)] --> B;
  D[Truncated JSON Responses] --> B;
  E[Request Path Allowlist] --> B;
  F[Safe Methods (GET/HEAD)] --> B;
  G[Existing Retry Logic (429, etc.)] --> B;
  B --> H[HTTP Client];
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>network-retry.go</strong><dd><code>Add utility for checking network retry enablement.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-5bbcbdb8d18f41c27d9a36cee7e76aec5faa9386fedcce4dea66c73d4e7a9648">+9/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>app.go</strong><dd><code>Integrate network retry opt-in and default paths.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-63facfa8b028c9f397524787fac074737e5534ae2591640c1a029549ec3d7450">+32/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>configuration.go</strong><dd><code>Support []interface{} for GetStringSlice.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-a0e28739a3c85fb2099ec7cdf95f0579f5f84488cca12aee25fb2e62702db5d6">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>constants.go</strong><dd><code>Define new constants for network retries.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-b346bdeac320c8fe8bc31a635e59da3e8bf59c7c00ed11bf8cd6a17dccda15b7">+12/-10</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>netstack_error_handler.go</strong><dd><code>Add Windows WSAECONNRESET to connection reset errors.</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-7dacd2506c6cf8b1baea777de2e13b2efe89f4ebb92006cba62ab63741e90837">+7/-3</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>retry_middleware.go</strong><dd><code>Implement opt-in transport error and truncation retries.</code>&nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-295d20b71bb7f7fdc47d3503d7fb17595142bb2c48dd5801cc4ad0517afb1a55">+105/-5</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>retry_transport_error.go</strong><dd><code>Define retryability logic for transport errors and requests.</code></dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-4b4a89a71c3679f899fc9a03595befb77c5a313a6ff49a89c616ecc068d64a26">+62/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>6 files</summary><table>
<tr>
  <td><strong>network-retry_test.go</strong><dd><code>Test network retry enablement utility.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-4babd6cb6294f642d046da6b5b7948e8a6e7f7aba13a35b80761966b1c62c34c">+33/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>app_test.go</strong><dd><code>Add tests for network retry configuration.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-c7d8cc80ef42a7d06d8053179322063692c2714db211e9393be0114fd9744c7a">+759/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>configuration_test.go</strong><dd><code>Test GetStringSlice for []interface{}.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-b4d3be1c834b6eb56ae18ccc5f5aedc682c4835a6db7044c71edf778a0534632">+33/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>netstack_error_handler_test.go</strong><dd><code>Test WSAECONNRESET error handling.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-099eccc3cc0194c993e4af8cf9617e0ef09d2fff10f219ec4e7054d58713ab42">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>retry_middleware_test.go</strong><dd><code>Add extensive tests for new retry middleware.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-781d13f3ad7692fdffe97f224e749e9e4cbbab29fbefafcfb11f78de8772aa26">+924/-19</a></td>

</tr>

<tr>
  <td><strong>retry_transport_error_test.go</strong><dd><code>Test retryable transport error and request logic.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-e6cc55b69c642f8dce4f52e7e8286d9ec12b06b19513c267b93e361159cc34aa">+201/-0</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>authHost.go</strong><dd><code>Clarify configuration key behavior in comments.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-fa935165069f215250be111f798aa2105409a63f1762396feb48caf5990f1098">+5/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>oauth2authenticator.go</strong><dd><code>Clarify configuration key behavior in comments.</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/snyk/go-application-framework/pull/686/changes#diff-acc5a39a076353bcd08feb6612f6ffd2a5206c30de023748a5ed97150d7a7ea4">+4/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr></tr></tbody></table>

</details>

___


pkg/networking/middleware/retry_transport_error.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	var netErr net.Error
+	return errors.As(err, &netErr) && netErr.Timeout()
+}
+
+func isRetryableRequest(req *http.Request) bool {
+	if req.Body != nil && req.Body != http.NoBody && req.GetBody == nil {
+		return false
+	}
+	return !isMonitorPath(req.URL.Path)
```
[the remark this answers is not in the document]
basti-snyk: Not applying this — it asks to reinstate a guard that was deliberately removed.

The method/idempotency-key check existed in an earlier revision of this PR and was taken out by an explicit product decision. The reason is concrete: the customer failure this whole initiative exists for is `POST /v1/test-dep-graph`. Under the suggested rule that call is never retried, so the change would ship without helping the customer it was written for.

`/v1/test-dep-graph` is a query expressed as a POST — it sends a dep-graph and returns vulnerabilities, and creates nothing server-side. Retrying it duplicates no state. The general "POST is unreplayable" rule that `net/http.Transport` applies is correct for *unknown* requests; here the set of endpoints is known.

The one endpoint that genuinely creates a resource is monitor, and that is exactly what `isMonitorPath` excludes. That is the whole policy, deliberately: retries are opt-in, and when a user opts in they get retries rather than a method-based refusal.

The duplicate-write risk on other non-idempotent endpoints is real and accepted as the cost of the opt-in. Recorded in ADR-1 (amended) and on IDE-2419.

pkg/app/app_test.go
```diff
@@ -30,10 +40,117 @@ import (
...
 	pkgMocks "github.com/snyk/go-application-framework/pkg/mocks"
+	"github.com/snyk/go-application-framework/pkg/networking/middleware"
 	"github.com/snyk/go-application-framework/pkg/runtimeinfo"
+	"github.com/snyk/go-application-framework/pkg/utils"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
+// tests that exercise a retry must set this small, or backoff sleeps for real
+// seconds. Mirrors middleware's unexported configurationKeyRetryAfter.
+const networkRequestRetryAfterSecondsKey = "internal_network_request_retry_after_seconds"
```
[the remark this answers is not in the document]
basti-snyk: Acknowledged, not changing it. The duplication is real and the failure mode you describe (tests silently fall back to the 5s backoff and get slow rather than failing loudly) is accurate.

Leaving it because the fix is to export `ConfigurationKeyRetryAfter` from `pkg/`, and exported GAF API is a permanent contract that downstream consumers pin against. Growing that surface for a test-only convenience isn't a trade we want to make; the key is unexported today on purpose. If a consumer ever needs to tune retry-after, exporting it then will be a deliberate API decision rather than a side effect of test wiring.

internal/constants/constants.go
```diff
@@ -20,3 +20,6 @@ const SNYK_DEFAULT_ALLOWED_HOST_REGEXP = `^(https?://)?api(\.(.+))?\.(snyk|snykg
 // SNYK_DEFAULT_ALLOWED_HOST_DOMAINS is the default allowlist of registrable
 // domains that an OAuth callback instance host is permitted to belong to.
 var SNYK_DEFAULT_ALLOWED_HOST_DOMAINS = []string{"snyk.io", "snykgov.io"}
+
+// Shared between pkg/app's default-value registration and pkg/networking/middleware's fallback so the two cannot drift.
+var DEFAULT_RETRY_ALLOWED_PATHS = []string{"test-dep-graph", "verify/token", "feature_flags/evaluation"}
```
basti-snyk, line 25: Just as a note: I'm aware these defaults are in the Snyk services domain and thus maybe should excluded from GAF, as GAF is a library. On the other hand these are just defaults. Happy to remove if wanted.

PeterSchafer: Just make them configurable from the CLI please.

pkg/app/app_test.go
```diff
@@ -1120,3 +1319,544 @@ func Test_defaultMaxNetworkRequestAttempts(t *testing.T) {
...
+	engine := CreateAppEngineWithOptions(WithConfiguration(config))
+	client := engine.GetNetworkAccess().GetUnauthorizedHttpClient()
+
+	resp, err := client.Get(server.URL)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	assert.Equal(t, http.StatusServiceUnavailable, resp.StatusCode)
+	assert.Equal(t, int32(3), atomic.LoadInt32(requestCount))
+}
```
rrama, line 1452: Nitpick: the 4 tests above are quite similar, consider converting into a table test like this:

```go
func Test_NetworkRetryOptIn(t *testing.T) {
	tests := []struct {
		name            string
		statusSequence  []int
		retriesEnabled  bool
		requestAttempts int // 0 = leave unset
		wantStatus      int
		wantCount       int32
	}{
		{
			name:           "not opted in, single attempt",
			statusSequence: []int{http.StatusServiceUnavailable, http.StatusOK},
			wantStatus:     http.StatusServiceUnavailable,
			wantCount:      1,
		},
		{
			name:            "explicit attempt count wins over opt-in",
			statusSequence:  []int{http.StatusServiceUnavailable, http.StatusOK},
			retriesEnabled:  true,
			requestAttempts: 1,
			wantStatus:      http.StatusServiceUnavailable,
			wantCount:       1,
		},
		{
			name:           "non-retryable response not retried",
			statusSequence: []int{http.StatusNotFound},
			retriesEnabled: true,
			wantStatus:     http.StatusNotFound,
			wantCount:      1,
		},
		{
			name:           "gives up after policy limit",
			statusSequence: []int{http.StatusServiceUnavailable},
			retriesEnabled: true,
			wantStatus:     http.StatusServiceUnavailable,
			wantCount:      3,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			server, requestCount := newSequencedStatusServer(t, tt.statusSequence)

			config := configuration.NewWithOpts()
			if tt.retriesEnabled {
				config.Set(configuration.NETWORK_REQUEST_RETRIES_ENABLED, true)
			}
			if tt.requestAttempts != 0 {
				config.Set(middleware.ConfigurationKeyRequestAttempts, tt.requestAttempts)
			}
			config.Set(middleware.ConfigurationKeyRetryAfter, 1)

			engine := CreateAppEngineWithOptions(WithConfiguration(config))
			client := engine.GetNetworkAccess().GetUnauthorizedHttpClient()

			resp, err := client.Get(server.URL)
			require.NoError(t, err)
			defer resp.Body.Close()

			assert.Equal(t, tt.wantStatus, resp.StatusCode)
			assert.Equal(t, tt.wantCount, atomic.LoadInt32(requestCount))
		})
	}
}
```

pkg/app/app_test.go
```diff
@@ -1120,3 +1319,544 @@ func Test_defaultMaxNetworkRequestAttempts(t *testing.T) {
...
+
+func Test_NetworkRetryOptIn_AllowedPathsFromPersistedJSONConfig(t *testing.T) {
+	fakehome := t.TempDir()
+	t.Setenv("HOME", fakehome)
+	t.Setenv("USERPROFILE", fakehome)
+
+	configFile, err := configuration.CreateConfigurationFile("gaf-retry-test.json")
+	require.NoError(t, err)
+	require.NoError(t, os.WriteFile(configFile, []byte(`{"internal_network_request_retry_allowed_paths": ["custom-retryable"]}`), 0600))
```
rrama: Nitpicks: Remove the hardcodings of the internal key:
```go
require.NoError(t, os.WriteFile(configFile, fmt.Appendf(nil,
    `{%q: ["custom-retryable"]}`, configuration.NETWORK_REQUEST_RETRY_ALLOWED_PATHS), 0600))
```

pkg/app/app_test.go
```diff
@@ -1120,3 +1319,544 @@ func Test_defaultMaxNetworkRequestAttempts(t *testing.T) {
...
+	req, err := http.NewRequest(http.MethodPost, server.URL+"/v1/monitor/npm", bytes.NewReader([]byte(`{}`)))
+	require.NoError(t, err)
+
+	resp, err := client.Do(req)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	assert.Equal(t, http.StatusServiceUnavailable, resp.StatusCode)
+	assert.Equal(t, int32(1), atomic.LoadInt32(requestCount))
+}
```
rrama, line 1648: Nitpick: can table test with the test above:

```go
func Test_NetworkRetryOptIn_AllowedPathsFromPersistedJSONConfig(t *testing.T) {
	tests := []struct {
		name        string
		allowlist   []string
		path        string
		wantStatus  int
		wantRetried bool
	}{
		{"declared path is retried", []string{"custom-retryable"}, "/v1/custom-retryable", http.StatusOK, true},
		{"blank entry allows nothing", []string{""}, "/v1/monitor/npm", http.StatusServiceUnavailable, false},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			fakehome := t.TempDir()
			t.Setenv("HOME", fakehome)
			t.Setenv("USERPROFILE", fakehome)

			configFile, err := configuration.CreateConfigurationFile("gaf-retry-test.json")
			require.NoError(t, err)
			contents, err := json.Marshal(map[string][]string{
				configuration.NETWORK_REQUEST_RETRY_ALLOWED_PATHS: tt.allowlist,
			})
			require.NoError(t, err)
			require.NoError(t, os.WriteFile(configFile, contents, 0600))

			server, requestCount := newSequencedStatusServer(t, []int{http.StatusServiceUnavailable, http.StatusOK})

			config := configuration.NewWithOpts(
				configuration.WithFiles("gaf-retry-test"),
				configuration.WithSupportedEnvVarPrefixes("snyk_", "internal_"),
				configuration.WithCachingEnabled(configuration.NoCacheExpiration),
			)
			config.Set(configuration.NETWORK_REQUEST_RETRIES_ENABLED, true)
			config.Set(middleware.ConfigurationKeyRetryAfter, 1)

			engine := CreateAppEngineWithOptions(WithConfiguration(config))
			client := engine.GetNetworkAccess().GetUnauthorizedHttpClient()

			req, err := http.NewRequest(http.MethodPost, server.URL+tt.path, bytes.NewReader([]byte(`{}`)))
			require.NoError(t, err)

			resp, err := client.Do(req)
			require.NoError(t, err)
			defer resp.Body.Close()

			assert.Equal(t, tt.wantStatus, resp.StatusCode)
			if tt.wantRetried {
				assert.GreaterOrEqual(t, atomic.LoadInt32(requestCount), int32(2))
			} else {
				assert.Equal(t, int32(1), atomic.LoadInt32(requestCount))
			}
		})
	}
}
```

pkg/configuration/configuration.go
```diff
@@ -644,6 +644,12 @@ func (ev *extendedViper) GetStringSlice(key string) []string {
 	switch v := result.(type) {
 	case []string:
 		return v
+	// JSON arrays read back from a config file are decoded as []interface{}.
+	case []interface{}:
+		for _, item := range v {
+			output = append(output, fmt.Sprintf("%v", item))
+		}
+		return output
```
rrama, line 652: This change will make this comment a lie: https://github.com/snyk/go-application-framework/blob/a23b7612120268c299ab98ad62f628fa23e784a1/pkg/auth/authHost.go#L62-L65 Please remove it (and check for others like it).

pkg/networking/middleware/retry_middleware.go
```diff
@@ -98,6 +113,15 @@ func drainAndClose(body io.ReadCloser) {
 	_ = body.Close()
 }
 
+func isJSONResponse(response *http.Response) bool {
+	mimeType := strings.ToLower(strings.TrimSpace(strings.Split(response.Header.Get("Content-Type"), ";")[0]))
```
rrama: Consider `mime.ParseMediaType`.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -177,6 +203,23 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 
 		// errors from the next round tripper cannot be retried
 		if rtErr != nil {
+			attemptLimit := maxAttempts
+			if cachedMaxRetries != nil {
```
rrama, line 241: A code comment explaining what is happening and what is being overriden so the reader doesn't have to scroll down to understand would be nice.
Something like:
```go
// If on the non-first loop, a previous response dictates that we use a different max limit, use that instead.
```
Then I know I need to go looking later in the code.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -186,8 +229,9 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 			cachedMaxRetries = &calculated
```
rrama, line 264: While I get it, this is just very confusing and non-obvious as to what it is doing.
Also, if someone sets a retry limit of 5, then this code runs on a 429 it will bump them down to 3.

rrama: I have a refactoring proposal (probably for a follow-up, not this PR), just use `maxAttempt` as the source of truth and override it directly, then just code comment `maxAttempt` to explain where its values come from.

I got AI to write up what I am thinking:

> Right now the effective attempt limit is threaded through three things: the baseline `maxAttempts`, the `*int cachedMaxRetries`, and the recomputed local `attemptLimit` in the transport-error block (pkg/networking/middleware/retry_middleware.go:206-209). `cachedMaxRetries` being a pointer is doing double duty: it's both the effective value and the "have we computed it yet?" flag (`nil` = not yet).
> 
> I think this collapses to a single mutable value plus an explicit flag:
> 
> - Keep `maxAttempts` as the one source of truth, mutated in place.
> - Replace the `if cachedMaxRetries == nil` guard (pkg/networking/middleware/retry_middleware.go:227-230) with a `gotMaxRetryAttemptFromResponse` bool, set `true` in the body:
>     ```go
>     if !gotMaxRetryAttemptFromResponse {
>         maxAttempts = getMaxRetryAttempts(response, maxAttempts)
>         gotMaxRetryAttemptFromResponse = true
>     }
>     ```
> - Then the transport-error block just reads `maxAttempts` directly — `attemptLimit` and the nil-check become dead code and go away.
> `getMaxRetryAttempts` returns `override ?? maxAttempts`, so applying the override back onto `maxAttempts` under the one-shot guard is safe (the call still sees the baseline on its single invocation). All the downstream `*cachedMaxRetries` reads (pkg/networking/middleware/retry_middleware.go:234, :246, :259, :262) become plain `maxAttempts`.
> 
> Non-blocking, and I realise `cachedMaxRetries` predates this PR — this touches existing code the change is built on, so happy to punt it to a follow-up if you'd rather keep this PR scoped. But since the new transport-error path adds a third view of the same number, now seemed like the moment to raise it.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -177,6 +203,23 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 
 		// errors from the next round tripper cannot be retried
 		if rtErr != nil {
+			attemptLimit := maxAttempts
+			if cachedMaxRetries != nil {
+				attemptLimit = *cachedMaxRetries
```
rrama, line 242: Nit: Should probably warn if the value has been overridden. Otherwise the log line below suddenly has a larger `attemptLimit` in "(attempt `actualAttempts`/`attemptLimit`)" without any explanation.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -177,6 +203,23 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
+				actualAttempts < attemptLimit &&
+				isRetryableTransportError(rtErr) &&
+				isRetryableRequest(&localRequest, rm.config) {
+				rm.logger.Warn().Err(rtErr).Msgf("Retrying request after transient transport error (attempt %d/%d)", actualAttempts, attemptLimit)
+				if response != nil {
+					drainAndClose(response.Body)
+				}
+				return response, rtErr
+			}
+			if response != nil && transportRetryEnabled {
```
rrama, line 254: Do we really need the leak fix guard behind the FF? I guess it is safer this way.
Can you please leave a comment explaining that this is what this is, as it confused me for a good 5 seconds.
```go
// leak fix gated on the opt-in so flag-off behavior stays byte-identical to today
```

pkg/networking/middleware/retry_middleware.go
```diff
@@ -204,6 +248,31 @@ func (rm RetryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 			}
 		}
 
+		if transportRetryEnabled && isSuccessResponse(response.StatusCode) && isRetryableRequest(&localRequest, rm.config) && isJSONResponse(response) {
```
rrama, line 292: A cheeky code comment explaining that
```go
// check for the streamed response being cut off in the first maxBufferedJSONResponseBytes bytes and if it is, cause a retry
```
would be nice. Or maybe this should be extracted to a function (then it can be easily unit tested).

pkg/networking/middleware/retry_transport_error.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Contiguous-segment matching avoids near-misses: an entry "foo" must not match /v1/foo-something.
+// A blank entry would otherwise match the empty leading segment of every absolute path.
+func isAllowedPath(path string, allowed []string) bool {
+	segments := strings.Split(path, "/")
+	for _, entry := range allowed {
+		if strings.TrimSpace(entry) == "" {
+			continue
+		}
+		if containsContiguous(segments, strings.Split(entry, "/")) {
```
rrama: I think this would be simpler: 
```go
// Check but padded with slashes to ensure matching full sections.
if strings.Contains("/"+path+"/", "/"+strings.Trim(strings.TrimSpace(entry), "/")+"/") {
```

AI says the only difference would be I'd also need a `strings.HasPrefix(entry, "/")` check before and then match from only the start if so, but it then said that may not even be intentional behaviour of yours.

pkg/networking/middleware/retry_middleware_test.go
```diff
@@ -1418,6 +1421,833 @@ func setupRetryMiddleware(
...
+	config := configuration.NewWithOpts()
+	config.Set(configuration.NETWORK_REQUEST_RETRIES_ENABLED, true)
+	config.Set(ConfigurationKeyRequestAttempts, 3)
+	config.Set(ConfigurationKeyRetryAfter, 1)
+
+	sut := NewRetryMiddleware(config, &logger, rt)
+	resp, err := sut.RoundTrip(httptest.NewRequest(http.MethodGet, "/", nil))
+
+	require.Error(t, err)
+	require.NotNil(t, resp, "CLI's handleResponse dereferences resp.Header even on the terminal failure")
```
rrama, line 2026: Maybe we should just fix CLI caller :p
We already empty the body, so what's the harm in making the contract that the full resp is unavailable?

rrama (commented):
Some AI found bits:
1. The behaviour of POSTs hitting 429s has changed. I think it is saying they no longer retry unless on allowed paths. I'd suggest any 429 can be retried, as we know the server hasn't processed the actual request.
2. We may want `/oauth2/token` added to the path allow list by default. I haven't looked into whether this is 100% safe or not (I quickly asked it and then it gave me a long response about it potentially not being), but worth considering.

rrama (approved):

pkg/app/app_test.go
basti-snyk: Converted into a table-driven test (commit c84c6b0).

pkg/app/app_test.go
basti-snyk: Replaced the hardcoded key literal with the exported configuration.NETWORK_REQUEST_RETRY_ALLOWED_PATHS constant (commit c84c6b0).

pkg/app/app_test.go
basti-snyk: Converted into a table-driven test alongside the one above (commit c84c6b0).

pkg/configuration/configuration.go
basti-snyk: Reworded — the blanket "GetStringSlice does not split on commas" claim is now scoped to CONFIG_KEY_ALLOWED_HOSTS specifically, since NETWORK_REQUEST_RETRY_ALLOWED_PATHS now does split via its own DefaultValueFunction. Checked for similar comments elsewhere; none found (commit 2a0afee).

pkg/networking/middleware/retry_middleware.go
basti-snyk: Switched to mime.ParseMediaType for content-type detection (commit 420a7b1).

pkg/networking/middleware/retry_middleware.go
basti-snyk: Added an explanatory comment at the override site (commit 420a7b1).

pkg/networking/middleware/retry_middleware.go
basti-snyk: This is a floor, not an override: getMaxRetryAttempts uses max(), so a configured maxAttempts=5 can never be lowered to 3, only raised. Added a clarifying comment so the next reader isn't confused the same way; existing tests already assert the floor direction (commit 420a7b1).

basti-snyk: Agreed this would be cleaner, but keeping this PR scoped as you suggested — happy to pick this up as a follow-up.

pkg/networking/middleware/retry_middleware.go
basti-snyk: Added a debug log line for when the override actually raises the effective limit above the configured value (commit 420a7b1).

pkg/networking/middleware/retry_middleware.go
basti-snyk: Kept the FF-gating as-is (intentional — flag-off must stay byte-identical) and expanded the comment explaining what the leak fix is and why it's gated (commit 420a7b1).

pkg/networking/middleware/retry_middleware.go
basti-snyk: Extracted into handleSuccessfulJSONResponse() with its own unit tests (commit 420a7b1).

pkg/networking/middleware/retry_transport_error.go
basti-snyk: Simplified to the slash-padded strings.Contains check you proposed; also normalizes leading-slash entries, which was previously an undocumented, untested anchoring quirk — added a test for that (commit b4e6158).

pkg/networking/middleware/retry_middleware_test.go
basti-snyk: Leaving the CLI caller as-is for this PR — that's a downstream contract change I don't want to bundle in here.

PR #693, opened 2026-08-06
feat: file filter considers tracked files before applying ignores

CatalinSnyk:
### Description

TLDR: The file filter now considers the tracked file state before applying .gitignore rules. Meaning, a tracked file will be included even if it's matched by a rule. 

### Checklist

- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI (local tested) :loading: CLI PR

---

> [!NOTE]
> **Medium Risk**
> Changes which files are included in scans when the flag is enabled, with git index reads on the filtering path; behavior is feature-flagged and degrades to legacy matching on git errors.
> 
> **Overview**
> Adds **tracked-file-aware `.gitignore` handling** behind `internal_snyk_gitignore_respect_tracked_files_enabled` (CLI-1411), wired to rollout flag `clientFileFilterGitignore_TrackedFilesRollout`.
> 
> When the flag is on, `.gitignore` globs are tagged with a `#gitignore:` prefix and `GetFilteredFiles` uses a git-index-backed predicate: files **in the index** that match `.gitignore` stay in the scan set (Git-style), while **`.dcignore` and `.snyk` exclusions still drop tracked paths**. Untracked paths matching `.gitignore` are excluded as before.
> 
> The implementation opens the repo via **go-git**, resolves the scan root against the worktree (subdirectory scans and symlinked parents), and falls back to the previous glob-only matcher if the repo cannot be opened.
> 
> Adds broad regression tests for staged/committed tracked files, nested rules, negation, and index edge cases, plus an optional repo benchmark comparing flag on vs off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c703bd66714643aabc30defa4b7493eb5d028922. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/utils/file_filter.go
+++ b/pkg/utils/file_filter.go
@@ -14,6 +14,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/go-git/go-git/v5"
 	"github.com/rs/zerolog"
 	gitignore "github.com/sabhiram/go-gitignore"
 	"golang.org/x/sync/semaphore"
@@ -32,6 +33,8 @@ const (
 var defaultInvalidRules = []string{}
 
 var nextFileFilterMetricScopeID atomic.Uint64
+
+const gitIgnoreGlobPrefix = "#gitignore:"
 
 type FileFilter struct {
 	path            string
@@ -233,8 +236,16 @@ func (fw *FileFilter) GetRules(ruleFiles []string) ([]string, error) {
 func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan string {
 	var filteredFilesCh = make(chan string)
 
-	// create pattern matcher used to match filesToFilter to glob patterns
-	globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
+	var isFileExcluded func(string) bool
+	if fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES) {
+		isFileExcluded = fw.trackedFileExclusionPredicate(globs)
+	} else {
+		globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
+		isFileExcluded = func(filePath string) bool {
+			return globPatternMatcher.MatchesPath(filePath)
+		}
+	}
+
 	go func() {
 		ctx := context.Background()
 		availableThreads := semaphore.NewWeighted(fw.max_threads)
@@ -250,7 +261,7 @@ func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan
 			go func(f string) {
 				defer availableThreads.Release(1)
 				// filesToFilter that do not match the glob pattern are filtered
-				if !globPatternMatcher.MatchesPath(f) {
+				if !isFileExcluded(f) {
 					filteredFilesCh <- f
 				}
 			}(file)
@@ -273,6 +284,7 @@ func (fw *FileFilter) buildGlobs(ignoreFiles []string) ([]string, error) {
 	}
 
 	enableMetacharacterFix := fw.config.GetBool(FF_FILE_FILTER_METACHARACTER_FIX)
+	respectGitIgnoreTrackedFiles := fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES)
 
 	var globs = make([]string, 0)
 	for _, ignoreFile := range ignoreFiles {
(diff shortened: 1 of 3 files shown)


pkg/utils/file_filter.go
```diff
@@ -233,8 +236,16 @@ func (fw *FileFilter) GetRules(ruleFiles []string) ([]string, error) {
 func (fw *FileFilter) GetFilteredFiles(filesCh chan string, globs []string) chan string {
 	var filteredFilesCh = make(chan string)
 
-	// create pattern matcher used to match filesToFilter to glob patterns
-	globPatternMatcher := gitignore.CompileIgnoreLines(globs...)
+	var isFileExcluded func(string) bool
+	if fw.config.GetBool(FF_GITIGNORE_RESPECT_TRACKED_FILES) {
+		isFileExcluded = fw.trackedFileExclusionPredicate(globs)
+	} else {
```
danskmt, line 242: @PeterSchafer @CatalinSnyk the strategy of prepending a character to .gitignore rules creates a question that must be answered here:
-> How do we guarantee the list of **globs** contains the prefixes attached to it? Consumers may call with any kind of list, they can add, remove, modify the way they want before calling **GetFilteredFiles**, so we are not ensuring the behavior will be applied as we want. For example, what if the consumers do not use the "GetRules" method to build the globs, and build their own?

pkg/utils/file_filter.go
danskmt, line 233: @PeterSchafer @CatalinSnyk Please note that prepending a character to .gitignore rules is changing the behavior of **GetRules** function. Consumers may expect the rules as they are, with no modifications. However, **buildGlobs** at line 227 is now modifying such rules. The question is:
-> Do we need to properly communicate this change of behavior of **GetRules**? It will not be honest in the sense of giving the raw rules, but will be giving the modified rules instead. Also, is there any impact if the consumer wants to add their own rules to globs list?

pkg/utils/file_filter.go
danskmt: e.g. **cli-extension-iac** builds its own globs and never calls **GetRules**

pkg/utils/file_filter.go
danskmt: Someone relying on **GetRules** for some other logic will take the gitignore rules as comments, this is a risk for consumers

pkg/utils/file_filter.go
```diff
@@ -287,13 +299,115 @@ func (fw *FileFilter) buildGlobs(ignoreFiles []string) ([]string, error) {
...
+		if err != nil {
+			return allMatcher.MatchesPath(normalizedPath)
+		}
+
+		if trackedFilesToPreserve[filepath.ToSlash(relativePath)] {
+			return otherMatcher.MatchesPath(normalizedPath)
+		}
+
+		return allMatcher.MatchesPath(normalizedPath)
+	}
```
danskmt, line 384: What about:
```go
	trackedPaths := make(map[string]struct{}, len(gitIndex.Entries))
	for _, entry := range gitIndex.Entries {
		trackedPaths[entry.Name] = struct{}{}
	}

	return func(file string) bool {
		normalizedPath := filepath.ToSlash(file)
		relativePath, err := filepath.Rel(repoRoot, file)
		if err != nil {
			return allMatcher.MatchesPath(normalizedPath)
		}

		if _, tracked := trackedPaths[filepath.ToSlash(relativePath)]; tracked {
			if gitignoreMatcher.MatchesPath(normalizedPath) {
				return otherMatcher.MatchesPath(normalizedPath)
			}
		}

		return allMatcher.MatchesPath(normalizedPath)
	}
```

CatalinSnyk:
```go
if _, tracked := trackedPaths[filepath.ToSlash(relativePath)]; tracked {
    if gitignoreMatcher.MatchesPath(normalizedPath) {
        return otherMatcher.MatchesPath(normalizedPath)
    }
}
```
This more/less does the same thing as #685 (almost, since there we used a slice of sorted paths & binary search). But this will store all the tracked files, and not the intersection of tracked & .gitignored files in the map. On bigger repos this might take up more memory.

danskmt: I ran a benchmarking with claude and it seems it improves speed by quite a lot:
image

image

CatalinSnyk: i would say this should & will be investigated further in the follow-up ticket. The benchmark numbers do look great, but not sure how accurate they are given the entries kept column is yielding quite different results 😄

pkg/utils/file_filter.go
basti-snyk: That's ok, we do not need to cater to people NOT using a function that was exported.

pkg/utils/file_filter.go
basti-snyk: This is theoretical. Currently, external users are osflows, secrets and snyk-ls. They all want this change.

danskmt (approved):

PR #697, opened 2026-08-12
feat: [IANDT-237] add active contributor emitter

jonnyowenpowell:
### Description

This PR adds an API client for the public submit active contributors endpoint (`internal/apiclients/contributors_ingest`), a helper function to collect contributor data (`contributors.go`) from the Git repository the command is being run in, and an emitter object (`emitter.go`) which ties the two together, exposing an `Emit` function to perform the work.

### Checklist

- [x] Linted (`make lint`)
- [ ] Test your changes work for the CLI (currently dead code, will be done for later PRs in the stack)


internal/contributors/collect.go
```diff
@@ -0,0 +1,90 @@
...
+		if email == "" || when.Before(since) || when.After(now) {
+			continue
+		}
+
+		key := strings.ToLower(email)
+		if seen, ok := latest[key]; ok && !when.After(seen.CommitDate) {
+			continue
+		}
+		latest[key] = contributors_ingest.Contributor{Email: email, CommitDate: when}
+	}
```
jonnyowenpowell: Subject to change based on decisions:
- do we prefer a max commits limit & partial data or no limit and slower CLI calls for big repos?
- do we want to use commits from all branches or just main?

Also needs to be reworked to align on using exactly one of commit date and authored date, not this current mixture.

pkg/apiclients/contributors_ingest/ingest.go
```diff
@@ -0,0 +1,174 @@
...
+	)
+
+	return client
+}
+
+// CreateContributingDevs records contributors against one entity in one org.
+//
+// The call is bounded by requestTimeout, retries included. A shorter deadline on
+// ctx still wins, so callers can tighten the bound but not loosen it.
+func (c *Client) CreateContributingDevs(
```
NikkiL-Snyk: nit: creating contributing devs to me sounds like getting the list of contributors, rather than the post to entitlements service

jonnyowenpowell: Yeah good point, lets call it something like SubmitContributingDevelopers (or is it active contributors?).

internal/contributorbilling/contributors.go
```diff
@@ -0,0 +1,130 @@
...
+	head, err := repo.Head()
+	if errors.Is(err, plumbing.ErrReferenceNotFound) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("read head: %w", err)
+	}
+
+	since := now.AddDate(0, 0, -contributorWindowDays)
+	iter, err := repo.Log(&git.LogOptions{From: head.Hash(), Order: git.LogOrderCommitterTime})
```
jonnyowenpowell: The `From: head.Hash()` should Go if we decide to scan _all_ commits instead of just those from the current branch.

jonnyowenpowell:
Remember to assess bringing back capability check

internal/contributorbilling/contributors.go
jonnyowenpowell: Update, we want all branches, change this: https://snyk.slack.com/archives/C0BD83YU3V0/p1786621865477609?thread_ts=1786381946.266679&cid=C0BD83YU3V0

jonnyowenpowell: Done!

jonnyowenpowell:
> Remember to assess bringing back capability check

Done, this doesn't belong here - and will instead be checked in the code that may call `Emit`.

internal/apiclients/contributors_ingest/client.go
```diff
@@ -0,0 +1,174 @@
+// Package contributors_ingest is a client for the Contributing Devs ingest API,
+// which records the git contributors associated with a Snyk entity for billing.
+package contributors_ingest
```
jonnyowenpowell, line 3: Move this package to internal

brettgurman-snyk (approved):

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+		if !name.IsBranch() && !name.IsRemote() {
+			return nil
+		}
+
+		tip, commitErr := object.GetCommit(repo.Storer, ref.Hash())
+		if commitErr != nil {
+			return nil //nolint:nilerr // ingore refs that don't resolve to a commit
+		}
+
+		return object.NewCommitIterCTime(tip, seen, nil).ForEach(func(commit *object.Commit) error {
```
basti-snyk: **🔴 Critical — fails on shallow clones, which is the default in CI.**

`object.NewCommitIterCTime(...).Next()` calls `object.GetCommit` on **every parent hash before handing the commit to the callback**, and returns the error if the object is absent. It never consults `repo.Storer.Shallow()`. On a shallow clone the boundary commit's parents are missing, so the walk dies with `plumbing.ErrObjectNotFound` before a single contributor is recorded, and `Emit` returns `collect contributors: read commits: object not found`.

Reproduced against this repo — see the PR-level comment for the depth table. `--depth=1/5/20` all return 0 contributors and an error; `--depth=50/1000` succeed only because the shallow boundary lands outside the 90-day window, so the committer-date `ErrStop` fires before the walk reaches the missing parents.

I don't think this is fairly characterised as an upstream go-git limitation: this function deliberately returns `nil, nil` for a non-repository *and* for a repository with no commits, so hard-failing on the commonest CI clone shape is an inconsistency in its own error contract.

**Fix:** check `repo.Storer.Shallow()` up front — either treat a shallow repo like the other "can't measure" cases (return `nil, nil` and log), or tolerate `plumbing.ErrObjectNotFound` inside the walk and mark the result partial. Worth a shallow-clone test either way.

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+
+	seen := make(map[plumbing.Hash]bool)
+	latest := make(map[string]contributors_ingest.Contributor)
+	err = refs.ForEach(func(ref *plumbing.Reference) error {
+		name := ref.Name()
+		if !name.IsBranch() && !name.IsRemote() {
+			return nil
+		}
+
+		tip, commitErr := object.GetCommit(repo.Storer, ref.Hash())
```
basti-snyk: **Suggestion** — this swallows every `GetCommit` failure on a branch or remote ref.

Tags are already excluded by the `IsBranch()`/`IsRemote()` guard above, so a failure reaching this line means real object corruption or the shallow-clone case. Silently dropping it loses billable contributors with no signal. Worth logging at debug/warn rather than discarding.

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	contributors := slices.Collect(maps.Values(latest))
+	slices.SortFunc(contributors, func(a, b contributors_ingest.Contributor) int {
+		return strings.Compare(a.Email, b.Email)
+	})
+	return contributors, nil
+}
+
+// openRepositoryFast opens the git repository at path the same way
```
basti-snyk, line 169: **Suggestion** — ~50 lines of hand-rolled go-git storage plumbing whose performance win is never measured.

The only benchmark (`contributors_bench_test.go`) is env-gated behind `CONTRIBUTOR_BENCH_REPO` and never runs in CI, so there's no before/after number backing the `KeepDescriptors` claim in the doc comment. It also couples to go-git internals (`filesystem.Storage`, `cache.NewObjectLRUDefault`) and silently falls back on three separate paths, so a regression here would be invisible.

Either land a measured number in the comment, or drop it for plain `git.PlainOpenWithOptions` and add it back when it's shown to matter.

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+	switch {
+	case errors.Is(err, git.ErrIsBareRepository):
+		// no worktree
+	case err != nil:
+		return repo, noopClose, nil //nolint:nilerr // worktree is invalid, assume no worktree
+	default:
+		wtFs = osfs.New(wt.Filesystem.Root(), osfs.WithBoundOS())
+	}
+
+	fs := osfs.New(gitDir, osfs.WithBoundOS())
```
basti-snyk: **Suggestion** — `osfs.WithBoundOS()` quietly relaxes symlink containment versus the path it replaces.

Verified in go-billy v5.9.0 (`osfs/os_bound.go` vs `osfs/os_chroot.go`): BoundOS does **not** rewrite symlink targets, so a symlink inside the tree can point outside `baseDir` and will be followed. The default `ChrootOS` that `git.PlainOpenWithOptions` uses *does* rewrite them.

Since `openRepositoryFast` exists specifically to swap out `PlainOpenWithOptions`, it's inheriting a different containment posture as a side effect. If that's a considered trade, a comment saying so would help; if not, it's worth a second look.

internal/contributors/emitter.go
```diff
@@ -0,0 +1,107 @@
...
+// Emit reports the repository's current contributors against item.
+//
+// It returns nil when there is no git repository or no commits in the contributor
+// window.
+func (e *Emitter) Emit(ctx context.Context, repoPath string, orgID uuid.UUID, item Item) error {
+	if orgID == uuid.Nil {
+		return errors.New("org ID is required")
+	}
+	if err := item.validate(); err != nil {
+		return err
```
basti-snyk, line 85: **Should Fix** — `Emit` takes a `ctx` but drops it before the expensive work.

`collectContributors(repoPath, e.now())` walks every ref of a possibly-huge monorepo with no cancellation path; `ctx` is first honoured only downstream in `SubmitContributors`. A caller's deadline can't interrupt the git walk, which is the slow part.

This matters more given `WithRetryInterval`'s stated motivation is teardown-time work on a tight budget — the network side is bounded but the collection side isn't.

**Fix:** thread `ctx` into `collectContributors` and check `ctx.Err()` in the `ForEach` callback.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -106,6 +109,15 @@ func WithErrorHandler(handler networktypes.ErrorHandlerFunc) RetryMiddlewareOpti
 	}
 }
 
+// WithRetryInterval sets the initial backoff between attempts for this middleware
```
basti-snyk, line 112: **Should Fix** — the doc comment over-promises.

It sells "callers on a tighter time budget… such as work done during teardown", but the override only replaces `backoffMethod.InitialInterval`, which is consulted **only** when `shouldRetry` returns `errRetryNecessary`. When the server sends `Retry-After` or `X-RateLimit-Reset`, `shouldRetry` returns a `*backoff.RetryAfterError` and the server's delay wins unchanged, capped only by `maxRetryAfter = 10m`.

So the option does not actually bound the time budget on the path most likely to blow it. The contributors client survives this purely because its own 10s `context.WithTimeout` is the real ceiling — not because of this option.

**Fix:** either say so explicitly ("server-suggested `Retry-After` delays are not affected by this override"), or add a max-total-backoff option that genuinely expresses the constraint.

pkg/networking/middleware/retry_middleware.go
```diff
@@ -106,6 +109,15 @@ func WithErrorHandler(handler networktypes.ErrorHandlerFunc) RetryMiddlewareOpti
...
+// instance, overriding the configured value. Use it for callers on a tighter time
+// budget than a user-facing API request, such as work done during teardown.
+func WithRetryInterval(interval time.Duration) RetryMiddlewareOption {
```
basti-snyk, line 116: **Should Fix** — new permanent public API in `pkg/` for one internal, currently-uncalled caller.

CLAUDE.md is explicit that `pkg/` symbols are a forever contract and go in only when a consumer outside this repo needs them *today*; the PR description itself says this is dead code for now.

One clarification, since the obvious counter-proposal doesn't work: the internal caller **cannot** just set the existing key, because `configurationKeyRetryAfter` is unexported — I checked, there's no exported `ConfigurationKeyRetryAfter`. So the real options are:

1. Export the existing config key (smaller, additive change), or
2. Keep the option, justified on the one thing config genuinely can't express — the key is int-seconds, so sub-second intervals are impossible.

Worth noting that production uses exactly `1s`, which the existing key covers fine; only the tests need 50ms. That's a thin justification for permanent public surface. Your call, but it'd be good to have it be a deliberate one.

internal/apiclients/contributors_ingest/client.go
```diff
@@ -0,0 +1,166 @@
...
+
+	"github.com/google/uuid"
+	"github.com/rs/zerolog"
+
+	v20241015 "github.com/snyk/go-application-framework/internal/apiclients/contributors_ingest/2024-10-15"
+	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/networking/middleware"
+)
+
+const (
```
basti-snyk, line 20: **Should Fix** — request policy is hard-coded rather than driven through `Configuration`.

`requestTimeout`, `requestAttempts`, `retryInterval` and `apiVersion` are unexported constants, and `retrying()` builds its own `configuration.NewWithOpts()` (line 158) instead of taking the caller's. CLAUDE.md says behaviour that may differ by consumer belongs in `Configuration` so CLI/IDE/MCP hosts can wire it differently — an IDE may want a longer timeout, an MCP host may want fail-fast.

Counterweight, in fairness: this is `internal/`, single-consumer, and YAGNI cuts the other way. I'd just like this to be a decision rather than a default — a one-line comment saying "policy intentionally owned here until a second consumer exists" would settle it.

internal/apiclients/contributors_ingest/client.go
```diff
@@ -0,0 +1,166 @@
...
+		if base.Transport != nil {
+			transport = base.Transport
+		}
+	}
+	if logger == nil {
+		nop := zerolog.Nop()
+		logger = &nop
+	}
+
+	// The retry middleware reads only these two keys, so the policy is owned here
```
basti-snyk: **Suggestion** — comment says "these two keys" but only one is set.

Only `ConfigurationKeyRequestAttempts` is set on the next line; the retry-after key is left unset and falls through to `defaultRetryAfterSeconds = 5` — which is then overridden anyway by `WithRetryInterval(retryInterval)` on line 162. Worth rewording so the next reader doesn't go looking for the second `Set`.

internal/apiclients/contributors_ingest/client.go
```diff
@@ -0,0 +1,166 @@
...
+	resp, err := c.api.CreateContributingDevsWithApplicationVndAPIPlusJSONBodyWithResponse(
+		ctx,
+		orgID,
+		&v20241015.CreateContributingDevsParams{Version: apiVersion},
+		requestBody(entityType, entityID, contributors),
+	)
+	if err != nil {
+		return fmt.Errorf("post contributors: %w", err)
+	}
+	if resp.StatusCode() != http.StatusCreated {
```
basti-snyk, line 77: **Suggestion** — only `201` is accepted.

This matches the spec as vendored, so it's correct today. But a proxy returning `200`, or an upstream that later adds `200`/`202` as a success shape, turns into `unexpected status` and a silently lost billing record. Given the payload is billing data, it may be worth accepting the 2xx family, or at minimum leaving a comment tying the strictness back to the spec so a future reader knows it's deliberate.

internal/apiclients/contributors_ingest/2024-10-15/spec.yaml
```diff
@@ -0,0 +1,113 @@
+openapi: 3.0.3
+info:
+  title: Contributing Devs Ingest API
+  description: |
+    Hidden ingest API for CLI Active Contributor billing.
+    Vendored from contributingDevsIngestApi (2024-10-15); sweater-comb refs inlined for codegen.
```
basti-snyk, line 6: **Should Fix** — hand-inlined vendored spec with no drift detection.

"sweater-comb refs inlined for codegen" means this is a manually-resolved snapshot of upstream `contributingDevsIngestApi`, and nothing in the repo notices when the two diverge. If upstream adds a required field, `make generate` regenerates happily from the stale local copy and requests start getting 400s with no signal until someone tests by hand.

**Fix:** at minimum record where the authoritative spec lives (in `gen.go` or a README next to it) so the next person knows what to diff against; better, add a CI check that compares the vendored copy to upstream.

basti-snyk (commented):
## Automated verification review — verdict: **FAIL** (1 Critical)

Ran a two-pass review (semantic + adversarial) over all 18 files in three chunks. Build/lint/test are **tool-confirmed PASS** — I adopted CI's result for `51d128f` rather than re-running, since every relevant CircleCI/Snyk check is green on this exact SHA.

Summary: **1 Critical, 5 Should Fix, 11 Suggestions.** The Critical is reproducible and is the only one I'd consider blocking. Inline comments below.

### 🔴 The one that matters

`collectContributors` fails outright on a shallow clone — the default clone shape in CI (`actions/checkout` defaults to `fetch-depth: 1`). I reproduced it against this repo:

| clone | oldest commit | result |
|---|---|---|
| `--depth=1` | 2026-08-17 | `read commits: object not found`, 0 contributors |
| `--depth=5` | 2026-08-11 | `read commits: object not found`, 0 contributors |
| `--depth=20` | 2026-07-24 | `read commits: object not found`, 0 contributors |
| `--depth=50` | 2026-04-29 | ✅ 17 contributors |
| `--depth=1000` | 2022-05-25 | ✅ 17 contributors |

It only works when the shallow boundary happens to fall *outside* the 90-day window, so the committer-date `storer.ErrStop` fires before the walk reaches the missing parents. Details in the inline comment on `collect.go`.

### 🔵 Suggestions not worth their own thread

- `contributorWindowDays = 90` is hard-coded; the billing window is exactly the kind of thing a host may need to vary.
- No `maxItems` on `contributors` in the spec and no batching in the client — a large monorepo posts one unbounded body against a 10s deadline.
- Raw author emails leave the machine with no config gate or opt-out anywhere in the path. Presumably intentional for billing, but worth being an explicit, stated decision rather than an implicit one.
- No `t.Parallel()` in any of the 37 new tests.
- No coverage for context cancellation, 429 + `Retry-After`, worktrees, submodules, or detached HEAD.
- `collect.go:79`: `// ingore` → `// ignore`.
- `collect.go:127`: the `case err != nil:` branch is dead — `Worktree()` returns only `ErrIsBareRepository` or nil.

### Checked and dismissed

So you don't have to re-litigate these — I verified each against the actual library source and they do **not** hold:

- *"Retry backoff is linear (1s, 1s, 1s)"* — no; `backoff/v5` gives 1.5× growth with ±50% jitter: `[0.5s,1.5s]`, `[0.75s,2.25s]`.
- *"`retrying()` mutates the caller's `http.Client`"* — no; it copies into a fresh client before touching `Transport`, and `TestNewClient_LeavesTheCallersClientUntouched` covers it.
- *"The wall-clock timing tests are flaky"* — I did the arithmetic against the real jitter ranges; the tightest assertion needs ~490ms of scheduling overhead to invert. They're fine.

CODEOWNERS and the `go-billy` indirect→direct promotion both check out.

<sub>🤖 Generated by an automated verification pass (Claude Code). Findings were filtered against verified upstream library source; the Critical was reproduced locally.</sub>

internal/apiclients/contributors_ingest/client.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if logger == nil {
+		nop := zerolog.Nop()
+		logger = &nop
+	}
+
+	// The retry middleware reads only these two keys, so the policy is owned here
+	// rather than by whatever configuration the caller happens to hold.
+	config := configuration.NewWithOpts()
+	config.Set(middleware.ConfigurationKeyRequestAttempts, requestAttempts)
```
basti-snyk, line 159: https://github.com/snyk/go-application-framework/pull/686 may be interesting, as billing could opt-in to additional retry-functionality once it's merged

internal/apiclients/contributors_ingest/retry_test.go
basti-snyk, line 1: If you just use the middleware - why not assume that it works or move the tests to the middleware package if they are missing there?

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+// commit date >= authored date, which means a cutoff for commit date contains all
+// our potential commits with an authored date in the window. Additionally, commits
+// with authored dates in the future are excluded, because these are obviously
+// incorrect and so including them risks over-counting.
+//
+// Authors are case-insensitively deduplicated and reported in lowercase.Results
+// are sorted by email so payloads are stable.
+//
+// A path that is not a git repository, or a repository with no commits, yields no
+// contributors and no error, because we proceed with a contributor count of 0.
```
basti-snyk: Isn't all this deductible from the code? I don't believe in documenting what happens in non-exported functions, duplicating the code. If there is something that cannot be deduced from code, e.g. the reason, why a certain solution was selected, a comment may be reasonable. But I'd err on the side of leaving these comment blocks out. Not a blocker, but AI tends to over-comment.

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("read commits: %w", err)
+	}
+
+	contributors := slices.Collect(maps.Values(latest))
+	slices.SortFunc(contributors, func(a, b contributors_ingest.Contributor) int {
+		return strings.Compare(a.Email, b.Email)
+	})
+	return contributors, nil
```
basti-snyk, line 108: This should probably be cached for sub-sequent calls. GAF already has a caching mechanism - see https://github.com/snyk/go-application-framework/blob/9b50cddc663a77214fea5e41756dc528f3feb92a/pkg/configuration/configuration.go#L98

basti-snyk: This would become more relevant with IDE usage.

internal/contributors/collect.go
```diff
@@ -0,0 +1,148 @@
...
+
+	fs := osfs.New(gitDir, osfs.WithBoundOS())
+	st := filesystem.NewStorageWithOptions(fs, cache.NewObjectLRUDefault(), filesystem.Options{KeepDescriptors: true})
+
+	fast, err := git.Open(st, wtFs)
+	if err != nil {
+		_ = st.Close()
+		return repo, noopClose, nil //nolint:nilerr // use the slow repo if our fast open fails
+	}
+	return fast, st.Close, nil
```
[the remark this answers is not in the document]
jonnyowenpowell: FP: This doesn't understand how `PlainOpenWithOptions` works - there's no close to call.

internal/contributors/collect.go
jonnyowenpowell: Will look into this! 👍

internal/contributors/collect.go
jonnyowenpowell: This debug log is incredibly unlikely ever to be read, because we don't know what contributors we're expecting. Even if it was read, we need local access to a customers git repo to reproduce. I think the best approach here is to make sure we have good test coverage, and monitor contributor data for inconsistencies, not write unactionable logs.

internal/contributors/collect.go
jonnyowenpowell: I absolutely did measure this, it took the time down from over 4s for some larger repositories to sub 200ms. Benchmarking this in CI doesn't make a lot of sense, because it's hard to get access to a good set of repositories to test there. This is a tool engineers can use to check their changes if they change this file.

internal/contributors/collect.go
jonnyowenpowell: There's no risk here! If a maliciously constructed symlink exists in a customers git repository, at worst we go real a file which isn't a git object, fail to read it and move on (without aborting the whole count).

internal/contributors/emitter.go
jonnyowenpowell: Fair, I'll include this!

pkg/networking/middleware/retry_middleware.go
jonnyowenpowell: I'll add the explicit comment about server-suggested retry-after; there's no point disobeying it when we're talking to _our own_ APIs!

pkg/networking/middleware/retry_middleware.go
jonnyowenpowell: We need this option, the code will not be dead for long - this is a strange complaint.

internal/apiclients/contributors_ingest/client.go
jonnyowenpowell: I'm not going to add this kind of commentary, because I don't think it's valuable to litter a codebase with "we aren't doing X because it doesn't make sense". The timeouts being scoped to this API makes sense - it's not something that blocks users of the CLI from using the product if it does timeout, there's no reason they'd ever want to configure it.

internal/apiclients/contributors_ingest/client.go
jonnyowenpowell: I will reword the comment.

internal/apiclients/contributors_ingest/client.go
jonnyowenpowell: This is noise - the API returns a 201, and if proxy changes it (?) then we already submitted the data so it really doesn't matter; there's no "lost billing record".

internal/apiclients/contributors_ingest/2024-10-15/spec.yaml
jonnyowenpowell: I'll put a note in gen.go 👍

internal/apiclients/contributors_ingest/retry_test.go
jonnyowenpowell: Can do, these check we actually hook up the retries correctly, but give me the 👍 and I'll remove these tests.

internal/contributors/collect.go
jonnyowenpowell: AI didn't write this, I did. You _could_ read this reasonably complicated code to understand that it includes all branches, but it's considerably easier to be sure if it's stated explicitly. You can't really deduce from the code that the commit date vs. authored date distinction has been considered, rather than accidentally blurred.

If you like, I can remove the final two paragraphs, although they do explain _why_ we sort emails and _why_ we don't return an error when certain things fail.

internal/contributors/collect.go
jonnyowenpowell: It's non-trivial to invent the cache key - we're reading from all branches, so you need to capture the entire state of the repo, e.g. the current HEAD and directory wouldn't suffice. You could hash the whole `.git` directory or something, but that has huge performance concerns.

Given that we're not expecting this to run in the IDE, it's unclear this is a problem space that we need to dive into.

internal/apiclients/contributors_ingest/client.go
jonnyowenpowell: What's the timeline for it merging, what does it give us, and how easy is it for us to opt-in?

basti-snyk: Today :)

internal/contributors/collect.go
```diff
@@ -0,0 +1,147 @@
...
+		if commitErr != nil {
+			return nil //nolint:nilerr // ingore refs that don't resolve to a commit
+		}
+
+		return object.NewCommitIterCTime(tip, seen, nil).ForEach(func(commit *object.Commit) error {
+			seen[commit.Hash] = true
+
+			if commit.Committer.When.Before(since) {
+				return storer.ErrStop
+			}
```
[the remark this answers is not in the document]
basti-snyk: @jonnyowenpowell have you seen this?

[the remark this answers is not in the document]
jonnyowenpowell: Yes, this comes back repeatedly in bot reviews. The assumption is included in the comment above the function. You can't commit before you author, so this only happens when the author date is wrong, and we have no ability to stop users of the CLI from completely forging all author dates, so instead we assume they haven't, because the alternative is to literally scan every single commit in the repo, which is performance prohibitive.

basti-snyk (approved):

PR #700, opened 2026-08-12
fix: improve FileFilter metrics

octavian-snyk:
### Description

This PR improves FileFilter analytics by replacing per-run metric scopes with stable keys that aggregate repeated runs. Runs are grouped by the feature-flag combination they use, allowing input/output file counts, rule counts, and durations, to be compared across variants. Temporary legacy variants: var0 (no feature flags), var1 (metaCharFix), var2 (gitIgnoreRespectTrackedFiles).

This PR also introduces a concurrency-safe metrics accumulator.

Tests cover every feature-flag variant, repeated and concurrent runs, nil recorders, aggregation behavior, and FileFilter metrics reaching invocation analytics.

### Checklist

- [x] Linted (`make lint`)

---

> [!NOTE]
> **Medium Risk**
> Analytics key schema and aggregation semantics change for FileFilter metrics consumed via invocation analytics; filtering behavior is unchanged when metrics are disabled.
> 
> **Overview**
> **FileFilter analytics** no longer use per-run scope IDs (`file-filter.<scopeID>.<metric>`). Metrics are written under stable keys grouped by feature-flag combination (`var0` / `var1` / `var2`, or `file-filter.<metric>` when both flags are on), and repeated runs **aggregate** into the same keys instead of overwriting each other.
> 
> A new **`metrics.Accumulator`** backs this: it sums integer metrics (input/output counts, durations), keeps per-key maximums, and records bools without aggregation, with mutex-protected global state shared across accumulators on different recorders. **`ResetAccumulated`** clears that state for tests. **`FileFilter`** wires `WithMetrics` through an accumulator and skips flag reads when nothing is recording.
> 
> Metric names and semantics shift: e.g. `survivingFileCount` → `filter.outputFileCount`, `rulesBuildDurationMs` → `rules.durationMs`, plus `filter.inputFileCount`; duration fields are **summed** across calls for a variant. Feature flags are recorded as `feature.metaCharFix` and `feature.includeTracked`. **`RecorderFake`** gains **`NewRecorderFake()`** so maps are always initialized.
> 
> Tests cover aggregation, concurrency, variants, and invocation analytics extension keys; workflow tests expect the new `file-filter.var0.*` shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b8d12259f3ebdb1bb020b9e7a0b52baa17c5a89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/metrics/accumulator.go
+++ b/internal/metrics/accumulator.go
@@ -2,28 +2,32 @@ package metrics
 
 import "sync"
 
-// Accumulator aggregates the metrics of repeated operations into one value per key, so that callers
-// report under a fixed set of keys instead of namespacing every operation with a run identifier.
-// A key must use the same aggregation throughout.
-//
-// A Recorder keeps only the last value per key, so every update writes the running aggregate
-// through and there is nothing to flush. Two Accumulators sharing a key on one Recorder overwrite
-// each other rather than aggregating, so operations reporting under one key must share one.
-//
-// An Accumulator is safe for concurrent use.
+// accumulated is shared by every Accumulator, so two of them reporting under one key add up
+// instead of overwriting each other in a Recorder, which keeps only the last value per key.
+var (
+	accumulatedMu sync.Mutex
+	accumulated   = map[string]int{}
+)
+
+// Accumulator aggregates repeated operations into one running value per key, so callers report
+// under a fixed set of keys instead of namespacing every operation. Safe for concurrent use.
 type Accumulator struct {
-	mu       sync.Mutex
 	recorder Recorder
-	ints     map[string]int
 }
 
-// NewAccumulator returns an Accumulator writing through to recorder. A nil recorder discards
-// everything; use Recording to skip obtaining values that are expensive to compute.
+// NewAccumulator returns an Accumulator writing through to recorder; a nil recorder discards
+// everything. It continues the shared aggregate rather than starting one of its own.
 func NewAccumulator(recorder Recorder) *Accumulator {
-	return &Accumulator{
-		recorder: recorder,
-		ints:     make(map[string]int),
-	}
+	return &Accumulator{recorder: recorder}
+}
+
+// ResetAccumulated discards every accumulated value. It exists for tests, which assert the values a
+// single Recorder saw and would otherwise observe the aggregate of preceding tests.
+func ResetAccumulated() {
+	accumulatedMu.Lock()
+	defer accumulatedMu.Unlock()
+
+	accumulated = map[string]int{}
 }
 
 // Recording reports whether values written to this Accumulator reach a Recorder.
@@ -37,11 +41,11 @@ func (a *Accumulator) AddToSum(key string, delta int) {
 		return
 	}
 
-	a.mu.Lock()
-	defer a.mu.Unlock()
+	accumulatedMu.Lock()
+	defer accumulatedMu.Unlock()
 
-	a.ints[key] += delta
-	a.recorder.AddExtensionIntegerValue(key, a.ints[key])
+	accumulated[key] += delta
+	a.recorder.AddExtensionIntegerValue(key, accumulated[key])
 }
 
 // RecordBool records value under key, overwriting whatever was recorded for it before. Unlike
@@ -62,13 +66,13 @@ func (a *Accumulator) KeepMaximum(key string, value int) {
 		return
 	}
 
-	a.mu.Lock()
-	defer a.mu.Unlock()
+	accumulatedMu.Lock()
+	defer accumulatedMu.Unlock()
 
-	if current, recorded := a.ints[key]; recorded && current >= value {
+	if current, recorded := accumulated[key]; recorded && current >= value {
 		return
 	}
 
-	a.ints[key] = value
+	accumulated[key] = value
 	a.recorder.AddExtensionIntegerValue(key, value)
 }
--- a/internal/metrics/accumulator.go
+++ b/internal/metrics/accumulator.go
@@ -1,0 +1,74 @@
+package metrics
+
+import "sync"
+
+// Accumulator aggregates the metrics of repeated operations into one value per key, so that callers
+// report under a fixed set of keys instead of namespacing every operation with a run identifier.
+// A key must use the same aggregation throughout.
+//
+// A Recorder keeps only the last value per key, so every update writes the running aggregate
+// through and there is nothing to flush. Two Accumulators sharing a key on one Recorder overwrite
+// each other rather than aggregating, so operations reporting under one key must share one.
+//
+// An Accumulator is safe for concurrent use.
+type Accumulator struct {
+	mu       sync.Mutex
+	recorder Recorder
+	ints     map[string]int
+}
+
+// NewAccumulator returns an Accumulator writing through to recorder. A nil recorder discards
+// everything; use Recording to skip obtaining values that are expensive to compute.
+func NewAccumulator(recorder Recorder) *Accumulator {
+	return &Accumulator{
+		recorder: recorder,
+		ints:     make(map[string]int),
+	}
+}
+
+// Recording reports whether values written to this Accumulator reach a Recorder.
+func (a *Accumulator) Recording() bool {
+	return a != nil && a.recorder != nil
+}
+
+// AddToSum adds delta to the running sum recorded under key.
+func (a *Accumulator) AddToSum(key string, delta int) {
+	if !a.Recording() {
+		return
+	}
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	a.ints[key] += delta
+	a.recorder.AddExtensionIntegerValue(key, a.ints[key])
+}
+
+// RecordBool records value under key, overwriting whatever was recorded for it before. Unlike
+// AddToSum or KeepMaximum, a bool is not aggregated: callers reporting a fact that does not change
+// between calls (e.g. whether a feature flag applied to the run) simply see it written through.
+func (a *Accumulator) RecordBool(key string, value bool) {
+	if !a.Recording() {
+		return
+	}
+
+	a.recorder.AddExtensionBoolValue(key, value)
+}
+
+// KeepMaximum records value under key unless a larger value was recorded for it before. It keeps an
+// outlier visible where a sum would hide it, e.g. the slowest of many operations.
+func (a *Accumulator) KeepMaximum(key string, value int) {
+	if !a.Recording() {
+		return
+	}
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	if current, recorded := a.ints[key]; recorded && current >= value {
+		return
+	}
+
+	a.ints[key] = value
+	a.recorder.AddExtensionIntegerValue(key, value)
+}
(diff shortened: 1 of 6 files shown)


pkg/utils/file_filter.go
```diff
@@ -122,35 +134,54 @@ func WithConfig(config configuration.Configuration) FileFilterOption {
...
 	}
 }
 
-func (fw *FileFilter) recordMetricLazy(scopeID, key string, getMetric func() int) {
-	if fw.metricsRecorder != nil {
-		fw.metricsRecorder.AddExtensionIntegerValue(metricKey(scopeID, key), getMetric())
+// The record*Lazy helpers obtain a metric only once it is known to be recorded, so that a FileFilter
+// without a recorder pays for none of them.
+func (fw *FileFilter) recordSumLazy(variant, key string, getMetric func() int) {
+	if fw.metrics.Recording() {
```
danskmt: shall we have the name isRecording instead? is better when reading the if

pkg/utils/file_filter.go
```diff
@@ -241,22 +264,23 @@ func (fw *FileFilter) GetRules(ruleFiles []string) ([]string, error) {
 		return nil, err
 	}
 
-	return append(fw.defaultRules, globs...), nil
+	rules := append(fw.defaultRules, globs...)
+	fw.recordSumLazy(variant, metricFileFilterRuleCount, func() int { return len(rules) })
```
danskmt: I'd suggest moving the inner function to a separate variable, it is better to read

danskmt (approved):
Approved, left some comments.
Please remove some of the comments as they are not really needed

internal/metrics/accumulator.go
```diff
@@ -0,0 +1,71 @@
+package metrics
+
+import "sync"
+
+// accumulated is shared by all Accumulators.
+var (
+	accumulatedMu sync.Mutex
+	accumulated   = map[string]int{}
```
PeterSchafer, line 8: Suggestion: Ignoring for now that this is called Accumulator, You could actually track all the entries using `map[string][]int` and calculate different things from it, sum, mean, median ...

pkg/utils/file_filter.go
```diff
@@ -74,14 +72,25 @@ type DotSnykRule struct {
...
+// File-filter metric keys have the shape "file-filter.<variant>.<metric>".
 const (
 	metricFileFilterPrefix = "file-filter" // prefix for all file-filter analytics keys
 
-	metricFileFilterDurationMs           = "durationMs"           // elapsed time for GetFilteredFiles: exclusion-predicate build (incl. git index read) and filtering, including caller drain of the result channel
-	metricFileFilterSurvivingFileCount   = "survivingFileCount"   // number of files that passed exclusion in this filter run
-	metricFileFilterRulesBuildDurationMs = "rulesBuildDurationMs" // elapsed time for GetRules: directory walk, ignore discovery, and buildGlobs
-	metricFileFilterMetacharacterFix     = "metacharacterFix"     // whether FF_FILE_FILTER_METACHARACTER_FIX was enabled for this run
-	metricFileFilterRespectTrackedFiles  = "respectTrackedFiles"  // whether FF_GITIGNORE_RESPECT_TRACKED_FILES was enabled for this run
+	metricFileFilterVariantLegacy       = "var0" // neither feature flag enabled
```
PeterSchafer: Question: what do we do when the feature flags are being removed and the feature will just be enabled?

PeterSchafer: Should `var0` just be something like `default` file filter configuration?

octavian-snyk: I'm not sure we have the character allowance for a long descriptive name. I went with var0, var1, etc., in the first place because names like "legacy" or "metaFix" were too long.
What if we drop the var3 prefix entirely instead? Then var0, var1, and var2 can act as our temporary ones.

`file-filter.filter.inputFileCount` instead of `file-filter.var3.filter.inputFileCount`
`file-filter.var0.filter.inputFileCount` would remain `file-filter.var0.filter.inputFileCount`

PeterSchafer: sounds good, `default` is no `var`

internal/metrics/accumulator.go
PeterSchafer: This is more an optional suggestion. or design feedback

pkg/utils/file_filter.go
```diff
@@ -74,14 +72,27 @@ type DotSnykRule struct {
...
+	metricFileFilterVariantMetacharFix  = "var1" // FF_FILE_FILTER_METACHARACTER_FIX only
+	metricFileFilterVariantTrackedFiles = "var2" // FF_GITIGNORE_RESPECT_TRACKED_FILES only
+	metricFileFilterVariantBothFixes    = ""     // both feature flags enabled; the variant segment is omitted
+
+	metricFileFilterInputFileCount = "filter.inputFileCount" // files offered to GetFilteredFiles, before exclusion
+	metricFileFilterRuleCount      = "filter.ruleCount"      // glob patterns GetRules produced
+
+	// Record feature flags alongside the variant so consumers need not decode its name.
+	metricFileFilterFeatureMetacharFix  = "feature.metaCharFix"    // whether FF_FILE_FILTER_METACHARACTER_FIX applied to the run
+	metricFileFilterFeatureTrackedFiles = "feature.includeTracked" // whether FF_GITIGNORE_RESPECT_TRACKED_FILES applied to the run
```
danskmt: they all have same prefix name **metricFileFilter**, but their values are different. I would suggest improving this naming. Since we are inside file_filter.go, we could use **metricVariantXYZ** for the var ones, **metricFilterXYZ** for the 'filter.*' ones and **metricFeatureXYZ** for the 'feature.*' ones, WDYT?

pkg/utils/file_filter.go
```diff
@@ -74,14 +72,27 @@ type DotSnykRule struct {
...
+
+	metricFileFilterDurationMs           = "filter.durationMs"      // elapsed time for GetFilteredFiles, including the caller's drain of the result channel
+	metricFileFilterRulesBuildDurationMs = "rules.durationMs"       // elapsed time for GetRules: directory walk, ignore discovery, and buildGlobs
+	metricFileFilterSurvivingFileCount   = "filter.outputFileCount" // number of files that passed exclusion
```
danskmt: same here, and organize them better so the similar ones stays together

pkg/utils/file_filter.go
```diff
@@ -74,14 +72,27 @@ type DotSnykRule struct {
...
 
-	metricFileFilterDurationMs           = "durationMs"           // elapsed time for GetFilteredFiles: exclusion-predicate build (incl. git index read) and filtering, including caller drain of the result channel
-	metricFileFilterSurvivingFileCount   = "survivingFileCount"   // number of files that passed exclusion in this filter run
-	metricFileFilterRulesBuildDurationMs = "rulesBuildDurationMs" // elapsed time for GetRules: directory walk, ignore discovery, and buildGlobs
-	metricFileFilterMetacharacterFix     = "metacharacterFix"     // whether FF_FILE_FILTER_METACHARACTER_FIX was enabled for this run
-	metricFileFilterRespectTrackedFiles  = "respectTrackedFiles"  // whether FF_GITIGNORE_RESPECT_TRACKED_FILES was enabled for this run
+	metricFileFilterVariantLegacy       = "var0" // neither feature flag enabled
+	metricFileFilterVariantMetacharFix  = "var1" // FF_FILE_FILTER_METACHARACTER_FIX only
+	metricFileFilterVariantTrackedFiles = "var2" // FF_GITIGNORE_RESPECT_TRACKED_FILES only
+	metricFileFilterVariantBothFixes    = ""     // both feature flags enabled; the variant segment is omitted
```
danskmt: I see these metrics are actually calculating time?
```go
defer fw.recordSumLazy(variant, metricFileFilterRulesBuildDurationMs, func() int {
		return int(time.Since(start).Milliseconds())
	})
```

pkg/utils/file_filter.go
```diff
@@ -122,35 +133,53 @@ func WithConfig(config configuration.Configuration) FileFilterOption {
...
+	case metacharacterFix:
+		variant = metriVariantMetacharFix
+	case respectTrackedFiles:
+		variant = metricVariantTrackedFiles
+	default:
+		variant = metricVariantLegacy
+	}
+
+	fw.metrics.RecordBool(metricKey(variant, metricFeatureMetacharFix), metacharacterFix)
+	fw.metrics.RecordBool(metricKey(variant, metricFeatureTrackedFiles), respectTrackedFiles)
```
danskmt: I would suggest moving the inner call to a separate variable with meaningful name, e.g.:
```go
metricFeatureMetacharFixFQN := metricKey(variant, metricFeatureMetacharFix)
metricFeatureTrackedFilesFQN := metricKey(variant, metricFeatureTrackedFiles)
fw.metrics.RecordBool(metricFeatureMetacharFixFQN, metacharacterFix)
fw.metrics.RecordBool(metricFeatureTrackedFilesFQN, respectTrackedFiles)
```

pkg/utils/file_filter.go
```diff
@@ -122,35 +133,53 @@ func WithConfig(config configuration.Configuration) FileFilterOption {
...
 }
 
-func newFileFilterMetricScopeID() string {
-	return fmt.Sprintf("%d", nextFileFilterMetricScopeID.Add(1))
+func metricKey(variant, key string) string {
+	if variant == metricVariantBothFixes {
+		return fmt.Sprintf("%s.%s", metricPrefix, key)
+	}
+	return fmt.Sprintf("%s.%s.%s", metricPrefix, variant, key)
 }
```
danskmt: Instead of **metricKey**, what about **buildMetricKey** or **buildMetricFQN** or **buildMetricKeyFQN**?

pkg/utils/file_filter.go
```diff
@@ -207,21 +237,12 @@ func (fw *FileFilter) GetAllFiles() chan string {
...
-	// them alongside the build duration rather than relying on a GetFilteredFiles run that may
-	// never happen or may read a different flag value.
-	scopeID := newFileFilterMetricScopeID()
+	// Rule building and filtering may resolve feature flags at different times.
+	variant := fw.metricVariant()
 	start := time.Now()
-	defer fw.recordMetricLazy(scopeID, metricFileFilterRulesBuildDurationMs, func() int {
+	defer fw.recordSumLazy(variant, metricRulesBuildDurationMs, func() int {
 		return int(time.Since(start).Milliseconds())
 	})
```
danskmt, line 249: I see **recordSumLazy** calls the `metricKey` function inside, and thus no need to concat before calling the function. This pattern is not repeated in `fw.metrics.RecordBool` that requires otherwise.
Can we have some unified contract between these functions?

danskmt (approved):
LGTM - Added some suggestions

pkg/utils/file_filter.go
danskmt: Ignore this comment, the problem was in reading how this function was being used together with the "variant" part. Comments below are more clear about it.

PR #703, opened 2026-08-13
feat: contributor capture middleware

NikkiL-Snyk:
### Description

This PR implements contributor capturing middleware which captures project IDs and passes them to a sink provided in the constructor. A future PR will implement and export a post invoke hook to collect data from the sync and call the contributor emitter, which will own the sink, and code will be written in the CLI to check the feature flag, construct the middleware and post invoke hook, register the middleware via the new hook, and register the post invoke hook.

### Checklist

- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI - done using `feat/contributor-capture-wiring` (in draft, enables capture and wires the middleware output to the contributor emitter), and have run against the real commands we're intending to capture data for

---

> [!NOTE]
> **Medium Risk**
> Middleware sits on the HTTP round-trip path and parses API bodies for entity IDs when registered, but host gating matches auth targets and failures are logged/recovered without altering successful responses.
> 
> **Overview**
> Adds **`NetworkAccess.AddMiddleware`** so callers can wrap the HTTP round tripper outside built-in retry/auth/error middleware (last registered runs outermost; clones preserve the chain).
> 
> Introduces **`contributor_capture`** middleware that, for configured Snyk API hosts only, matches specific product endpoints (monitor, IaC share, hidden test create/components, AI BOM upload, deeproxy report), peeks bounded request/response bodies without breaking downstream reads, and reports **project** and **revision** IDs to an injected **`Sink`**. The hidden test flow tracks `publish_report` / `report` on create-test and only captures project IDs from components polling when that test was marked pending.
> 
> Also re-exports contributor **`EntityType`** constants from `internal/contributors` and extends networking mocks/tests for middleware registration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 785ee0d935f1c6fd8bca2fada94870dcc79a0fa8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	m.clearPendingTest(testID)
+	m.recordProjects(entityIDs, state.interactionID)
+}
+
+func (m *ContributorCaptureMiddleware) recordProjects(ids []string, interactionID string) {
+	sink := m.sinkProvider()
+	for _, entityID := range ids {
+		sink.RecordEntity(contributors.EntityTypeProject, entityID, interactionID)
+	}
```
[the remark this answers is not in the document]
jonnyowenpowell: This is intentional, the project is used a reference to find the target, and since all the projects share the same target, we can pick any of them.

pkg/networking/middleware/contributor_capture/parse.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// parseCreateTestPublishReport reports whether a CreateTest request body asked for publish_report.
+func parseCreateTestPublishReport(body []byte) bool {
+	var doc createTestRequestDoc
+	if err := json.Unmarshal(body, &doc); err != nil {
+		return false
+	}
+
+	return doc.Data.Attributes.Configuration.Output.Report
```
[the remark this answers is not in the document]
jonnyowenpowell: I have tested this. As of today, the requests sent by the commands we are capturing use the hidden endpoint, and this JSON path.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,224 @@
...
+	testID := testIDFromPath(path)
+	if testID == "" || !m.isPendingTest(testID) {
+		return
+	}
+	entityIDs := extract(body)
+	if len(entityIDs) == 0 {
+		return
+	}
+	m.clearPendingTest(testID)
+	m.recordProjects(entityIDs, state.interactionID)
```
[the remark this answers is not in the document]
jonnyowenpowell: There will be no later "component polls" because it's not a polling API - the call happens exactly once after test status polling finishes.

jonnyowenpowell:
In response to review bot:

> [Broken Seeker Contract](https://github.com/snyk/go-application-framework/pull/703/changes#diff-7cab3e7523c467b751b514521b10318e9fa37e3acc6cdad57f1f2fc1fc4b711eR54-R59)

Not true, it never broke it, it just made it potentially take a less optimized path, and it appears the bot is looking at stale data, there's now extra work to restore seeking, with a test.

> [Public API Layering Violation](https://github.com/snyk/go-application-framework/pull/703/changes#diff-b6404eb4b8d476736fa60f1d4a0a001c5d86e25cabbbe73fde0cdbea42094fcfR3-R8)

It's fine that the middleware isn't usable outside this module. We _can_ move it to internal, but it separates it from all the other middleware. Regardless, small change, happy to do it if a maintainer agrees with 🤖.

> [Memory Leak in Singletons](https://github.com/snyk/go-application-framework/pull/703/changes#diff-33c2d166834d67e0069c832a0f4aa161925fa7dd643fa2ad94f0eaa225bf1300R31-R31)

Memory leak is a dramatic word for this, even if it is technically accurate. You'd need thousands of tests before you got to 1MB of memory - it's a non-issue, and implementing a TTL is overengineering.

 > [Hot-path Performance Issue](https://github.com/snyk/go-application-framework/pull/703/changes#diff-33c2d166834d67e0069c832a0f4aa161925fa7dd643fa2ad94f0eaa225bf1300R101-R101)

This happens once for every HTTP request and adds ns, which will be statistically insignificant compared to request durations. Again, this is rather dramatic.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,224 @@
...
+// captureIfPendingTest records entities from a GET components response, but
+// only if its test ID was marked pending by an earlier create-test call.
+func (m *ContributorCaptureMiddleware) captureIfPendingTest(state captureState, path string, body []byte, extract func([]byte) []string) {
+	testID := testIDFromPath(path)
+	if testID == "" || !m.isPendingTest(testID) {
+		return
+	}
+	entityIDs := extract(body)
+	if len(entityIDs) == 0 {
+		return
```
basti-snyk: [Should Fix] `pendingTestsSingleton.ids` is only cleared on the success path. `captureIfPendingTest` returns early at line 212 when `extract(body)` finds no matching entity IDs, so `clearPendingTest` (line 215) is never reached for a create-test flow that never gets a matching components response. Since this is process-global state (`var pendingTestsSingleton`, line 31) with no eviction on that path, entries accumulate for the life of the process. It's harmless for a short-lived CLI run, but this repo's AGENTS.md explicitly calls out avoiding package-level mutable state for hosts that reuse a process across many invocations (IDE/MCP), so it's worth closing before this middleware is wired up more broadly.

Fix: clear the pending-test entry on any outcome for a tracked test ID (not just a successful capture), or replace the singleton with a bounded/expiring cache.

— AI review

pkg/networking/middleware/contributor_capture/parse.go
```diff
@@ -0,0 +1,157 @@
...
+func parseCreateTestID(body []byte) string {
+	var doc createTestResponseDoc
+	if err := json.Unmarshal(body, &doc); err != nil {
+		return ""
+	}
+	return parseUUID(doc.Data.ID)
+}
+
+// parseMonitorProjectIDs extracts project IDs from a monitor/monitor-dependencies response.
+func parseMonitorProjectIDs(body []byte) []string {
```
basti-snyk: [Suggestion] None of the parse functions in this file (`parseMonitorProjectIDs`, `parseIaCShareProjectIDs`, `parseComponentsProjectIDs`, etc.) log anything when parsing fails — they just return an empty result. If one of these API response shapes changes, capture stops silently with no signal anywhere that something broke.

Probably better as a follow-up than expanding this PR: a debug-level log with the endpoint kind on parse failure would make a schema drift visible.

— AI review

pkg/networking/middleware/contributor_capture/sink.go
```diff
@@ -0,0 +1,12 @@
+package contributor_capture
+
+import "github.com/snyk/go-application-framework/internal/contributors"
+
+// Sink is the mechanism by which this middleware reports captured entities.
+type Sink interface {
+	RecordEntity(entityType contributors.EntityType, entityID, interactionID string)
```
basti-snyk: [Suggestion] `RecordEntity` is called synchronously from `RoundTrip`, so a slow `Sink` implementation would add latency to every captured request. The current implementation (`internal/contributors/sink.go`) is just a mutex-protected in-memory map write, so this is harmless today, but the interface itself doesn't say implementations must be fast/non-blocking.

Probably better as a follow-up than expanding this PR: a doc comment on the interface stating the non-blocking expectation would prevent a future implementation from violating it silently.

— AI review

pkg/networking/middleware/contributor_capture/parse.go
jonnyowenpowell: This mechanism is essentially invisible to users of the CLI, so I don't think a debug log is the right call - we won't see it anyway. What I plan to do (in the final PR for this work) is introduce data for analytics to use which indicates whether a capture occurred, because that is reported and can be analyzed.

pkg/networking/middleware/contributor_capture/sink.go
jonnyowenpowell: Added the comment.

pkg/networking/middleware/contributor_capture/parse.go
basti-snyk: Could also be an error log - this is the only way to analyze things in the CLI, so error logging is essential. In every single support case, we require the debug level logs for this reason.

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: Clearing it on any outcome is not correct, because we don't know when we're observing the final call related to that test, so we may clear it early and miss the one containing the project ID.

As I describe above, I think adding a TTL is over-engineering and it's not possible to replace the singleton because there is no workflow scope to attach to in the middleware.

pkg/networking/middleware/contributor_capture/contributor_capture.go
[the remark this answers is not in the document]
basti-snyk: I'd clean it up. Right now, IDEs may be out of scope, but they would run into exactly this kind of error, where a long running, re-used process needs to be tidy.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	if m.sinkProvider == nil || m.sinkProvider() == nil {
+		return m.next.RoundTrip(req)
+	}
+
+	state, needsResponse := m.beginRequestCapture(req)
+
+	if !needsResponse {
+		return m.next.RoundTrip(req)
```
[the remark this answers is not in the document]
jonnyowenpowell: This is intentional, the data is only used later when corresponding entities exist in the platform which happens on success. If a response failure occurs, we don't know if a test started or not.

pkg/networking/middleware/contributor_capture/contributor_capture.go
basti-snyk: TTL wouldn't be good, I agree. How about attaching to the context?

pkg/networking/middleware/contributor_capture/parse.go
jonnyowenpowell: The point is that this is invisible to users of the CLI, and so it's very unlikely for them to ever bring us the logs. Nevertheless, I can include this in a follow up which wires this up and adds the analytics.

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: It's possible, but only by making changes to set this up in `engineImpl` - and relies on code outside of this codebase not changing how it handles contexts, which isn't great.

jonnyowenpowell: Also, the changes actually require attaching it to a context that hangs around for the lifetime of the process anyway, so it doesn't solve test IDs sticking around after they're not required.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,309 @@
...
+	default:
+	}
+
+	return state, true
+}
+
+// classifyRequest decides whether req is one contributor-capture cares
+// about and, if so, what kind it is.
+func (m *ContributorCaptureMiddleware) classifyRequest(req *http.Request) (EndpointKind, bool) {
+	if req.Header.Get(interactionIDHeader) == "" {
```
basti-snyk: That's a rather harsh early return - really wanted?

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,309 @@
...
+	sinkProvider SinkProvider
+	logger       *zerolog.Logger
+}
+
+type pendingTests struct {
+	mu  sync.Mutex
+	ids map[string]struct{}
+}
+
+var pendingTestsSingleton = &pendingTests{ids: make(map[string]struct{})}
```
basti-snyk: I think using dependency injection and a pointer instead of a global variable would be better. The pending tests struct could be created in the middleware constructor and stored in the CaontributorCaptureMiddleware.

pkg/networking/middleware/contributor_capture/contributor_capture.go
NikkiL-Snyk: We use the interaction id from the header, so if it's not there, better to return early? Why would you not want to stop early here?

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	return bodyBytes
+}
+
+// completeRequestCapture extracts and records entities from a matched
+// request's response, using the state beginRequestCapture gathered.
+func (m *ContributorCaptureMiddleware) completeRequestCapture(state captureState, req *http.Request, res *http.Response) {
+	defer m.recover(req)
+
+	if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusMultipleChoices {
```
basti-snyk: Why not >= 400? For example, 301, 307 etc should IMHO be fine. Only the 4xx, 5xx seem erroneous to me.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,309 @@
...
+
+// beginRequestCapture performs whatever capture the request alone allows and
+// reports whether the response still has to be processed to finish the job.
+func (m *ContributorCaptureMiddleware) beginRequestCapture(req *http.Request) (state captureState, needsResponse bool) {
+	kind, matched := m.classifyRequest(req)
+	if !matched {
+		return captureState{}, false
+	}
+
+	defer m.recover(req)
```
basti-snyk: [Should Fix] `beginRequestCapture()` calls `m.classifyRequest(req)` before registering `defer m.recover(req)`, so a panic during classification isn't caught. `completeRequestCapture()` right below correctly puts its `defer m.recover(req)` first — this one doesn't.

Fix: move the defer to the top of the function, before the classification call.

```go
func (m *ContributorCaptureMiddleware) beginRequestCapture(req *http.Request) (state captureState, needsResponse bool) {
	defer m.recover(req)

	kind, matched := m.classifyRequest(req)
	if !matched {
		return captureState{}, false
	}
	...
```

— AI review

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,309 @@
...
+			if m.logger != nil {
+				m.logger.Debug().
+					Err(decodeErr).
+					Str("path", req.URL.Path).
+					Str("content_encoding", encoding).
+					Msg("contributor capture: failed to decode response body, using raw bytes")
+			}
+		} else {
+			parseBytes = decoded
+		}
```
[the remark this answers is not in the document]
basti-snyk: Seems to be a false positive to me, but please double-check

internal/contributors/sink.go
```diff
@@ -0,0 +1,70 @@
...
+const EntityTypeRevision = contributors_ingest.EntityTypeRevision
+
+var captureEnabled atomic.Bool
+
+// Enable turns on contributor capture for this process.
+func Enable() {
+	captureEnabled.Store(true)
+}
+
+var singleton = &contributorSink{recordsByInteractionID: make(map[string]entityRecord)}
```
basti-snyk: I see this pattern with global (singleton) variables as problematic, especially when we want to run tests in parallel. Instead, dependency injection should be used.

internal/contributors/sink.go
```diff
@@ -0,0 +1,70 @@
...
+// EntityType identifies the kind of Snyk entity a captured ID refers to.
+type EntityType = contributors_ingest.EntityType
+
+// EntityTypeProject identifies a captured ID as a Snyk project ID.
+const EntityTypeProject = contributors_ingest.EntityTypeProject
+
+// EntityTypeRevision identifies a captured ID as a Snyk revision ID.
+const EntityTypeRevision = contributors_ingest.EntityTypeRevision
+
+var captureEnabled atomic.Bool
```
basti-snyk: Global variable :(

internal/contributors/sink.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Enable turns on contributor capture for this process.
+func Enable() {
```
basti-snyk: If it's turned on by one thread, it's gonna be turned on for all threads. Is there a scenario, where we don't want to capture in some threads?

internal/contributors/sink_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetSink(t *testing.T) {
+	assert.Nil(t, GetSink(), "capture must be disabled until Enable is called")
+
+	Enable()
```
basti-snyk: this prevents concurrency of tests (e.g. integration tests)

pkg/networking/middleware/contributor_capture/body_test.go
basti-snyk, line 1: Please consolidate the tests into table tests. There's a lot of duplication here.

pkg/networking/middleware/contributor_capture/classify.go
```diff
@@ -0,0 +1,99 @@
...
+
+var (
+	monitorPathPattern    = regexp.MustCompile(`(?i)^/v1/monitor(?:/|$)`)
+	monitorDepsPattern    = regexp.MustCompile(`(?i)^/v1/monitor-dependencies(?:/|$)`)
+	iacSharePattern       = regexp.MustCompile(`(?i)^/v1/iac-cli-share-results(?:/|$)`)
+	createTestPathPattern = regexp.MustCompile(`(?i)^/hidden/orgs/[0-9a-fA-F-]{36}/tests$`)
+	componentsPathPattern = regexp.MustCompile(`(?i)^/hidden/orgs/[0-9a-fA-F-]{36}/tests/([0-9a-fA-F-]{36})/components$`)
+	aiBomUploadPattern    = regexp.MustCompile(`(?i)^/rest/orgs/[0-9a-fA-F-]{36}/ai_boms/upload$`)
+	deeproxyReportPattern = regexp.MustCompile(`(?i)^/report/([0-9a-fA-F-]{36})/?$`)
+)
```
basti-snyk: This is significantly slower that `strings.Index` and `strings.HasPrefix`. It also adds gc pressure. I don't think we need to use regexp here. Something like this would work, too, and be much more efficient:
```
func isMonitorPath(p string) bool {
    p = strings.ToLower(p) // or better: compare case-insensitively without allocating
    return p == "/v1/monitor" || strings.HasPrefix(p, "/v1/monitor/")
}
```

pkg/networking/middleware/contributor_capture/classify.go
```diff
@@ -0,0 +1,99 @@
...
+
+func firstSubmatchUUID(pattern *regexp.Regexp, path string) string {
+	matches := pattern.FindStringSubmatch(path)
+	if len(matches) < 2 {
+		return ""
+	}
+	return parseUUID(matches[1])
+}
+
+func parseUUID(value string) string {
```
basti-snyk, line 177: a little bit longer, but claude gives me this as an example:
```
const uuidLen = 36

// isUUIDShape checks the 8-4-4-4-12 hex-with-hyphens pattern, case-insensitive,
// without allocating or invoking the regexp engine.
func isUUIDShape(s string) bool {
	if len(s) != uuidLen {
		return false
	}
	for i, c := range s {
		switch i {
		case 8, 13, 18, 23:
			if c != '-' {
				return false
			}
		default:
			if !isHexDigit(byte(c)) {
				return false
			}
		}
	}
	return true
}

func isHexDigit(c byte) bool {
	return (c >= '0' && c <= '9') ||
		(c >= 'a' && c <= 'f') ||
		(c >= 'A' && c <= 'F')
}

// matchCreateTestPath replaces createTestPathPattern.MatchString(path).
// Matches: /hidden/orgs/{uuid}/tests
func matchCreateTestPath(path string) bool {
	const prefix = "/hidden/orgs/"
	const suffix = "/tests"

	if !strings.HasPrefix(path, prefix) {
		return false
	}
	rest := path[len(prefix):]

	if !strings.HasSuffix(rest, suffix) {
		return false
	}
	id := rest[:len(rest)-len(suffix)]

	return isUUIDShape(id)
}
```

pkg/networking/middleware/contributor_capture/contributor_capture_test.go
```diff
@@ -0,0 +1,902 @@
...
+	require.NoError(t, res.Body.Close())
+
+	records := sink.Records()
+	require.Len(t, records, 1)
+	assert.Equal(t, contributors.EntityTypeProject, records[0].EntityType)
+	assert.Equal(t, projectID, records[0].EntityID)
+	assert.Equal(t, "urn:snyk:interaction:monitor-1", records[0].InteractionID)
+}
+
+func TestContributorCaptureMiddleware_capturesIaCShareProjectIDs(t *testing.T) {
```
basti-snyk: almost the same as the previous test - please use table tests to reduce duplications. Please apply that strategy to all tests, so we don't duplicate test setup etc everywhere.

pkg/networking/middleware/contributor_capture/parse.go
```diff
@@ -0,0 +1,264 @@
...
+import (
+	"encoding/json"
+	"regexp"
+	"strings"
+)
+
+var (
+	projectIDFromURI               = regexp.MustCompile(`/project/([0-9a-fA-F-]{36})(?:/|$)`)
+	deeproxyReportCompletePattern  = regexp.MustCompile(`(?i)"status"\s*:\s*"COMPLETE"`)
+	deeproxyUploadProjectIDPattern = regexp.MustCompile(`"uploadResult"\s*:\s*\{[^}]*"projectId"\s*:\s*"([0-9a-fA-F-]{36})"`)
```
basti-snyk: same here with the regexp. as this is running within the network stack, let's use better performing options.

internal/contributors/sink.go
```diff
@@ -0,0 +1,70 @@
...
+
+// RecordEntity records the first entity seen for a given interaction ID, and
+// does nothing on subsequent calls for the same interaction ID.
+func (s *contributorSink) RecordEntity(entityType EntityType, entityID, interactionID string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if _, exists := s.recordsByInteractionID[interactionID]; exists {
+		return
+	}
+	s.recordsByInteractionID[interactionID] = entityRecord{entityType: entityType, entityID: entityID}
```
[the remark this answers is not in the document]
NikkiL-Snyk: This is known and by design

pkg/networking/middleware/contributor_capture/classify.go
NikkiL-Snyk: What's wrong with using the uuid.Parse function here? I think that's safer than making our own checks?

Agree conceptually on not using regex, and will look into it

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	// Oversized body: return truncated prefix for parsing.
+	req.Body = stitch(buf, req.Body)
+	return buf[:maxBytes], nil
+}
+
+// readResponseBody peeks up to maxBytes of res.Body for parsing, restoring
+// res.Body so it can still be read normally afterwards regardless of outcome.
+func readResponseBody(res *http.Response, maxBytes int64) ([]byte, error) {
```
basti-snyk: [Should Fix] `readResponseBody` and `peekAndRestoreBody` (line 105) have no production caller — the middleware only ever calls `readResponseBodyForParse`, which reimplements the same peek/restore logic with a different oversize contract (a truncated prefix + `fullyRead=false` instead of `errBodyTooLarge`). The only reference to `readResponseBody` is the `export_test.go` alias used by `body_test.go`'s `TestReadResponseBody_*` tests, so those tests exercise a code path the middleware never runs, and a future fix applied to one of these twins could silently not apply to the other.

Fix: delete `readResponseBody`/`peekAndRestoreBody` and their tests, or have `readResponseBodyForParse` call through to a single shared implementation instead of duplicating it.

— AI review

pkg/networking/middleware/contributor_capture/classify.go
```diff
@@ -0,0 +1,204 @@
...
+
+// isValidUUID checks if a string is a valid UUID without allocating.
+func isValidUUID(s string) bool {
+	_, err := uuid.Parse(s)
+	return err == nil
+}
+
+// EndpointKind identifies which known product-API request/response shape a request matches.
+type EndpointKind int
+
```
basti-snyk, line 25: [Should Fix] `EndpointKind` and its 7 constants are exported public API with zero consumers outside this package — `classifyEndpoint`, the only function that produces an `EndpointKind`, is deliberately kept unexported and re-exposed to tests only via the `export_test.go` alias (`ClassifyEndpoint = classifyEndpoint`). This enum doesn't get the same treatment even though nothing outside the package references it. Per this repo's own stability guidance (AGENTS.md): a type goes in the public surface only when an external consumer genuinely needs it today — moving from unexported to exported later is easy, the reverse is a breaking change.

Fix: lowercase `EndpointKind` and its constants (`endpointKind`, `endpointNone`, ...) and, if tests need them, add an `export_test.go` alias the same way `classifyEndpoint` already has one.

— AI review

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	switch strings.ToLower(strings.TrimSpace(contentEncoding)) {
+	case "gzip", "x-gzip":
+		reader, err := gzip.NewReader(bytes.NewReader(bodyBytes))
+		if err != nil {
+			return nil, err
+		}
+		defer reader.Close()
+		decoded, err := io.ReadAll(reader)
```
basti-snyk: [Should Fix] `decodeCaptureBody`'s gzip branch has no bound on decompressed output size. The *compressed* input is capped at `maxCaptureBodyBytes` (64 KiB) by the caller, but gzip routinely achieves ratios well over 100x on repetitive data, so a 64 KiB compressed body can expand to tens or hundreds of MB in memory before `io.ReadAll` returns. This runs on real network response bodies for any matched endpoint whose response sets `Content-Encoding: gzip`, in a middleware meant to run in long-lived host processes (IDE/MCP) — a decompression-bomb-style vector if an upstream response is ever hostile or misconfigured.

Fix: wrap the gzip reader in an `io.LimitReader` bounded to a fixed decompressed-size ceiling, and treat hitting that limit as a decode failure.

— AI review

pkg/networking/middleware/contributor_capture/parse.go
```diff
@@ -0,0 +1,265 @@
...
+
+	return parseUUID(uuidCandidate)
+}
+
+// extractDeeproxyProjectIDFromTruncated extracts project ID from truncated/malformed JSON.
+func extractDeeproxyProjectIDFromTruncated(body []byte) string {
+	s := string(body)
+
+	// Check for COMPLETE status
+	if !strings.Contains(strings.ToUpper(s), `"COMPLETE"`) {
```
basti-snyk, line 220: [Should Fix] `extractDeeproxyProjectIDFromTruncated`'s status check is a plain, unanchored substring search (`strings.Contains(strings.ToUpper(s), "COMPLETE")`) over the whole truncated body, not tied to the actual `status` field. This path only runs once the structured `json.Unmarshal` has already failed — i.e. on input that's already known to be malformed or truncated — so a `COMPLETE` substring appearing anywhere else in the payload (a nested message, a filename, unrelated JSON) would let this fall through to the `projectId`/`project_id` scan below and attribute a nearby UUID as the project ID, even though the real status might not be COMPLETE.

Fix: anchor the check to the actual `status` field shape (e.g. search for `"status":"COMPLETE"` allowing for whitespace) instead of a bare substring match, so a malformed/truncated body fails closed rather than risking a misattributed project ID.

— AI review

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,313 @@
...
+	kind, matched := classifyEndpoint(req.Method, req.URL.Path)
+	if !matched {
+		return EndpointNone, false
+	}
+
+	if kind == EndpointDeeproxyReport {
+		return m.classifyDeeproxyRequest(req)
+	}
+
+	apiURL, err := url.Parse(m.config.GetString(configuration.API_URL))
```
basti-snyk: [Should Fix] `classifyRequest` matches monitor/IaC-share/test-create/AI-BOM-upload requests with a plain hostname-equality check against `configuration.API_URL` (`req.URL.Hostname() != apiURL.Hostname()`), while `classifyDeeproxyRequest` a few lines below uses `parentmiddleware.ShouldRequireAuthentication`, which additionally canonicalizes hosts and honors `AUTHENTICATION_SUBDOMAINS`/`AUTHENTICATION_ADDITIONAL_URLS`. A CLI/host configured with either of those (or hitting a canonicalized host like `app.snyk.io` instead of `api.snyk.io`) would get contributor capture on the deeproxy route but silently miss it on every other matched route.

Fix: use `ShouldRequireAuthentication` for the host check in both places, or document why the deeproxy route intentionally needs a different (broader) host match than the rest.

— AI review

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,313 @@
...
+	case EndpointTestComponents:
+		return m.projectIDsFromComponentsResponse(path, parseBytes)
+	case EndpointDeeproxyReport:
+		return []string{parseDeeproxyReportProjectID(parseBytes)}
+	default:
+		return nil
+	}
+}
+
+func (m *ContributorCaptureMiddleware) projectIDsFromComponentsResponse(path string, parseBytes []byte) []string {
```
basti-snyk, line 219: [Suggestion] `projectIDsFromComponentsResponse` calls `isPendingTest` and `clearPendingTest` as two separate lock acquisitions rather than one atomic check-and-clear. If two requests for the same `testID` are ever in flight concurrently (a retried poll racing the original, or two goroutines polling the same test), both can observe `isPendingTest == true` before either calls `clearPendingTest`, so both proceed to parse a project ID and both call `m.record(...)` — a double capture for one test if those two calls carry different interaction IDs. This is a different failure mode from the already-discussed "cleared before capture completes" ordering issue on this file — it's specifically about two concurrent callers both passing the check.

Fix: collapse the check-and-clear into one method that takes the lock once (e.g. `takeIfPending(testID) bool` that returns whether this call gets to consume the pending flag), so it can only be consumed once.

— AI review

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -0,0 +1,140 @@
...
+	"strings"
+)
+
+// maxCaptureBodyBytes bounds bodies read for contributor capture parsing.
+const maxCaptureBodyBytes = 64 << 10 // 64 KiB
+
+// errBodyTooLarge is returned when a body exceeds maxCaptureBodyBytes.
+var errBodyTooLarge = errors.New("contributor capture: body exceeds capture limit")
+
+// readRequestBodyForParse peeks a request body for capture parsing. When the
```
basti-snyk: [Suggestion] The doc comment on `readRequestBodyForParse` is missing a clause and doesn't parse as a sentence: "peeks a request body for capture parsing. When the / on oversized bodies instead of an error, for callers that can parse from a prefix." Looks like a copy-edit from `readResponseBodyForParse`'s comment lost a line.

Fix: `// readRequestBodyForParse peeks a request body for capture parsing. When the body exceeds maxBytes, it returns a truncated prefix instead of an error, for callers that can parse from a prefix.`

— AI review

pkg/networking/middleware/contributor_capture/classify.go
basti-snyk: in general `uuid.Parse` is better, but you'll need to separate the UUID from the path. As mentioned above, regexp are not a good solution as they are very slow and not a good fit for middleware usage.

pkg/networking/middleware/contributor_capture/classify.go
jonnyowenpowell: I will unexport them.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -0,0 +1,321 @@
...
+	config configuration.Configuration,
+	sinkProvider SinkProvider,
+	logger *zerolog.Logger,
+) *ContributorCaptureMiddleware {
+	return &ContributorCaptureMiddleware{
+		next:         roundTripper,
+		config:       config,
+		sinkProvider: sinkProvider,
+		logger:       logger,
+		pendingTests: &pendingTests{ids: make(map[string]struct{})},
```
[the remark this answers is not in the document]
jonnyowenpowell: This is resolved by the new approach of having the middleware constructed once in the CLI and the same instance of `network_types.Middleware` used across all calls.

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: Deeproxy has a separate (deprecated, replaced by the Test API calls we also capture here) ingress, at a different domain. I'll include a comment.

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: Double capture is not a major concern, but this reduces total lines of code and is more correct, so I've implemented it.

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: Moved the defer!

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: I have actually removed this - the capture can safely be attempted on all requests (it makes no modifications and leaks no data), and is not a performance issue because the request URLs are checked. This CLI is not making lots of performance sensitive calls to non-snyk hosts which exactly match the paths checked here.

pkg/networking/middleware/contributor_capture/contributor_capture.go
jonnyowenpowell: This change has been made.

pkg/networking/middleware/contributor_capture/body_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	cc "github.com/snyk/go-application-framework/pkg/networking/middleware/contributor_capture"
+)
+
+func TestPeekResponseBody_restoresBodyOnReadError(t *testing.T) {
```
basti-snyk, line 17: suggestion: lots of these tests look like they could be merged into a table test. please check.

pkg/networking/middleware/contributor_capture/classify.go
jonnyowenpowell: As discussed offline, I've benchmarked the entire `classifyEndpoint` function on my machine, and found the following:
```
goos: darwin
goarch: arm64
pkg: github.com/snyk/go-application-framework/pkg/networking/middleware/contributor_capture
cpu: Apple M5 Pro
BenchmarkClassifyEndpoint/match/monitor-18              95274682                12.24 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/match/monitor-deps-18         50678262                23.39 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/match/iac-share-18            64915891                18.58 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/match/create-test-18          30326642                39.42 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/match/components-18           18150352                66.76 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/match/aibom-upload-18         26974323                44.10 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/match/deeproxy-report-18      40317838                30.05 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/miss/unknown-method-18        157617445                7.601 ns/op           0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/miss/wrong-method-for-path-18                 98251933                12.01 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/miss/unknown-path-18                          87855769                13.42 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/miss/invalid-uuid-18                          56102292                21.15 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/miss/uuid-parse-then-reject-18                25434303                47.36 ns/op            0 B/op          0 allocs/op
BenchmarkClassifyEndpoint/miss/query-string-trimmed-18                  91863391                12.86 ns/op            0 B/op          0 allocs/op
```

Granted, this is a fairly fast machine, but with the common case (unknown-path) being 13.4ns, then even executing 100x slower, that's still completely negligable at once per HTTP call. Say the word, and I'll commit the benchmark.

jonnyowenpowell: Discussed offline, committing the benchmark.

pkg/networking/middleware/contributor_capture/body_test.go
jonnyowenpowell: Merged the two that can 👍

pkg/networking/middleware/contributor_capture/sink.go
```diff
@@ -0,0 +1,10 @@
+package contributor_capture
+
+import "github.com/snyk/go-application-framework/internal/contributors"
+
+// Sink is the mechanism by which this middleware reports captured entities. It
+// is call synchronously from the middleware, and so is required to be fast and
+// non-blocking.
+type Sink interface {
+	RecordEntity(entityType contributors.EntityType, entityID string)
```
basti-snyk, line 11: **[consumer-advocate]:** `RecordEntity(entityType contributors.EntityType, entityID string)` takes `contributors.EntityType` from `internal/contributors` — an internal package importable only from within this module. As it stands, no code outside `go-application-framework` (the CLI included) can write a type that satisfies `Sink`, and there is no exported `pkg/` constructor that hands back a ready-made one either (the concrete implementation lives in `internal/contributors/sink.go`, and `internal/contributors/emitter.go`'s `New` is unexported too). The PR description says a future PR has the CLI "construct the middleware... which will own the sink" — that wiring can't happen from the CLI's own module today unless `EntityType` (or a sink constructor) gets exported from `pkg/`, or GAF itself does this wiring internally (e.g. in `pkg/app`) rather than the CLI. Worth pinning down which of those is the actual plan before this lands, since exporting `EntityType` later is its own breaking change to work through.

pkg/networking/middleware/contributor_capture/body.go
basti-snyk, line 1: ideally, this whole package would be exported in a public lib repository outside of gaf. but that's for another day.

pkg/networking/middleware/contributor_capture/fakesink_test.go
basti-snyk, line 1: this should be a gomock instead of a handcrafted mock.

pkg/networking/middleware/contributor_capture/sink.go
jonnyowenpowell: We don't want to expose the contributors API client directly, but the type will be used by the post invoke hook. It doesn't actually matter if it's never made available externally, but for the sake of allowing external tests to be written. We'll alias it from the public post invoke hook package, which is a more natural place for this domain type to be exported than from the middleware.

pkg/networking/middleware/contributor_capture/fakesink_test.go
jonnyowenpowell: Done 👍

pkg/networking/middleware/contributor_capture/contributor_capture_test.go
```diff
@@ -0,0 +1,743 @@
...
+				assert.Equal(t, tt.path, r.URL.Path)
+				w.WriteHeader(http.StatusOK)
+				_, err := w.Write([]byte(tt.responseBody(tt.expectedProjectID...)))
+				require.NoError(t, err)
+			}))
+			t.Cleanup(server.Close)
+			rt, sink := newTestMiddleware(t, http.DefaultTransport, server.URL)
+			for _, id := range tt.expectedProjectID {
+				sink.EXPECT().RecordEntity(contributors.EntityTypeProject, id)
+			}
```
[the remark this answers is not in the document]
jonnyowenpowell: ^ wrong

pkg/networking/middleware/contributor_capture/parse.go
basti-snyk: @jonnyowenpowell what about this?

jonnyowenpowell: This is all pretty fragile, because we've been forced down the path of sniffing traffic, and we don't want to load the potentially very large SARIF response into memory, so we have to parse partial JSON - there won't be a project ID to find if the result is pending/failed, so I don't think that iterating on this further actually buys us anything.

basti-snyk (approved):

PR #704, opened 2026-08-14
Wire secret-shape scrubbing into the analytics extension chokepoint

nick-y-snyk:
## Summary
- Exports `logging.Scrub` so it's reusable outside `pkg/logging` — a thin wrapper around the existing private `scrub`, no new patterns.
- Adds `analytics.WithConfiguration(cfg)`, matching the shape of the existing `WithLogger` option.
- When supplied, `sanitizeExtensionData` shape-scrubs the extension payload via `logging.GetScrubDictFromConfig(cfg)` before the existing key-name sanitization runs. Omitted, behavior is unchanged.

This closes the gap where every product line's `AddExtension*` analytics data went out with only key-name-based filtering — none of GAF's mature debug-log secret-shape patterns (Bearer tokens, PATs, basic auth, etc.) were applied to analytics. One fix at this shared chokepoint protects every current and future extension caller, including `cli-extension-axi`'s `agent feedback` command, without any of them needing their own redaction code.

Implements ticket 01 of the `agent-feedback-secret-redaction` spec tracked in `cli-extension-axi`'s `.scratch/`. Follow-on tickets (wiring `cli`'s `sendInstrumentation` call site with `WithConfiguration`, then bumping this dependency in `cli-extension-axi`) are separate, sequenced after this merges and releases.

## Test plan
- [x] `go test ./pkg/logging/... ./pkg/analytics/...`
- [x] `go test -cover ./... -race` (full suite, green)
- [x] `go vet ./pkg/analytics/... ./pkg/logging/...`
- [x] New test proves `Scrub` matches the private `scrub` path and actually redacts (not just an identity check)
- [x] Extended `instrumentation_collector_test.go`: secret-shaped extension value (no sensitive key name nearby) is redacted with `WithConfiguration`, left untouched without it

---

> [!NOTE]
> **Medium Risk**
> Changes how sensitive data may appear in outbound analytics when enabled; behavior is opt-in but touches PII/secret handling on a shared path used by many product lines.
> 
> **Overview**
> Adds **opt-in secret-shape redaction** for V2 analytics `extension` data at the shared `sanitizeExtensionData` chokepoint, so callers that pass `WithConfiguration` get the same scrub dictionary GAF already uses for debug logs (tokens, Bearer patterns, config-derived terms).
> 
> **`WithConfiguration(cfg)`** runs **before** JSON marshal: it walks extension maps/slices and applies `logging.Scrub` only on **string leaves**, avoiding whole-JSON scrubbing that could redact unrelated fields that share a substring with a secret. Recursive scrubbing includes **cycle guards** for self-referential extension values.
> 
> **`logging.Scrub`** is exported (wrapper around existing private scrub). **`REDACTION_TERMS`** config adds extra literal terms to `GetScrubDictFromConfig`. Without `WithConfiguration`, extension handling stays as before (key-name sanitization only).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ea4a4f24d1e69c9835e9be544d4e256f2c873db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/configuration/constants.go
+++ b/pkg/configuration/constants.go
@@ -53,12 +53,6 @@ const (
 	AUTHENTICATION_ADDITIONAL_URLS string = "internal_additional_auth_urls" // AUTHENTICATION_ADDITIONAL_URLS ([]string) array of additional urls to add authentication for
 
 	// ---------
-	// redaction related configuration
-	// ---------
-
-	REDACTION_TERMS string = "internal_redaction_terms" // REDACTION_TERMS ([]string) arbitrary literal terms to redact from analytics/log output, in addition to the token/OAuth-derived terms GetScrubDictFromConfig already adds
-
-	// ---------
 	// general application configuration
 	// ---------
 
--- a/pkg/configuration/constants.go
+++ b/pkg/configuration/constants.go
@@ -53,6 +53,12 @@ const (
 	AUTHENTICATION_ADDITIONAL_URLS string = "internal_additional_auth_urls" // AUTHENTICATION_ADDITIONAL_URLS ([]string) array of additional urls to add authentication for
 
 	// ---------
+	// redaction related configuration
+	// ---------
+
+	REDACTION_TERMS string = "internal_redaction_terms" // REDACTION_TERMS ([]string) arbitrary literal terms to redact from analytics/log output, in addition to the token/OAuth-derived terms GetScrubDictFromConfig already adds
+
+	// ---------
 	// general application configuration
 	// ---------
 
--- a/pkg/analytics/instrumentation_collector.go
+++ b/pkg/analytics/instrumentation_collector.go
@@ -7,6 +7,7 @@ import (
 	"github.com/snyk/go-application-framework/pkg/logging"
 	"maps"
 	"os/user"
+	"reflect"
 	"sync"
 	"time"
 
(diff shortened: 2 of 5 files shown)


pkg/configuration/constants.go
```diff
@@ -52,6 +52,12 @@ const (
 	AUTHENTICATION_SUBDOMAINS      string = "internal_auth_subdomain"       // AUTHENTICATION_SUBDOMAINS ([]string) array of additional subdomains to add authentication for
 	AUTHENTICATION_ADDITIONAL_URLS string = "internal_additional_auth_urls" // AUTHENTICATION_ADDITIONAL_URLS ([]string) array of additional urls to add authentication for
 
+	// ---------
+	// redaction related configuration
+	// ---------
+
+	REDACTION_TERMS string = "internal_redaction_terms" // REDACTION_TERMS ([]string) arbitrary literal terms to redact from analytics/log output, in addition to the token/OAuth-derived terms GetScrubDictFromConfig already adds
```
PeterSchafer: Suggestion: Move this constant to the logging package to not pollute the configuration package more then it already is

PeterSchafer (approved):
Looks good! please ensure to apply the one suggestion.

PR #707, opened 2026-08-18
fix: render the asset link once below the overall test summary

MoisBrigitte:
### Description

The UFM human readable template renders the asset inventory link inside `resultFooter`, so it appears under an individual result, before the overall test summary.

That is wrong for any test whose results are reported per component: each result carries a copy of the same run-wide link, so each one prints it. An SBOM test covering six components printed this six times, none of them at the end of the output:

```
╭───────────────────────────────────────────────────────╮
│ Test Summary                                          │
│   ...                                                 │
╰───────────────────────────────────────────────────────╯

View your asset(s) at: https://app.dev.snyk.io/org/…/overview   <- once per component

─────────────────────────────────────────────────────

╭─────────────────────────────────────────────────────────╮
│ Overall Test Summary                                    │
```

A run has a single asset, so the link now renders once below the overall test summary:

```
╰────────────────────────────────────────────────────────────╯

View your asset(s) at: https://app.snyk.io/org/my-org/inventory/assets/…/overview

💡 Tip

   To view ignored issues, use the --include-ignores option.
```

`getFirstMetadataValue(testResults, key)` reads the link from the first result that carries one. It scans rather than taking `index 0` because the producer strips the run-wide keys from every result but the last, so the first result has no asset — scanning keeps this correct whichever result carries it, or if they all do.

Verified by rendering:

| Input | Output |
|---|---|
| Multiple results, one asset repeated per component | one link, below the overall summary |
| Single result (no overall summary box) | link still last, spacing unchanged |
| No asset metadata | no link line |

The single-result path needs a conditional newline: a result footer ends its own line, the summary box does not.

Two notes for reviewers:

- **`report-url`** sits in the same block and has the same problem. Left alone to keep this focused — happy to fold it in.
- **This leaves a modelling gap open.** `asset` is a property of the *test*, but the UFM payload is a bare array of results with nowhere to put test-level facts, so it is copied onto every result and then stripped from all but one by the producer. Giving the payload a test to hang those facts on would delete both this helper and the producer-side stripping. That is a wire-format change and is prepared separately on `feat/ufm-test-metadata`, stacked on this branch.

### Checklist

- [ ] Regenerated mocks, etc. (`make generate`) — n/a, no mocked interfaces changed
- [x] Linted (`make lint`)
- [ ] Test your changes work for the CLI — see below

Notes on the unticked boxes:

- There was no test covering the asset link at all before this. Added `multi_project.with-asset.{testresult.json,human.readable}` and a `multi_project_with_asset` case in `Test_UfmPresenter_HumanReadable` — seven results all carrying the same asset, which is the shape that surfaced the bug.
- `go test ./...` passes apart from `connectivity_check_extension/connectivity`, which fails identically on a clean tree in my environment (a local `NODE_EXTRA_CA_CERTS` value leaks into `TestDetectProxyConfig`). Unrelated.
- Paired CLI PR: **snyk/cli#7149**. I have not run the CLI end to end.

---

> [!NOTE]
> **Low Risk**
> Presentation-only changes to CLI human-readable output; no auth, API, or data handling.
> 
> **Overview**
> Fixes duplicate **asset inventory** lines in UFM human-readable output when a run has multiple per-component results (e.g. SBOM multi-project scans) that all carry the same run-wide `asset` metadata.
> 
> The **“View your asset(s) at:”** line is removed from `resultFooter` and rendered once at the end of `main`, after the overall test summary box when applicable. A new template helper **`assetLink`** scans `$.TestResults` for the first non-empty `asset` metadata value instead of reading only the current result.
> 
> Coverage adds **`multi_project_with_asset`** in `Test_UfmPresenter_HumanReadable` with fixture `multi_project.with-asset.*` (seven results sharing one asset URL).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53729422c5cdf6b8abea5b995b424ee929fc236e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/presenters/funcs.go
+++ b/internal/presenters/funcs.go
@@ -357,6 +357,7 @@ func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinf
 	}
 	defaultMap["getFindingTypesFromTestResult"] = getFindingTypesFromTestResult
 	defaultMap["getFindingTypesFromMultipleTestResults"] = getFindingTypesFromMultipleTestResults
+	defaultMap["assetLink"] = assetLink
 	defaultMap["getIssuesFromTestResult"] = func(testResults testapi.TestResult, findingType ...testapi.FindingType) []testapi.Issue {
 		return utils.ValueOf(testapi.GetIssuesFromTestResult(testResults, findingType))
 	}
@@ -602,6 +603,19 @@ func getFindingTypesFromMultipleTestResults(testResults []testapi.TestResult) []
 	slices.Sort(findingTypesList)
 	slices.Reverse(findingTypesList)
 	return findingTypesList
+}
+
+func assetLink(testResults []testapi.TestResult) string {
+	for _, result := range testResults {
+		if result == nil {
+			continue
+		}
+		if link, ok := result.GetMetadataValue(testapi.TestResultMetadataKeyAsset).(string); ok && link != "" {
+			return link
+		}
+	}
+
+	return ""
 }
 
 func getDefaultFindingTypesFromConfig(testResults testapi.TestResult) []testapi.FindingType {
--- a/internal/presenters/presenter_ufm_test.go
+++ b/internal/presenters/presenter_ufm_test.go
@@ -1264,6 +1264,13 @@ func Test_UfmPresenter_HumanReadable(t *testing.T) {
 			severityThreshold: "",
 		},
 		{
+			name:              "multi_project_with_asset",
+			expectedPath:      "testdata/ufm/multi_project.with-asset.human.readable",
+			testResultPath:    "testdata/ufm/multi_project.with-asset.testresult.json",
+			includeIgnores:    false,
+			severityThreshold: "",
+		},
+		{
 			name:              "secrets_duplicated_rules",
 			expectedPath:      "testdata/ufm/secrets.duplicated-sarif-rules.human.readable",
 			testResultPath:    "testdata/ufm/secrets.duplicated-sarif-rules.testresult.json",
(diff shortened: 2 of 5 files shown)


internal/presenters/funcs.go
```diff
@@ -604,6 +605,25 @@ func getFindingTypesFromMultipleTestResults(testResults []testapi.TestResult) []
 	return findingTypesList
 }
 
+// getFirstMetadataValue returns the first non-empty string value stored under key in the
+// metadata of the given test results, or an empty string if none of them carry one.
+//
+// It is used for metadata that describes the run rather than an individual result, such as
+// the asset link, so that it is rendered once instead of under every result. A test whose
+// results are reported per component repeats the same value on each of them.
+func getFirstMetadataValue(testResults []testapi.TestResult, key string) string {
```
CatalinSnyk: suggestion/question: If every result is guaranteed to carry the same asset link, couldn't the template read it directly from the first result? That would remove the need for this function && ff that guarantee does not hold, I think we should instead collect distinct links.

internal/presenters/funcs.go
```diff
@@ -604,6 +605,19 @@ func getFindingTypesFromMultipleTestResults(testResults []testapi.TestResult) []
 	return findingTypesList
 }
 
+func getFirstMetadataValue(testResults []testapi.TestResult, key string) string {
```
CatalinSnyk: suggestion/question: If every result is guaranteed to carry the same asset link, couldn't the template read it directly from the first result? That would remove the need for this function && if that guarantee does not hold, I think we should instead collect distinct links.

MoisBrigitte: At the moment it happens to be on one component out of the array. It shouldn't be on the components in the first place - we'll refactor the model after the release

CatalinSnyk (approved):

PR #712, opened 2026-08-26
fix: prevent log scrubbing from corrupting adjacent JSON

nick-y-snyk:
## Summary

Fixes [CLI-1732](https://snyksec.atlassian.net/browse/CLI-1732): customers running `snyk monitor -d` (or any command with debug logging) intermittently see `zerolog: could not write event: cannot decode event: invalid character '*' looking for beginning of value`. Cosmetic (the underlying command still succeeds), but it makes debug output unusable for troubleshooting.

Two independent bugs in `pkg/logging/scrubbingLogWriter.go`, both stemming from redaction not respecting the boundaries of the already-serialized JSON log event it's operating on:

- **`scrub()`** redacted a matched capture group by re-searching the whole string for the captured *text* and replacing every occurrence, instead of replacing just the matched *span*. A short or structural redacted value (a one-character field value that happens to equal `{`, `,`, or `"`) could wipe out identical text elsewhere in the same event, corrupting the JSON. This reproduces the customer's exact error string.
- **`addMandatoryMasking`** had unbounded greedy captures in two places (the legacy snyk-config `_: [...]` debug dump, and the `'p='/'u='` / `-p`/`-u` short forms) that matched through to the *last* delimiter anywhere later in the event instead of their own closing one, swallowing any JSON that followed.

## Changes

- `scrub()` now redacts by matched position (`FindAllStringSubmatchIndex` + span rebuild) rather than re-searching for captured text.
- Bounded the two greedy regexes so they stop at their own delimiter; split the single-pattern short-form quote matcher into single- and double-quoted variants since a shared bound needs backreferences RE2 doesn't support.
- Added 9 new tests reproducing both failure modes and asserting the scrubbed output stays valid JSON; all 85 pre-existing tests pass unchanged.

## Test plan
- [x] `go test ./pkg/logging/...` — 85 passing (sub)tests, 0 failures
- [x] `go test ./pkg/analytics/...` (only other consumer of the scrub path) — clean
- [x] `go build ./...`, `go vet`, `golangci-lint run ./pkg/logging/...`, `gofmt -l` — all clean
- [x] Verified new tests actually catch the regression: reverting only the source file fails 8/9 new subtests, with the pre-fix failure being the customer's literal error string

[CLI-1732]: https://snyksec.atlassian.net/browse/CLI-1732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes core log redaction behavior on every debug/analytics write path; fixes are well-tested but incorrect regex edge cases could still leak secrets or skip redaction.
> 
> **Overview**
> Fixes debug-log corruption where zerolog could fail with invalid JSON after scrubbing (**CLI-1732**).
> 
> **Position-based redaction:** Regex redaction no longer replaces every occurrence of the captured *text* in the event. A new `redactMatchedGroup` helper rewrites only the matched byte span (via `FindAllStringSubmatchIndex`), so one-character or structural secret values like `{`, `,`, or `"` cannot wipe the same characters elsewhere in the JSON.
> 
> **Tighter mandatory masking patterns:** Greedy captures in the legacy snyk-config `_: [...]` dump and in `u`/`p` short-form patterns are bounded so matches stop at their own delimiters (quoted spans, nested `[...]`, escaped quotes) instead of running to the last `]`, `'`, or `"` in the whole line/event. CLI `-u`/`-p` values similarly stop at unescaped `"` while still treating escaped quotes as part of the value.
> 
> **Tests:** Adds CLI-1732 regression suites that assert scrubbed output stays valid JSON and only the intended fields change, plus cases for escaped quotes and bracket dumps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 951924fe58ba877bbd2ee7af3595b79fa3be408a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/logging/scrubbingLogWriter.go
+++ b/pkg/logging/scrubbingLogWriter.go
@@ -273,7 +273,12 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
 
 	// The legacy CLI's snyk-config package prints the entire configuration in debug mode.
 	// It begins with some pseudo-JSON structure, which we can redact.
-	s = `(?s)_:\s*\[(?<everything_inside_hard_brackets>.*)\]`
+	// The capture must not be a plain greedy `.*`: that runs to the *last* `]` in the whole
+	// buffer and swallows any structure that follows the dump (see CLI-1732). Instead the
+	// capture may cross a balanced `[...]` pair (the dump nests one), but never an unpaired
+	// closing bracket, so it always stops at the dump's own `]`. The final `\[` alternative
+	// keeps a stray, unbalanced `[` inside the dump from failing the match altogether.
+	s = `_:\s*\[(?<everything_inside_hard_brackets>(?:\[[^\]]*\]|[^\[\]]|\[)*)\]`
 	dict[s] = scrubStruct{
 		groupToRedact: 1,
 		regex:         regexp.MustCompile(s),
@@ -290,7 +295,7 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
 	// Same as above, only with short form
 	shorts := []string{"p", "u"}
 	shortForm := strings.Join(shorts, "")
-	s = fmt.Sprintf(`(?im)'[%s]=(?<value>.*)'`, shortForm)
+	s = fmt.Sprintf(`(?im)'[%s]=(?<value>[^'\n]*)'`, shortForm)
 	dict[s] = scrubStruct{
 		groupToRedact: 2,
 		regex:         regexp.MustCompile(s),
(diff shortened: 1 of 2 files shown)


pkg/logging/scrubbingLogWriter.go
```diff
@@ -299,17 +304,28 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
...
-		groupToRedact: 2,
-		regex:         regexp.MustCompile(s),
+	//
+	// Single- and double-quoted values get their own pattern so each value can be bounded by
+	// its own quote character. A single pattern bounded by `['"]` needs a greedy `.*` to allow
+	// the opposite quote inside the value, and that greedy run reaches the *last* quote on the
+	// line, swallowing every field that follows it (see CLI-1732). `\n` stays excluded so the
+	// bounded classes keep the line-at-a-time reach the previous `.` had.
+	for _, quote := range []string{`'`, `"`} {
+		s = fmt.Sprintf(`(?i)(?<short_form_key>\b[%s]\b)[,'":]+\s*(?:%s(?<short_form_value>[^%s\n]*)%s|([^,'"\s]+))[,}]?`, shortForm, quote, quote, quote)
```
robertolopezlopez: Could you please unit test this regexp? with strings containing escaped characters such as `Nick\'s`

pkg/logging/scrubbingLogWriter.go
```diff
@@ -273,7 +273,12 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
 
 	// The legacy CLI's snyk-config package prints the entire configuration in debug mode.
 	// It begins with some pseudo-JSON structure, which we can redact.
-	s = `(?s)_:\s*\[(?<everything_inside_hard_brackets>.*)\]`
+	// The capture must not be a plain greedy `.*`: that runs to the *last* `]` in the whole
+	// buffer and swallows any structure that follows the dump (see CLI-1732). Instead the
+	// capture may cross a balanced `[...]` pair (the dump nests one), but never an unpaired
+	// closing bracket, so it always stops at the dump's own `]`. The final `\[` alternative
+	// keeps a stray, unbalanced `[` inside the dump from failing the match altogether.
+	s = `_:\s*\[(?<everything_inside_hard_brackets>(?:\[[^\]]*\]|[^\[\]]|\[)*)\]`
```
robertolopezlopez: Could you please add a unit test checking that ] inside a quoted config value does not stop redaction too early?

robertolopezlopez (changes requested):
a couple of remarks, which were actually done by Cursor before

pkg/logging/scrubbingLogWriter.go
nick-y-snyk: Added a regression test for this (single-quoted value with a backslash-escaped apostrophe, e.g. `'u': 'Nick\'s'`). Turned up a real gap while I was at it: the value class didn't treat an escaped bounding-quote as part of the value, so it stopped early and leaked the rest. Fixed to match the same `\\.` escape idiom the CLI-argument pattern below already uses.

pkg/logging/scrubbingLogWriter.go
nick-y-snyk: Added regression tests for this (a `]` and a `[` inside a quoted dump value). This one was also a real bug, not just missing coverage: the balanced-bracket tracking treated a bracket inside a quoted value as real nesting and closed the capture early, leaking the rest of the entry. Fixed by consuming quoted spans as one atomic unit before the bracket-tracking alternatives get a look.

robertolopezlopez (approved):

pkg/logging/scrubbingLogWriter.go
```diff
@@ -273,7 +273,15 @@ func addMandatoryMasking(dict ScrubbingDict) ScrubbingDict {
...
-	s = `(?s)_:\s*\[(?<everything_inside_hard_brackets>.*)\]`
+	// The capture must not be a plain greedy `.*`: that runs to the *last* `]` in the whole
+	// buffer and swallows any structure that follows the dump (see CLI-1732). Instead the
+	// capture may cross a balanced `[...]` pair (the dump nests one), but never an unpaired
+	// closing bracket, so it always stops at the dump's own `]`. The final `\[` alternative
+	// keeps a stray, unbalanced `[` inside the dump from failing the match altogether.
+	// Quoted spans are consumed whole, before the bracket-tracking alternatives get a look —
+	// otherwise a `]` or `[` inside a quoted value (e.g. `'a]b'`) reads as real bracket nesting
+	// and the capture closes early, right at that inner bracket.
+	s = `_:\s*\[(?<everything_inside_hard_brackets>(?:'[^']*'|"[^"]*"|\[[^\]]*\]|[^\[\]]|\[)*)\]`
```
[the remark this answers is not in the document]
nick-y-snyk: Confirmed real, fixed in 951924f — the quoted-span classes now use the same (?:[^'\\]|\\.)* escape idiom as the sibling short-form/CLI patterns, and a regression test is added.

PR #713, opened 2026-08-27
fix: prevent static-term scrubbing from corrupting JSON

nick-y-snyk:
## Summary

Follow-up to #712 (CLI-1732). That fix made the regex-entry path in `scrub()` position-aware. The static-term branch (`AddTermsToReplace` / `GetScrubDictFromConfig` -- tokens, client secret/id, `REDACTION_TERMS`) still did a blind `strings.ReplaceAll`, with no JSON-quote awareness.

Fixed:

- A registered term matching a bare (unquoted) JSON value -- a number, bool, or null -- got swapped for unquoted `***`, producing invalid JSON (`"count":12345` -> `"count":***`). New `RedactStaticTerm` quotes the replacement when the match sits in a bare JSON value position.
- A term that's only a digit-fused slice of a larger number (e.g. redacting "1000" out of an unrelated "durationMs":10001234) is left alone instead of splitting one JSON number into two invalid pieces. A numeric username is a realistic trigger for this: some container setups set `$USER` to a raw UID with no matching `/etc/passwd` entry, not because `os/user.Current` itself falls back to one.
- A writer downstream of a non-JSON formatter (e.g. zerolog's `ConsoleWriter`, which decodes JSON and re-emits human-readable text before the scrubber sees it) was still getting JSON-only quoting rules applied to plain prose, corrupting it. `internalWrite` and the exported `Scrub` now decide JSON-aware mode from `json.Valid` on the actual bytes, not from what the caller assumed it was passing.
- New `ScrubValue` scrubs a single non-JSON leaf value (e.g. an analytics extension field before marshaling) without the JSON-value quoting or digit-fusion protections, which would otherwise stray-quote or under-redact plain text.
- `pkg/analytics.SanitizeStaticValues` had the identical bug on already-marshaled JSON, with two production call sites (`analytics.go`, `instrumentation_collector.go`) feeding numeric fields like `durationMs`. It now routes through the same `logging.RedactStaticTerm` helper when its input is valid JSON, and falls back to plain replacement otherwise.
- Quote/escape state tracking (`quoteState`) now carries correctly across `RedactStaticTerm`'s two scan calls per match (the text before it, then the match itself). A trailing unresolved backslash at the boundary between those two scans used to have its escaped character re-examined as fresh input by the next scan, desyncing whether a later match was inside a quoted string.

Two edge cases are deliberately left alone, both inherent to splicing an exact literal match into JSON text rather than parsing it: a redaction term that is itself a bare `"` character, and a term whose first character lands on the target of a preceding unresolved backslash escape.

## Test plan

- [x] `go test ./...` -- full module green
- [x] `make lint` -- 0 issues
- [x] `snyk code test` -- no new findings
- [x] New tests: `TestStaticTermReplacementPreservesJSONValidity`, `TestQuoteState_CarriesEscapeAcrossSeparateScans`, `TestScrub_NonJSONInputStillRedactsStaticTerms`, `Test_SanitizeUsername_NumericUsernameCollision`, `Test_GetRequest_RedactsNumericUsernameFromMarshaledPayload`, plus empty-term/empty-value no-op guards for both `RedactStaticTerm` and `SanitizeStaticValues`

---

> [!NOTE]
> **Medium Risk**
> Touches shared log/analytics redaction on every scrubbed write and marshaled analytics payload; behavior changes are intentional but subtle (JSON vs non-JSON, fusion skips).
> 
> **Overview**
> Replaces blind `strings.ReplaceAll` for configured literal redaction terms with **`RedactStaticTerm`**, which quotes replacements in bare JSON value positions, skips digit-fused substrings inside larger numbers/keywords, and tracks quote/escape state across matches. **`Scrub`** and log **`internalWrite`** enable that behavior only when `json.Valid` on the actual bytes; **`ScrubValue`** is added for pre-marshal string leaves (analytics extension scrubbing now uses it instead of **`Scrub`**).
> 
> **`SanitizeStaticValues`** in analytics follows the same JSON gate and **`RedactStaticTerm`** path, with empty-term no-ops, so numeric usernames no longer split unrelated JSON numbers or inject stray quotes into plain-text snippets.
> 
> Extensive regression tests cover numeric-username collisions, **`GetRequest`** redaction, console-writer non-JSON prose, and **`RedactStaticTerm`** edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed24e1e578957b02cd9e06c1104b864a756fadc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -368,11 +368,20 @@ func SanitizeUsername(rawUserName string, userHomeDir string, replacementValue s
 	return SanitizeStaticValues(valuesToSanitize, replacementValue, content)
 }
 
+// SanitizeStaticValues is exported API, so content isn't guaranteed to be the full, valid JSON its
+// two current callers (GetRequest, GetV2InstrumentationObject) always pass -- both feed it
+// json.Marshal output. RedactStaticTerm's bare-value quoting is only correct against real JSON; a
+// non-JSON snippet falls back to a plain literal replace, the same non-JSON path ScrubValue uses.
 func SanitizeStaticValues(valuesToSanitize []string, replacementValue string, content []byte) ([]byte, error) {
 	contentStr := string(content)
+	jsonAware := json.Valid(content)
 
 	for _, valueToReplace := range valuesToSanitize {
-		contentStr = logging.RedactStaticTerm(contentStr, valueToReplace, replacementValue)
+		if jsonAware {
+			contentStr = logging.RedactStaticTerm(contentStr, valueToReplace, replacementValue)
+		} else {
+			contentStr = strings.ReplaceAll(contentStr, valueToReplace, replacementValue)
+		}
 	}
 
 	return []byte(contentStr), nil
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -377,6 +377,9 @@ func SanitizeStaticValues(valuesToSanitize []string, replacementValue string, co
 	jsonAware := json.Valid(content)
 
 	for _, valueToReplace := range valuesToSanitize {
+		if valueToReplace == "" {
+			continue // strings.ReplaceAll inserts replacementValue between every rune for an empty old string
+		}
 		if jsonAware {
 			contentStr = logging.RedactStaticTerm(contentStr, valueToReplace, replacementValue)
 		} else {
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -368,10 +368,10 @@ func SanitizeUsername(rawUserName string, userHomeDir string, replacementValue s
 	return SanitizeStaticValues(valuesToSanitize, replacementValue, content)
 }
 
-// SanitizeStaticValues is exported API, so content isn't guaranteed to be the full, valid JSON its
-// two current callers (GetRequest, GetV2InstrumentationObject) always pass -- both feed it
-// json.Marshal output. RedactStaticTerm's bare-value quoting is only correct against real JSON; a
-// non-JSON snippet falls back to a plain literal replace, the same non-JSON path ScrubValue uses.
+// SanitizeStaticValues replaces every occurrence of each valuesToSanitize entry in content with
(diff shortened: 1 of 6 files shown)


robertolopezlopez:
hi, could you please run this test?

```go
func TestStaticTermReplacementPreservesJSONScalarValidity(t *testing.T) {
  	tests := []struct {
  		name     string
  		input    string
  		term     string
  		expected string
  	}{
  		{
  			name:     "top-level number",
  			input:    `12345`,
  			term:     "12345",
  			expected: `"***"`,
  		},
  		{
  			name:     "partial true literal",
  			input:    `{"enabled":true}`,
  			term:     "rue",
  			expected: `{"enabled":true}`,
  		},
  		{
  			name:     "partial false literal",
  			input:    `{"enabled":false}`,
  			term:     "alse",
  			expected: `{"enabled":false}`,
  		},
  		{
  			name:     "partial null literal",
  			input:    `{"value":null}`,
  			term:     "ull",
  			expected: `{"value":null}`,
  		},
  	}

  	for _, test := range tests {
  		t.Run(test.name, func(t *testing.T) {
  			var output bytes.Buffer
  			writer := NewScrubbingIoWriter(&output, ScrubbingDict{})

  			scrubbingWriter, ok := writer.(ScrubbingLogWriter)
  			require.True(t, ok)
  			scrubbingWriter.AddTermsToReplace([]string{test.term})

  			_, err := writer.Write([]byte(test.input))
  			require.NoError(t, err)

  			assert.True(
  				t,
  				json.Valid(output.Bytes()),
  				"scrubbing produced invalid JSON: %s",
  				output.Bytes(),
  			)
  			assert.Equal(t, test.expected, output.String())
  		})
  	}
  }
```

robertolopezlopez (changes requested):
I found some corner cases, please check my other comment

robertolopezlopez (approved):
good work!

PeterSchafer:
In the future please squash and merge!

PR #716, opened 2026-08-29
feat: Reduce requests for feature flag lookups

PeterSchafer:
### Description

`AddFeatureFlagsToConfig` looked flags up in one batch per call, so an application that registered flags from several places produced one request per registration group rather than one request overall.

Flags registered against a configuration are now collected in a registry, and the lookup happens lazily per flag:

- The first access to any backed config key looks up **every flag registered so far** in a single request.
- Flags registered **after** an earlier lookup are looked up by the next access that needs them, and flags already known are not requested again.
- Results are memorized per organization **and API endpoint**, so a clone addressing another endpoint does not reuse them, and they are shared across clones of the same configuration so a cloned config does not repeat a lookup.
- Flags the gateway does not report are memorized as disabled, so an unknown flag resolves to `false` instead of being requested on every read.

The registry is attached to the `Configuration` as an immutable default value rather than as a value, so it cannot be persisted or shadowed, and it is shared by every clone.

#### Fixed along the way

An earlier iteration of this branch memorized only *whether* an organization had been looked up. Any flag registered after the first access then resolved to `false` with no request and no error. This was not theoretical: the framework's own extension initializers all run before those of the consuming application, and one of them reads a feature flag while initializing, which sealed the set of known flags during `engine.Init()`. Every flag registered by a downstream extension would have resolved to `false`. `pkg/app/app_test.go` now carries an integration test that registers a flag from a downstream-style extension initializer and asserts it resolves; it fails against that earlier implementation.

#### Notes for reviewers and consumers

- **No API change.** `AddFeatureFlagsToConfig` keeps its signature.
- **Behaviour change:** lookup results are now shared across all clones of a configuration for the lifetime of the process, scoped by organization and API endpoint. There is no TTL, and `ClearCache()` does not clear them.
- **Test fixtures downstream:** the function now calls `Get` and `AddDefaultValue` on the `Configuration`. A test that wires a strict `MockConfiguration` into it will need the corresponding `EXPECT()` calls. No test in this repository is affected.
- **Request count:** for a `snyk test` this brings the lookups down from three requests to two, not one. The remaining request exists because a framework extension reads a flag during `Init()`, before the consuming application has registered its own. Removing that early read is a follow-up, since it changes when and how that extension registers its middleware and hook.

Known trade-offs left for follow-ups: the per-organization lock is held across the request, `EvaluateFlags` has no timeout, a failing lookup is retried per read rather than negatively cached, and one rejected flag name would now fail a whole batch rather than a single registration group.

### Checklist

- [x] Linted (`make lint`)

---

> [!NOTE]
> **Medium Risk**
> Behavior change around when and how often feature flags are fetched, with process-lifetime caching that `ClearCache()` does not reset; incorrect batching or ordering could mis-resolve flags for downstream extensions at init time.
> 
> **Overview**
> **`AddFeatureFlagsToConfig`** no longer issues one gateway request per registration call. Flags for a configuration are accumulated in a shared **`flagRegistry`** (attached as an immutable default on `Configuration`), and the first read of any backed key triggers a **single batched** remote evaluation of every flag registered so far.
> 
> Later registrations are resolved on the next access that needs them; results are **memoized per org and API URL** and reused across configuration clones (a different API URL gets a separate lookup). Flags omitted from the gateway response are cached as **false** so they are not re-requested. Config keys that are already set locally (via `Configuration.IsSet`) are **excluded from batches** so forced on/off values are not evaluated remotely.
> 
> Integration tests in **`app_test`** cover downstream extension flags after `engine.Init()` and locally overridden gitignore flags during startup batching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d926863a1b61d4f7f3a23aa9e0b6bb83a47f4920. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/local_workflows/config_utils/feature_flag_gateway.go
+++ b/pkg/local_workflows/config_utils/feature_flag_gateway.go
@@ -44,18 +44,25 @@ var registries sync.Map
 
 func getFlagRegistry(config configuration.Configuration) *flagRegistry {
 	if r, ok := registries.Load(config); ok {
-		return r.(*flagRegistry)
+		if reg, ok := r.(*flagRegistry); ok {
+			return reg
+		}
 	}
 	return nil
 }
 
 func getOrCreateFlagRegistry(config configuration.Configuration) *flagRegistry {
 	if r, ok := registries.Load(config); ok {
-		return r.(*flagRegistry)
+		if reg, ok := r.(*flagRegistry); ok {
+			return reg
+		}
 	}
 	created := &flagRegistry{flags: make(map[string]struct{})}
 	actual, _ := registries.LoadOrStore(config, created)
-	return actual.(*flagRegistry)
+	if reg, ok := actual.(*flagRegistry); ok {
+		return reg
+	}
+	return created
 }
 
 func AddFeatureFlagsToConfig(
@@ -109,7 +116,10 @@ func AddFeatureFlagsToConfig(
 				return false, fmt.Errorf("check feature flags batch: %w", err)
 			}
 
-			res := result.(map[string]bool)
+			res, ok := result.(map[string]bool)
+			if !ok {
+				return false, fmt.Errorf("check feature flags batch: unexpected result type")
+			}
 			return res[flagName], nil
 		}
 		config.AddDefaultValue(configKey, callback)
--- a/pkg/local_workflows/config_utils/feature_flag_gateway.go
+++ b/pkg/local_workflows/config_utils/feature_flag_gateway.go
@@ -3,28 +3,73 @@ package config_utils
 import (
 	"errors"
 	"fmt"
-	"maps"
-	"slices"
-	"sort"
-	"strings"
+	"sync"
 
 	"github.com/google/uuid"
+	"golang.org/x/sync/singleflight"
+
 	featureflaggateway "github.com/snyk/go-application-framework/pkg/apiclients/feature_flag_gateway"
 	v20241015 "github.com/snyk/go-application-framework/pkg/apiclients/feature_flag_gateway/2024-10-15"
 	"github.com/snyk/go-application-framework/pkg/configuration"
+	"github.com/snyk/go-application-framework/pkg/utils"
 	"github.com/snyk/go-application-framework/pkg/workflow"
 )
 
 var evaluateFlags = featureflaggateway.EvaluateFlags
 var errInvalidEvaluateFlagsResponse = errors.New("invalid evaluateFlags response")
 
+type flagRegistry struct {
+	mu    sync.Mutex
+	flags map[string]struct{}
+	sf    singleflight.Group
+}
+
+func (r *flagRegistry) addFlags(names []string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for _, name := range names {
+		r.flags[name] = struct{}{}
+	}
+}
+
+func (r *flagRegistry) allFlags() []string {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return utils.SortedMapKeys(r.flags)
+}
+
+// registries maps Configuration instances to their flag registry.
+// Stored outside the config keyspace to avoid env-var collisions and AllKeys() pollution.
+var registries sync.Map
+
+func getFlagRegistry(config configuration.Configuration) *flagRegistry {
+	if r, ok := registries.Load(config); ok {
+		return r.(*flagRegistry)
+	}
+	return nil
+}
+
+func getOrCreateFlagRegistry(config configuration.Configuration) *flagRegistry {
+	if r, ok := registries.Load(config); ok {
+		return r.(*flagRegistry)
+	}
+	created := &flagRegistry{flags: make(map[string]struct{})}
+	actual, _ := registries.LoadOrStore(config, created)
+	return actual.(*flagRegistry)
+}
+
 func AddFeatureFlagsToConfig(
 	engine workflow.Engine,
 	configKeyToFlag map[string]string,
 ) {
 	config := engine.GetConfiguration()
-	flags := slices.Collect(maps.Values(configKeyToFlag))
-	sort.Strings(flags)
+	registry := getOrCreateFlagRegistry(config)
+
+	flagNames := make([]string, 0, len(configKeyToFlag))
+	for _, flagName := range configKeyToFlag {
+		flagNames = append(flagNames, flagName)
+	}
+	registry.addFlags(flagNames)
 
 	for configKey, flagName := range configKeyToFlag {
 		err := config.AddKeyDependency(configKey, configuration.ORGANIZATION)
@@ -38,22 +83,33 @@ func AddFeatureFlagsToConfig(
 			}
 
 			orgID := c.GetString(configuration.ORGANIZATION)
-			cacheKey := fmt.Sprintf("hidden_flags_%s:%s", orgID, strings.Join(flags, ","))
+			cacheKey := fmt.Sprintf("hidden_flags_%s", orgID)
+
 			if cached := c.Get(cacheKey); cached != nil {
 				if m, ok := cached.(map[string]bool); ok {
-					return m[flagName], nil
+					if _, exists := m[flagName]; exists {
+						return m[flagName], nil
+					}
 				}
 			}
 
-			res, err := areFeaturesEnabled(c, engine, orgID, flags...)
+			result, err, _ := registry.sf.Do(cacheKey, func() (interface{}, error) {
+				allFlags := registry.allFlags()
+				res, err := areFeaturesEnabled(c, engine, orgID, allFlags...)
+				if err != nil {
+					return nil, err
+				}
+				if addErr := c.AddKeyDependency(cacheKey, configuration.ORGANIZATION); addErr != nil {
+					engine.GetLogger().Err(addErr).Msgf("failed to add dependency for %s", cacheKey)
+				}
+				c.Set(cacheKey, res)
+				return res, nil
+			})
 			if err != nil {
 				return false, fmt.Errorf("check feature flags batch: %w", err)
 			}
-			if err := config.AddKeyDependency(cacheKey, configuration.ORGANIZATION); err != nil {
-				engine.GetLogger().Err(err).Msgf("failed to add dependency for %s", cacheKey)
-			}
-			c.Set(cacheKey, res)
 
+			res := result.(map[string]bool)
 			return res[flagName], nil
 		}
 		config.AddDefaultValue(configKey, callback)
@@ -81,10 +137,24 @@ func areFeaturesEnabled(
 	}
 
 	results := make(map[string]bool, len(flags))
+	for _, f := range flags {
+		results[f] = false
+	}
+
 	evaluations := resp.ApplicationvndApiJSON200.Data.Attributes.Evaluations
+	evaluated := make(map[string]struct{}, len(evaluations))
 	for _, e := range evaluations {
+		evaluated[e.Key] = struct{}{}
 		if e.Value != nil {
 			results[e.Key] = *e.Value
+		}
+	}
+
+	if engine != nil {
+		for _, f := range flags {
+			if _, ok := evaluated[f]; !ok {
+				engine.GetLogger().Debug().Msgf("feature flag %q: no evaluation returned, defaulting to false", f)
+			}
 		}
 	}
 
--- a/pkg/local_workflows/config_utils/feature_flag_gateway.go
+++ b/pkg/local_workflows/config_utils/feature_flag_gateway.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	"github.com/google/uuid"
+
 	featureflaggateway "github.com/snyk/go-application-framework/pkg/apiclients/feature_flag_gateway"
 	v20241015 "github.com/snyk/go-application-framework/pkg/apiclients/feature_flag_gateway/2024-10-15"
 	"github.com/snyk/go-application-framework/pkg/configuration"
@@ -20,32 +21,51 @@ var errInvalidEvaluateFlagsResponse = errors.New("invalid evaluateFlags response
 
 const registryConfigKey = "hidden_feature_flag_registry"
 
+// registryMu guards the get-or-create of a registry, because Configuration
+// offers no atomic equivalent. It holds no state of its own.
 var registryMu sync.Mutex
 
+// fetchKey scopes lookup results. The API endpoint is part of the key because
+// cloned configurations may address different endpoints while sharing an org.
+type fetchKey struct {
+	orgID  string
+	apiURL string
+}
+
+// flagRegistry collects every flag registered against one configuration, so that
+// an access can look all of them up in a single request.
 type flagRegistry struct {
 	mu       sync.RWMutex
 	flags    map[string]struct{}
-	orgFetch sync.Map
+	orgFetch map[fetchKey]*orgFetchResult
 }
 
+// orgFetchResult memorizes the flags already looked up for one fetchKey. A flag
+// present in values has been looked up, an absent one still needs fetching.
 type orgFetchResult struct {
 	mu     sync.Mutex
 	values map[string]bool
-	done   bool
 }
 
 func getOrCreateRegistry(config configuration.Configuration) *flagRegistry {
 	registryMu.Lock()
 	defer registryMu.Unlock()
 
-	if v := config.Get(registryConfigKey); v != nil {
-		if r, ok := v.(*flagRegistry); ok {
-			return r
-		}
+	if registry, ok := config.Get(registryConfigKey).(*flagRegistry); ok {
+		return registry
 	}
-	r := &flagRegistry{flags: make(map[string]struct{})}
-	config.Set(registryConfigKey, r)
-	return r
+
+	registry := &flagRegistry{
+		flags:    make(map[string]struct{}),
+		orgFetch: make(map[fetchKey]*orgFetchResult),
+	}
+
+	// The registry is registered as an immutable default rather than being set as
+	// a value, so that it never enters the configuration's value space where it
+	// could be persisted or shadowed.
+	config.AddDefaultValue(registryConfigKey, configuration.ImmutableDefaultValueFunction(registry))
+
+	return registry
 }
 
 func (r *flagRegistry) addFlags(flags []string) {
@@ -67,26 +87,68 @@ func (r *flagRegistry) allFlags() []string {
 	return result
 }
 
-func (r *flagRegistry) resolve(c configuration.Configuration, engine workflow.Engine, orgID string, flagName string) (bool, error) {
-	fetchI, _ := r.orgFetch.LoadOrStore(orgID, &orgFetchResult{})
-	f := fetchI.(*orgFetchResult)
+func (r *flagRegistry) fetchResultFor(key fetchKey) *orgFetchResult {
+	r.mu.Lock()
+	defer r.mu.Unlock()
 
-	f.mu.Lock()
-	defer f.mu.Unlock()
-
-	if !f.done {
-		allFlags := r.allFlags()
-		var err error
-		f.values, err = areFeaturesEnabled(c, engine, orgID, allFlags...)
-		if err != nil {
-			return false, fmt.Errorf("check feature flags batch: %w", err)
-		}
-		f.done = true
+	result, ok := r.orgFetch[key]
+	if !ok {
+		result = &orgFetchResult{values: make(map[string]bool)}
+		r.orgFetch[key] = result
 	}
-
-	return f.values[flagName], nil
+	return result
 }
 
+// resolve returns the value of flagName and looks up every flag that is
+// registered but not yet known in a single request. Flags registered after an
+// earlier lookup are picked up here, flags already looked up are not requested
+// again.
+func (r *flagRegistry) resolve(c configuration.Configuration, engine workflow.Engine, orgID string, flagName string) (bool, error) {
+	fetched := r.fetchResultFor(fetchKey{orgID: orgID, apiURL: c.GetString(configuration.API_URL)})
+
+	fetched.mu.Lock()
+	defer fetched.mu.Unlock()
+
+	if value, known := fetched.values[flagName]; known {
+		return value, nil
+	}
+
+	missing := make([]string, 0)
+	for _, name := range r.allFlags() {
+		if _, known := fetched.values[name]; !known {
+			missing = append(missing, name)
+		}
+	}
+
+	// Requesting no flags at all would be rejected by the gateway and, since
+	// nothing would be memorized, repeated on every single access.
+	if len(missing) == 0 {
+		return false, nil
+	}
+
+	values, err := areFeaturesEnabled(c, engine, orgID, missing...)
+	if err != nil {
+		return false, fmt.Errorf("check feature flags batch: %w", err)
+	}
+
+	// Every requested flag is memorized, including the ones the gateway did not
+	// return, so that an unknown flag resolves to false instead of being
+	// requested again on every access.
+	for _, name := range missing {
+		fetched.values[name] = values[name]
+	}
+
+	return fetched.values[flagName], nil
+}
+
+// AddFeatureFlagsToConfig registers configuration keys that are backed by remote
+// feature flags. Flags are looked up lazily: the first access to any of them
+// looks up every flag registered so far in a single request, and flags
+// registered afterwards are picked up by the next access that needs them.
+//
+// Registration has to happen before a configuration is cloned. Clone copies the
+// default value functions that exist at that moment, so a configuration cloned
+// before a flag is registered resolves that flag to false.
 func AddFeatureFlagsToConfig(
 	engine workflow.Engine,
 	configKeyToFlag map[string]string,
--- a/pkg/local_workflows/config_utils/feature_flag_gateway.go
+++ b/pkg/local_workflows/config_utils/feature_flag_gateway.go
@@ -3,8 +3,6 @@ package config_utils
 import (
 	"errors"
 	"fmt"
-	"maps"
-	"slices"
 	"sort"
 	"sync"
 
@@ -35,9 +33,17 @@ type fetchKey struct {
 // flagRegistry collects every flag registered against one configuration, so that
 // an access can look all of them up in a single request.
 type flagRegistry struct {
-	mu       sync.RWMutex
-	flags    map[string]struct{}
+	mu sync.RWMutex
+	// flags maps a flag name to the configuration key it backs.
+	flags    map[string]string
 	orgFetch map[fetchKey]*orgFetchResult
+}
+
+// registeredFlag is a remote feature flag together with the configuration key it
+// backs.
+type registeredFlag struct {
+	name      string
+	configKey string
 }
 
 // orgFetchResult memorizes the flags already looked up for one fetchKey. A flag
@@ -56,7 +62,7 @@ func getOrCreateRegistry(config configuration.Configuration) *flagRegistry {
 	}
 
 	registry := &flagRegistry{
-		flags:    make(map[string]struct{}),
+		flags:    make(map[string]string),
 		orgFetch: make(map[fetchKey]*orgFetchResult),
 	}
 
@@ -68,22 +74,22 @@ func getOrCreateRegistry(config configuration.Configuration) *flagRegistry {
 	return registry
 }
 
-func (r *flagRegistry) addFlags(flags []string) {
+func (r *flagRegistry) addFlags(configKeyToFlag map[string]string) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	for _, f := range flags {
-		r.flags[f] = struct{}{}
+	for configKey, flagName := range configKeyToFlag {
+		r.flags[flagName] = configKey
 	}
 }
 
-func (r *flagRegistry) allFlags() []string {
+func (r *flagRegistry) allFlags() []registeredFlag {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	result := make([]string, 0, len(r.flags))
-	for f := range r.flags {
-		result = append(result, f)
+	result := make([]registeredFlag, 0, len(r.flags))
+	for flagName, configKey := range r.flags {
+		result = append(result, registeredFlag{name: flagName, configKey: configKey})
 	}
-	sort.Strings(result)
+	sort.Slice(result, func(i, j int) bool { return result[i].name < result[j].name })
 	return result
 }
 
@@ -102,7 +108,7 @@ func (r *flagRegistry) fetchResultFor(key fetchKey) *orgFetchResult {
 // resolve returns the value of flagName and looks up every flag that is
 // registered but not yet known in a single request. Flags registered after an
 // earlier lookup are picked up here, flags already looked up are not requested
-// again.
+// again, and neither are flags whose configuration key is locally overridden.
 func (r *flagRegistry) resolve(c configuration.Configuration, engine workflow.Engine, orgID string, flagName string) (bool, error) {
 	fetched := r.fetchResultFor(fetchKey{orgID: orgID, apiURL: c.GetString(configuration.API_URL)})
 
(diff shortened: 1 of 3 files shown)


pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -18,13 +17,43 @@ import (
...
 	engine workflow.Engine,
 	configKeyToFlag map[string]string,
 ) {
 	config := engine.GetConfiguration()
-	flags := slices.Collect(maps.Values(configKeyToFlag))
-	sort.Strings(flags)
+
+	registry := getOrCreateFlagRegistry(config)
+
+	registry.mu.Lock()
```
basti-snyk: **[16 lenses, unanimous]** blocking.

`getOrCreateFlagRegistry` can return nil and this line dereferences it. Line 41 re-reads the key it just wrote, and that read only round-trips on a `Configuration` that stores an arbitrary `interface{}`.

Thirteen lens passes reproduced the panic. Three separate triggers, and only the first needs a test double:

1. Any `Configuration` that isn't `*extendedViper`, including the `pkg/mocks.MockConfiguration` this repo ships and downstream repos wire into their tests.
2. A real `extendedViper` where anything already sits under `hidden_feature_flag_registry`. `WithoutOverriding()` declines to write, the type assertion fails, nil comes back.
3. `HIDDEN_FEATURE_FLAG_REGISTRY` set in the environment. The engine config is built with `WithAutomaticEnv()` and `WithFiles("snyk")` via `app.go:454`, so viper resolves the key and `IsSet` returns true. That panics the CLI during init at `app.go:415`.

```
panic: runtime error: invalid memory address or nil pointer dereference
	config_utils.AddFeatureFlagsToConfig(...)
		feature_flag_gateway.go:52
```

Deleting only lines 50-56 and re-running the same test makes it pass, so this arrived with this commit. Before the PR the function only called `AddKeyDependency` and `AddDefaultValue`, which any `Configuration` satisfies. Under the AGENTS.md stability checklist that is a change to behavior consumers rely on, and it isn't marked breaking.

Suggested fix. Keep the registry you just built so the function can never return nil:

```go
func getOrCreateFlagRegistry(config configuration.Configuration) *flagRegistry {
	if r := getFlagRegistry(config); r != nil {
		return r
	}
	created := &flagRegistry{flags: make(map[string]string)}
	configuration.Set(config, hiddenFlagRegistryKey, created, configuration.WithoutOverriding())
	if stored := getFlagRegistry(config); stored != nil {
		return stored // someone else won the race
	}
	return created
}
```

That also makes the nil branch at line 99 dead, so it can go.

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -38,14 +67,19 @@ func AddFeatureFlagsToConfig(
...
 				if m, ok := cached.(map[string]bool); ok {
-					return m[flagName], nil
+					if _, exists := m[flagName]; exists {
+						return m[flagName], nil
+					}
 				}
 			}
 
-			res, err := areFeaturesEnabled(c, engine, orgID, flags...)
+			allFlags := getAllFlagsToLookup(config)
```
basti-snyk: **[scenario-walker, concurrency-reviewer]** blocking.

One call per org only holds if the first reads are serial. `GetWithError` releases `ev.mutex` before invoking the default-value callback (`configuration.go:537`), so N goroutines cold-reading N different flag keys all miss the cache, all call `evaluateFlags`, and all write the same entry back.

Two passes measured this independently, with latency in the fake gateway:

```
10 flags, 10 goroutines, 20ms latency -> 10 API calls (PR claims 1)
20 flags, 20 goroutines, 20ms latency -> 20 API calls
```

At zero latency it collapses to 1, which is why the new test doesn't catch it.

Worth naming the regression precisely. The call count under concurrency is what it always was, but each of those N calls now carries the entire registry instead of the smaller per-registration group. Concurrent cold start costs more bytes than before this PR. snyk-ls hits exactly this shape, since every product's scan checks its own flag in parallel.

`singleflight` keyed on `cacheKey` is the cheaper fix. Holding a lock across the fetch would serialize every first flag read behind the slowest org lookup.

Separately on this line: everything else in the closure uses the callback's `c`, this reaches back to the captured outer `config`. It works only because `Clone` copies the registry pointer, which is a property of `Clone`'s current implementation rather than something this file should depend on. `getAllFlagsToLookup(c)` behaves the same and reads consistently. There is already a TODO at `app.go:399` describing this pattern as a bug.

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -18,13 +17,43 @@ import (
 var evaluateFlags = featureflaggateway.EvaluateFlags
 var errInvalidEvaluateFlagsResponse = errors.New("invalid evaluateFlags response")
 
+const hiddenFlagRegistryKey = "hidden_feature_flag_registry"
```
basti-snyk: **[security-escalator, architect, error-path-reviewer]** should-fix.

This parks internal Go state in the user-writable config keyspace. Three consequences, all verified:

- `HIDDEN_FEATURE_FLAG_REGISTRY=anything` in the environment turns into the startup panic at line 52.
- `AllKeys()` now returns `hidden_feature_flag_registry`, with a `*config_utils.flagRegistry` behind it. Anything walking and serializing config for debug output now hits a struct containing a `sync.Mutex`.
- `Clone()` shares the pointer, so parent and clone mutate one map. That is what the batching wants, but it is implicit and worth a comment at minimum.

The registry also gets none of the benefits of living in config. It hand-rolls a second lock, bypasses `defaultCache`, and cannot participate in `AddKeyDependency` invalidation.

If this were a package-level map in `config_utils` keyed by config instance, the env-var crash, the `AllKeys()` leak, and the whole new `configuration.Set` API would go away together.

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -60,6 +94,26 @@ func AddFeatureFlagsToConfig(
 	}
 }
 
+func getAllFlagsToLookup(config configuration.Configuration) []string {
+	registry := getFlagRegistry(config)
+	if registry == nil {
```
basti-snyk: **[rigor-reviewer, defensive-parser, idiom-checker]** should-fix.

Traced where this nil lands. `areFeaturesEnabled(c, engine, orgID)` with zero flags, `buildRequest` sends `"flags": []`, `results` comes back empty, the callback returns `false` and caches an empty map under `hidden_flags_<org>`. Every flag silently off, every subsequent read missing the cache and re-requesting, nothing logged. It is the one case where the batching optimization costs a request instead of saving one.

Right now this branch is unreachable, because line 52 panics first when the registry is nil. So it is either dead code or a silent kill switch, and it cannot be both. If you make `getOrCreateFlagRegistry` total per the comment above, delete lines 99-101. If you keep a defensive path, fall back to `[]string{flagName}` and log it.

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -18,13 +17,43 @@ import (
 var evaluateFlags = featureflaggateway.EvaluateFlags
 var errInvalidEvaluateFlagsResponse = errors.New("invalid evaluateFlags response")
 
+const hiddenFlagRegistryKey = "hidden_feature_flag_registry"
+
+type flagRegistry struct {
+	mu    sync.Mutex
+	flags map[string]string
```
basti-snyk: **[idiom-checker, necessity-hawk, error-path-reviewer]** should-fix.

`flags` is `map[configKey]flagName`, and nothing ever reads a key. Line 54 writes it, lines 107-113 read only the values, and then build a second `seen` map to dedupe those values back out.

It wants to be a set of flag names, and then `getAllFlagsToLookup` is one line against a helper already in the tree:

```go
type flagRegistry struct {
	mu    sync.Mutex
	flags map[string]struct{}
}

func getAllFlagsToLookup(config configuration.Configuration) []string {
	registry := getFlagRegistry(config)
	if registry == nil {
		return nil
	}
	registry.mu.Lock()
	defer registry.mu.Unlock()
	return utils.SortedMapKeys(registry.flags)
}
```

(`pkg/utils/mapkeys.go`.) That drops `sort.Strings` too. The diff removed the `slices` import while keeping `sort`, which is backwards on Go 1.26.

One trade-off to state out loud: a set means re-registering the same config key under a different flag name keeps both names. The cost is one extra flag in the batch request.

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -38,14 +67,19 @@ func AddFeatureFlagsToConfig(
...
+					if _, exists := m[flagName]; exists {
+						return m[flagName], nil
+					}
 				}
 			}
 
-			res, err := areFeaturesEnabled(c, engine, orgID, flags...)
+			allFlags := getAllFlagsToLookup(config)
+
+			res, err := areFeaturesEnabled(c, engine, orgID, allFlags...)
```
basti-snyk: **[consumer-advocate]** should-fix.

One shared registry means one shared failure. Every registrant's flags go into one request, so one rejected or failing flag fails every registered key rather than the caller's. Reproduced with two independent registrations and a stubbed failure:

```
flags sent on ext_a's read: [[extAFlag extBFlag]]
ext_a read err: check feature flags batch: assert.AnError general error for testing
```

An extension registering a flag the gateway rejects now breaks flag resolution for every other consumer in the process.

Two things worth confirming rather than assuming before all consumers get coupled into one call. Is there an upper bound on how many flags the endpoint accepts in one body? `pkg/apiclients/feature_flag_gateway/2024-10-15/spec.yaml:381` has no `maxItems`, so the spec cannot answer it. And which repos call `AddFeatureFlagsToConfig` today besides `pkg/app/app.go:415`?

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -81,6 +135,9 @@ func areFeaturesEnabled(
 	}
 
 	results := make(map[string]bool, len(flags))
+	for _, f := range flags {
+		results[f] = false
```
basti-snyk: **[necessity-hawk, defensive-parser]** should-fix.

Pre-seeding every requested flag to `false` makes "the API returned no evaluation for this flag" and "the flag is off" the same value. Combined with the new `exists` probe at line 74, that guess gets cached under `hidden_flags_<org>` and we stop asking.

The returned bool does not change, since a Go map already yields `false` for a missing key. What changes is that an API-side gap is now recorded as a decision. Before, the flag list in the cache key forced a fresh lookup on a mismatch.

If an absent evaluation is expected, log the requested flags that came back without one at debug. If it is not expected, do not cache it as an answer. Either way the current behavior is a feature quietly staying off with nothing in the log.

basti-snyk (commented):
## Review panel: 16 reviewer lenses

Automated review. I ran this PR through 16 reviewer lenses mined from a year of PR review history across nine Snyk repos. Each pass ran independently with the branch checked out, so it could run code rather than reason from the diff. Where a comment says verified, a pass actually ran it.

Submitting as comments only. The chat-side verdict was REQUEST_CHANGES, but that call belongs to you and the human reviewers, so nothing here changes the PR's review status.

Three things drove that verdict.

All 16 lenses raised the nil dereference at `feature_flag_gateway.go:52`, and 13 reproduced the panic. Three separate triggers, one of which is an unprivileged environment variable that crashes the CLI at startup. Details inline.

Five lenses found the new test is vacuous, three of them by mutation test. Revert the batching entirely and the test still passes, at 50 API calls.

Two lenses independently measured that concurrent first reads defeat the batching, at 10 and 20 goroutines. Under concurrency this sends more bytes than before the PR, since each of the N calls now carries the full registry.

The design is right, and several passes confirmed the sequential path genuinely collapses to one call. The problems are all at the edges.

### Two findings that could not be anchored to a diff line

`feature_flag_gateway.go:89`. `c.Set(cacheKey, res)` writes to whatever config the callback received, and `engineimpl.go:357` clones the config for every workflow invocation. Measured 1, 2, 3 cumulative API calls across three invocations, so the cache is discarded each time. This predates the PR, but it is the dominant remaining source of the traffic the PR is named after, and someone will look at the numbers after merge and wonder why they barely moved. Worth either writing to the outer config or filing a ticket linked from here. `app.go:399` already carries the honest version of this comment.

The PR body is the template plus a bot summary. This changes a cache key shape, adds three exported symbols, and changes when the flag API is called, so it needs a human-written note on what changes for consumers. Given the nil deref surfaces through the mock we ship, the template's "test your changes work for the CLI" step is worth ticking with the result written down.

### One question for outside this thread

A single flag read now evaluates every flag any workflow has registered for that org. Worth a sanity check with the gateway owners, since it changes what one client read discloses about which features that org is being probed for.

### Checked and clean

So you do not have to wonder: `go.mod` and `go.sum` untouched with `go mod tidy` a no-op, `gofmt -l` clean, no self-deadlock in the re-entrant `config.Get` inside the callback, no lost registrations on `extendedViper`, gateway errors correctly not cached as `false`, and late registration correctly re-batches. Nobody could run `-race` (no cgo in the sandbox), so please confirm `make test` was green locally.

Nit-level findings were left out of this review to keep the inline noise down.

pkg/configuration/configuration.go
```diff
@@ -34,6 +34,35 @@ const (
...
+// SetOpt configures the behavior of the Set function.
+type SetOpt func(*setOptions)
+
+// WithoutOverriding causes Set to store the value only when the key has no existing value.
+func WithoutOverriding() SetOpt {
+	return func(o *setOptions) { o.withoutOverriding = true }
+}
+
+// Set is a helper that extends Configuration.Set with functional options.
+func Set(config Configuration, key string, value interface{}, opts ...SetOpt) {
```
basti-snyk: **[8 lenses]** should-fix.

Three new exported symbols in the most-consumed package in the repo, for exactly one call site (`feature_flag_gateway.go:40`), with no tests. `rg 'WithoutOverriding|SetOpt' -g '*_test.go'` returns nothing.

Two concrete problems beyond the scope question.

The doc comment on line 44 says Set stores the value only when the key has no existing value. The fallback branch below does `config.IsSet(key)` then `config.Set(key, value)` as two separate calls with nothing holding them together, so two goroutines can both observe unset and both write. That matters here specifically: `getOrCreateFlagRegistry` leans on this option for a create-once guarantee this branch does not provide. On `*extendedViper` it works. On any other implementation two callers get two registries and half the registrations vanish.

The helper also reaches its actual guarantee by type-asserting down to an unexported type, so the atomicity is not part of the `Configuration` contract at all.

`configuration.Set(config, k, v)` sitting next to `Configuration.Set(k, v)` is also going to be a coin flip for readers. If it stays, `SetIfAbsent` says what it does.

AGENTS.md asks for public API to be tested and for speculative public API to be avoided. The simplest resolution is to keep this unexported in `config_utils` until a second consumer appears. If it stays exported it needs tests in `pkg/configuration` covering key unset, key already set, key set to a zero value, key set only via an alternative key, and key bound to an env var.

pkg/configuration/configuration.go
```diff
@@ -355,6 +384,32 @@ func (ev *extendedViper) clearCache(key string) {
...
 
+func (ev *extendedViper) set(key string, value interface{}, opts ...SetOpt) {
+	var so setOptions
+	for _, opt := range opts {
+		opt(&so)
+	}
+
+	ev.mutex.Lock()
+
+	if so.withoutOverriding && ev.viper.IsSet(key) {
```
basti-snyk: **[architect, concurrency-reviewer, defensive-parser, pattern-matcher]** should-fix.

`WithoutOverriding()` asks two different questions depending on the concrete type. Here it is `ev.viper.IsSet(key)`, raw. The fallback at line 60 is `config.IsSet(key)`, which calls `bindEnv` first and then falls through to `alternativeKeys`.

So on an `extendedViper` the option overwrites a value the same object reports as set. Verified:

```go
c.AddAlternativeKeys("primary", []string{"alt"})
c.Set("alt", "fromAlt")
c.IsSet("primary")   // true
c.Get("primary")     // "fromAlt"
configuration.Set(c, "primary", "overwritten", configuration.WithoutOverriding())
c.Get("primary")     // "overwritten", the doc comment on line 44 says this cannot happen
```

Worth defining "has an existing value" once. An unexported `isSetLocked(key)` holding the `IsSet` body, called from both `IsSet` and here, keeps the two from drifting further the next time someone extends `IsSet`.

pkg/configuration/configuration.go
```diff
@@ -355,6 +384,32 @@ func (ev *extendedViper) clearCache(key string) {
...
+	ev.mutex.Unlock()
+
+	if localStorage != nil && isPersisted {
+		//nolint:errcheck // breaking api change needed to fix this
+		_ = localStorage.Set(key, value)
+	}
+}
+
 // Set sets a configuration value.
 func (ev *extendedViper) Set(key string, value interface{}) {
```
basti-snyk, line 359: **[6 lenses]** should-fix.

`Set` is now `set` minus the options block, character for character, copied `//nolint:errcheck` included. Two copies of the storage, `persistedKeys` and `clearCache` sequence is two places to fix next time.

```go
// Set sets a configuration value.
func (ev *extendedViper) Set(key string, value interface{}) {
	ev.set(key, value)
}
```

pkg/local_workflows/config_utils/feature_flag_gateway_test.go
```diff
@@ -139,6 +142,87 @@ func Test_AddFeatureFlagGatewayToConfig(t *testing.T) {
...
+
+	// First flag read triggers the API call
+	result := config.GetBool("concurrent_key_0")
+	assert.True(t, result)
+
+	// All remaining flags should resolve from cache
+	for i := 1; i < flagCount; i++ {
+		configKey := fmt.Sprintf("concurrent_key_%d", i)
+		assert.True(t, config.GetBool(configKey), "flag %s should be resolvable", configKey)
+	}
```
basti-snyk, line 262: **[5 lenses, 3 by mutation test]** blocking.

The test is named `...AndBatching` and never asserts the batching. `apiCallCount` is checked once at line 212, and only to confirm it is 0 before any read. After the read at 216 and this loop, nothing looks at the counter again.

Three passes independently proved it is toothless by reverting the optimization (`allFlags := []string{flagName}` at line 80) and re-running this test unchanged:

```
=== RUN   Test_AddFeatureFlagsToConfig_ConcurrentRegistrationAndBatching
--- PASS (0.00s)
    apiCallCount after all reads = 50
```

Fifty round-trips, green test. The one behavior this PR exists to produce has no assertion anywhere in the suite. The real count today is 1, so this is a free line:

```go
assert.Equal(t, int32(1), atomic.LoadInt32(&apiCallCount),
	"all registered flags must resolve from a single batched API call")
```

Also worth noting the concurrency here is on registration, which is the half that works. Please add a case where the reads are concurrent, with a sleep in the fake gateway. That one goes red today, see the comment on line 80.

Three more cases nobody pins, all of which lens passes confirmed behave correctly right now: registering a flag after the org cache is warm (the `exists` probe at line 74 is the only thing keeping the simplified cache key correct), an org switch mid-session, and `areFeaturesEnabled` returning an error. Capturing the `flags` slice the stub receives and asserting it equals the 50 sorted names would pin the claim directly rather than through a counter.

pkg/local_workflows/config_utils/feature_flag_gateway_test.go
```diff
@@ -139,6 +142,87 @@ func Test_AddFeatureFlagGatewayToConfig(t *testing.T) {
...
+		flags []string,
+		org uuid.UUID,
+	) (*v20241015.ListFeatureFlagsResponse, error) {
+		atomic.AddInt32(&apiCallCount, 1)
+		trueVal := true
+		evals := make([]v20241015.FeatureFlagAttributes, len(flags))
+		for i, f := range flags {
+			evals[i] = v20241015.FeatureFlagAttributes{Key: f, Value: &trueVal}
+		}
+		return &v20241015.ListFeatureFlagsResponse{
```
basti-snyk: **[fixture-reviewer]** should-fix.

This anonymous `ApplicationvndApiJSON200` struct is now the fourth copy in this file, after lines 34-52, 88-99, and 283-292. The struct tags get retyped by hand every time, which is the kind of thing that rots silently. One `newFlagsResponse(evals []v20241015.FeatureFlagAttributes)` helper would cover all four.

Same for lines 146-147: `orgID` and `apiEndpoint` duplicate the existing `globalOrg` and `globalAPIEndpoint` at lines 65-66.

pkg/local_workflows/config_utils/feature_flag_gateway_test.go
```diff
@@ -139,6 +142,87 @@ func Test_AddFeatureFlagGatewayToConfig(t *testing.T) {
...
+			configKey := fmt.Sprintf("concurrent_key_%d", i)
+			flagName := fmt.Sprintf("concurrent-flag-%d", i)
+			AddFeatureFlagsToConfig(mockEngine, map[string]string{configKey: flagName})
+		}(i)
+	}
+	wg.Wait()
+
+	// All flags must be in the registry (no lost updates)
+	registry := getFlagRegistry(config)
+	registry.mu.Lock()
```
basti-snyk: **[duplication-hunter, diff-auditor]** should-fix.

`getFlagRegistry` can return nil, and this dereferences it straight away, which panics the whole test binary rather than failing this one test. `require.NotNil(t, registry)` first, and `require` is already imported in this file.

It is the same unguarded pattern as production line 52.

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -3,28 +3,80 @@ package config_utils
...
+	}
+	return nil
+}
+
+func getOrCreateFlagRegistry(config configuration.Configuration) *flagRegistry {
+	if r, ok := registries.Load(config); ok {
+		if reg, ok := r.(*flagRegistry); ok {
+			return reg
+		}
+	}
```
danskmt: Please reuse `getFlagRegistry` here as it contains the same logic

danskmt: i.e. 
```go
func getOrCreateFlagRegistry(config configuration.Configuration) *flagRegistry {
    if r := getFlagRegistry(config); r != nil {
        return r
    }
    // ...
```

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -3,28 +3,80 @@ package config_utils
...
+
+func (r *flagRegistry) allFlags() []string {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return utils.SortedMapKeys(r.flags)
+}
+
+// registries maps Configuration instances to their flag registry.
+// Stored outside the config keyspace to avoid env-var collisions and AllKeys() pollution.
+var registries sync.Map
```
danskmt: Is there any benefit by renaming here? Why to not use `sync.Map` direclty?

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -3,28 +3,80 @@ package config_utils
...
+		if reg, ok := r.(*flagRegistry); ok {
+			return reg
+		}
+	}
+	created := &flagRegistry{flags: make(map[string]struct{})}
+	actual, _ := registries.LoadOrStore(config, created)
+	if reg, ok := actual.(*flagRegistry); ok {
+		return reg
+	}
+	return created
```
danskmt: nit: I would improve the namings here, instead of `created` and `actual` could use `emptyFlagRegistry` and 'flagRegistry'

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -38,22 +90,36 @@ func AddFeatureFlagsToConfig(
...
-			if err := config.AddKeyDependency(cacheKey, configuration.ORGANIZATION); err != nil {
-				engine.GetLogger().Err(err).Msgf("failed to add dependency for %s", cacheKey)
-			}
-			c.Set(cacheKey, res)
 
+			res, ok := result.(map[string]bool)
+			if !ok {
+				return false, fmt.Errorf("check feature flags batch: unexpected result type")
+			}
 			return res[flagName], nil
```
danskmt: [From AI review] Instead of immediately returning this, could do:
```go
			if _, exists := res[flagName]; exists {
				return res[flagName], nil
			}

			result, err, _ = registry.sf.Do(cacheKey, func() (interface{}, error) {
				allFlags := registry.allFlags()
				retryRes, retryErr := areFeaturesEnabled(c, engine, orgID, allFlags...)
				if retryErr != nil {
					return nil, retryErr
				}
				c.Set(cacheKey, retryRes)
				return retryRes, nil
			})
			if err != nil {
				return false, fmt.Errorf("check feature flags batch: %w", err)
			}
			res, ok = result.(map[string]bool)
			if !ok {
				return false, fmt.Errorf("check feature flags batch: unexpected result type")
			}
			return res[flagName], nil
```
According to Claude:
"**narrow race window (low severity)**:

If a flag is registered while a singleflight call is already in-flight for the same org, that flag's callback could join the in-flight group and receive a result map that doesn't contain it. Line 123 (return res[flagName], nil) would then silently return false for that one invocation. The next read self-corrects because the cache-miss path (line 97) triggers a fresh fetch.

In practice this shouldn't happen because all AddFeatureFlagsToConfig calls occur during init, before any config reads. But if you want to be defensive, you could add an exists check after the singleflight returns (mirroring the cache check at line 97) and fall through to a solo fetch or Forget + retry.

Additionally: The retry itself is routed through singleflight, so if multiple late-registered flags detect the miss concurrently, their retries coalesce into a single API call rather than one per flag."

pkg/local_workflows/config_utils/feature_flag_gateway.go
```diff
@@ -3,28 +3,80 @@ package config_utils
...
+}
+
 func AddFeatureFlagsToConfig(
 	engine workflow.Engine,
 	configKeyToFlag map[string]string,
 ) {
 	config := engine.GetConfiguration()
-	flags := slices.Collect(maps.Values(configKeyToFlag))
-	sort.Strings(flags)
+	registry := getOrCreateFlagRegistry(config)
```
danskmt: nit: naming `registry` -> `flagRegistry`

pkg/local_workflows/config_utils/feature_flag_gateway_test.go
```diff
@@ -132,16 +152,162 @@ func Test_AddFeatureFlagGatewayToConfig(t *testing.T) {
...
+	}
+	wg.Wait()
+
+	for i, r := range results {
+		assert.True(t, r, "flag concurrent_read_key_%d should be true", i)
+	}
+
+	assert.Equal(t, int32(1), atomic.LoadInt32(&apiCallCount),
+		"concurrent reads for the same org should coalesce into a single API call")
 }
```
danskmt, line 474: [**Suggestion for new tests** - Correlates with one of my comments above https://github.com/snyk/go-application-framework/pull/716#discussion_r3949923730 ]

Given the following scenario: 

```
 during init, extensions A and B register their flags at slightly different times:

1. Extension A registers flag-a and reads it immediately, triggering a singleflight fetch for org abc. The batch sent to the API contains only [flag-a].
2. While that fetch is in-flight, extension B registers flag-b (it's now in the registry) and reads it. Its callback joins the in-flight singleflight.
3. The API responds with {flag-a: true}. Both callbacks receive this result.
4. Extension A finds flag-a in the map — done.
5. Extension B checks for flag-b — it's not in the map. It calls Forget on the singleflight key.
6. At the same moment, extension C registered flag-c late and also misses. It also calls Forget and enters sf.Do.
7. Only one of B or C actually executes the retry — singleflight coalesces them. The retry fetches [flag-a, flag-b, flag-c] (all registered flags), updates the cache, and both B and C get correct results.
8. Any subsequent read for any flag hits the cache.

Without the singleflight on the retry, steps 6-7 would produce two separate API calls instead of one.
```

Then we can have the following tests:
- Test_AddFeatureFlagsToConfig_LateRegistrationRetry — flag-b is registered and read while flag-a's singleflight fetch is in-flight. Without the retry, flag-b silently returns false. With it, flag-b detects it's missing from the result and triggers a retry that fetches all registered flags.
- Test_AddFeatureFlagsToConfig_LateRegistrationCoalescedRetry — same scenario but with two late flags (b and c). Verifies that both retries coalesce into a single API call (2 total, not 3). 

```go
func Test_AddFeatureFlagsToConfig_LateRegistrationRetry(t *testing.T) {
	var apiCallCount int32
	firstCallStarted := make(chan struct{})
	firstCallProceed := make(chan struct{})

	originalEvaluateFlags := evaluateFlags
	t.Cleanup(func() { evaluateFlags = originalEvaluateFlags })

	evaluateFlags = func(
		config configuration.Configuration,
		engine workflow.Engine,
		flags []string,
		org uuid.UUID,
	) (*v20241015.ListFeatureFlagsResponse, error) {
		call := atomic.AddInt32(&apiCallCount, 1)
		if call == 1 {
			close(firstCallStarted)
			<-firstCallProceed
		}
		trueVal := true
		evals := make([]v20241015.FeatureFlagAttributes, len(flags))
		for i, f := range flags {
			evals[i] = v20241015.FeatureFlagAttributes{Key: f, Value: &trueVal}
		}
		return newFlagsResponse(evals), nil
	}

	ctrl := gomock.NewController(t)
	t.Cleanup(ctrl.Finish)

	mockEngine := mocks.NewMockEngine(ctrl)
	logger := zerolog.Logger{}
	config := configuration.NewWithOpts()
	config.Set(configuration.API_URL, testAPIEndpoint)
	config.Set(configuration.ORGANIZATION, testOrgID)
	t.Cleanup(func() { registries.Delete(config) })

	mockEngine.EXPECT().GetConfiguration().Return(config).AnyTimes()
	mockEngine.EXPECT().GetLogger().Return(&logger).AnyTimes()

	AddFeatureFlagsToConfig(mockEngine, map[string]string{"key_a": "flag-a"})

	var wg sync.WaitGroup
	results := make([]bool, 2)

	wg.Add(1)
	go func() {
		defer wg.Done()
		results[0] = config.GetBool("key_a")
	}()

	<-firstCallStarted

	AddFeatureFlagsToConfig(mockEngine, map[string]string{"key_b": "flag-b"})

	wg.Add(1)
	go func() {
		defer wg.Done()
		results[1] = config.GetBool("key_b")
	}()

	time.Sleep(50 * time.Millisecond)
	close(firstCallProceed)

	wg.Wait()

	assert.True(t, results[0], "flag-a should be true")
	assert.True(t, results[1], "late-registered flag-b should be true via retry")
	assert.Equal(t, int32(2), atomic.LoadInt32(&apiCallCount),
		"expect 2 API calls: initial batch + retry for late-registered flag")
}

func Test_AddFeatureFlagsToConfig_LateRegistrationCoalescedRetry(t *testing.T) {
	var apiCallCount int32
	firstCallStarted := make(chan struct{})
	firstCallProceed := make(chan struct{})

	originalEvaluateFlags := evaluateFlags
	t.Cleanup(func() { evaluateFlags = originalEvaluateFlags })

	evaluateFlags = func(
		config configuration.Configuration,
		engine workflow.Engine,
		flags []string,
		org uuid.UUID,
	) (*v20241015.ListFeatureFlagsResponse, error) {
		call := atomic.AddInt32(&apiCallCount, 1)
		if call == 1 {
			close(firstCallStarted)
			<-firstCallProceed
		}
		trueVal := true
		evals := make([]v20241015.FeatureFlagAttributes, len(flags))
		for i, f := range flags {
			evals[i] = v20241015.FeatureFlagAttributes{Key: f, Value: &trueVal}
		}
		return newFlagsResponse(evals), nil
	}

	ctrl := gomock.NewController(t)
	t.Cleanup(ctrl.Finish)

	mockEngine := mocks.NewMockEngine(ctrl)
	logger := zerolog.Logger{}
	config := configuration.NewWithOpts()
	config.Set(configuration.API_URL, testAPIEndpoint)
	config.Set(configuration.ORGANIZATION, testOrgID)
	t.Cleanup(func() { registries.Delete(config) })

	mockEngine.EXPECT().GetConfiguration().Return(config).AnyTimes()
	mockEngine.EXPECT().GetLogger().Return(&logger).AnyTimes()

	AddFeatureFlagsToConfig(mockEngine, map[string]string{"key_a": "flag-a"})

	var wg sync.WaitGroup
	results := make([]bool, 3)

	wg.Add(1)
	go func() {
		defer wg.Done()
		results[0] = config.GetBool("key_a")
	}()

	<-firstCallStarted

	AddFeatureFlagsToConfig(mockEngine, map[string]string{
		"key_b": "flag-b",
		"key_c": "flag-c",
	})

	wg.Add(2)
	go func() {
		defer wg.Done()
		results[1] = config.GetBool("key_b")
	}()
	go func() {
		defer wg.Done()
		results[2] = config.GetBool("key_c")
	}()

	time.Sleep(50 * time.Millisecond)
	close(firstCallProceed)

	wg.Wait()

	assert.True(t, results[0], "flag-a should be true")
	assert.True(t, results[1], "late-registered flag-b should be true via coalesced retry")
	assert.True(t, results[2], "late-registered flag-c should be true via coalesced retry")
	assert.Equal(t, int32(2), atomic.LoadInt32(&apiCallCount),
		"expect 2 API calls: initial batch + one coalesced retry for both late flags")
}
```

CatalinSnyk (approved):

CatalinSnyk (approved):

PR #718, opened 2026-09-01
fix: IANDT-26: contributor capture code response parsing

jonnyowenpowell:
### Description

This PR follows #703 and resolves an issue where compressed responses from the legacy code flow were not handled correctly - it also tightens up the parsing of truncated (for performance reasons) responses to select the exact keys required rather than fuzzily matching.

### Checklist

- [x] Linted (`make lint`)

### DIFF
--- a/pkg/networking/middleware/contributor_capture/body.go
+++ b/pkg/networking/middleware/contributor_capture/body.go
@@ -2,6 +2,7 @@ package contributor_capture
 
 import (
 	"bytes"
+	"compress/gzip"
 	"io"
 	"net/http"
 )
@@ -28,8 +29,10 @@ func peekRequestBody(req *http.Request, maxBytes int64) ([]byte, error) {
 
 // peekResponseBody returns up to maxBytes of res.Body for parsing and splices what
 // it read back in front of the remainder, so res.Body still yields the whole body.
+// Deeproxy bodies arrive gzipped, so for that kind the peeked prefix is decoded
+// before it is returned.
 // fullyRead reports whether the body fit within maxBytes.
-func peekResponseBody(res *http.Response, maxBytes int64) (peeked []byte, fullyRead bool, err error) {
+func peekResponseBody(res *http.Response, kind endpointKind, maxBytes int64) (peeked []byte, fullyRead bool, err error) {
 	if res.Body == nil {
 		return nil, false, nil
 	}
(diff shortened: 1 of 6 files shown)


pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -43,10 +46,35 @@ func peekResponseBody(res *http.Response, maxBytes int64) (peeked []byte, fullyR
...
 
-	if int64(len(buf)) > maxBytes {
-		return buf[:maxBytes], false, nil
+	fullyRead = int64(len(buf)) <= maxBytes
+	if !fullyRead {
+		buf = buf[:maxBytes]
 	}
-	return buf, true, nil
+
+	if kind == endpointDeeproxyReport {
```
PeterSchafer, line 54: Question: instead of using the `kind` and therefore limiting the solution to deeproxy, shouldn't we use the content type?

jonnyowenpowell: I don't think it ends up making this more resilient. The reason we have to do decompression here is because the http.Transport didn't add the Accept-Encoding header itself, the legacy CLI did, and therefore it doesn't perform automatic decompression. The other responses may also use compression, but because this transport added the header, it decompresses the body before we read it. It's a property of the code path for legacy code scans (i.e. those which talk to deeproxy) more than it is a property of the request/response over the wire.

If we switch to reading the Content-Encoding header, which (as documented below) is stripped when automatic decompression is performed, then we have a choice to either _only_ support `gzip`, which doesn't do much in terms of guaranteeing that we can decompress some other response from another codepath, or we have to add more code to implement other compression standards we may encounter, which in case of newer formats like brotli would require additional third-party libraries.

```
> go doc net/http.Transport.DisableCompression
package http // import "net/http"

type Transport struct {
    // DisableCompression, if true, prevents the Transport from requesting compression
    // with an "Accept-Encoding: gzip" request header when the Request contains no
    // existing Accept-Encoding value. If the Transport requests gzip on its own and
    // gets a gzipped response, it's transparently decoded in the Response.Body.
    // However, if the user explicitly requested gzip it is not automatically
    // uncompressed.
    DisableCompression bool

    // ... other fields elided ...
}
```

```
> go doc net/http.Response.Uncompressed
package http // import "net/http"

type Response struct {
    // Uncompressed reports whether the response was sent compressed but was
    // decompressed by the http package. When true, reading from Body yields
    // the uncompressed content instead of the compressed content actually set
    // from the server, ContentLength is set to -1, and the "Content-Length" and
    // "Content-Encoding" fields are deleted from the responseHeader. To get the
    // original response from the server, set Transport.DisableCompression to true.
    Uncompressed bool

    // ... other fields elided ...
}
```

PeterSchafer (approved):

PR #723, opened 2026-09-03
feat: add TOON format selection and writer integration

robertolopezlopez:
### User description

Adds TOON output support to GAF.

- Adds the TOON config keys and `text/toon` MIME type.
- Supports TOON on stdout and in files using the existing writers.
- Treats TOON as structured output and routes raw `text/toon` data.
- Adds tests for writer selection and file output.

CLI flag validation will be handled in CLI-1827.

Related CLI PR: https://github.com/snyk/cli/pull/7211

### DIFF
--- a/pkg/local_workflows/output_workflow/constants.go
+++ b/pkg/local_workflows/output_workflow/constants.go
@@ -1,6 +1,9 @@
 package output_workflow
 
-import "github.com/snyk/go-application-framework/internal/presenters"
+import (
+	"github.com/snyk/go-application-framework/internal/presenters"
+	"github.com/snyk/go-application-framework/pkg/local_workflows/content_type"
+)
 
 const (
 	OUTPUT_CONFIG_KEY_JSON               = "json"
@@ -9,6 +12,8 @@ const (
 	OUTPUT_CONFIG_KEY_SARIF_FILE         = "sarif-file-output"
 	OUTPUT_CONFIG_KEY_HTML               = "html"
 	OUTPUT_CONFIG_KEY_HTML_FILE          = "html-file-output"
+	OUTPUT_CONFIG_KEY_TOON               = "toon"
+	OUTPUT_CONFIG_KEY_TOON_FILE          = "toon-file-output"
 	OUTPUT_CONFIG_TEMPLATE_FILE          = "internal_template_file"
 	OUTPUT_CONFIG_KEY_FILE_WRITERS       = "internal_output_file_writers"
 	OUTPUT_CONFIG_KEY_DEFAULT_WRITER_LUT = "internal_default_writer_mimetype_lut"
@@ -18,6 +23,7 @@ const (
 	SARIF_MIME_TYPE                      = presenters.ApplicationSarifMimeType
 	JSON_MIME_TYPE                       = presenters.ApplicationJSONMimeType
 	HTML_MIME_TYPE                       = presenters.ApplicationHTMLMimeType
+	TOON_MIME_TYPE                       = content_type.TOON
 )
 
 // DefaultTemplateFiles is an instance of TemplatePathsStruct with the template paths.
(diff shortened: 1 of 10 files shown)


pkg/local_workflows/output_workflow/constants.go
```diff
@@ -1,6 +1,9 @@
 package output_workflow
 
-import "github.com/snyk/go-application-framework/internal/presenters"
+import (
+	"github.com/snyk/go-application-framework/internal/presenters"
```
robertolopezlopez: check this

pkg/local_workflows/output_workflow/constants.go
```diff
@@ -18,11 +23,12 @@ const (
 	SARIF_MIME_TYPE                      = presenters.ApplicationSarifMimeType
 	JSON_MIME_TYPE                       = presenters.ApplicationJSONMimeType
 	HTML_MIME_TYPE                       = presenters.ApplicationHTMLMimeType
+	TOON_MIME_TYPE                       = content_type.TOON
```
octavian-snyk: can we use `presenters.AplicationTOONMimeType` instead?

robertolopezlopez: done

pkg/local_workflows/output_workflow/writer.go
```diff
@@ -167,6 +173,25 @@ func GetWritersFromConfiguration(config configuration.Configuration, outputDesti
 	return writerMap
 }
 
+// ValidateOutputConfiguration rejects incompatible output selections.
+func ValidateOutputConfiguration(config configuration.Configuration) error {
```
octavian-snyk: I noticed this expands GAF's public surface area, which should be beyond the scope of this specific PR for registering TOON output. Would you mind keeping this unexported, or removing it for now so we can discuss this type of contract changes separately, please?

getDefaultWriterMimeType seems to already have a silent precedence chain.

octavian-snyk: As a general guideline, we should avoid exporting new symbols from GAF unless strictly necessary, as we may need to maintain shims for them down the line. While consumers shouldn't need to pick up this specific change, it’s best to err on the side of caution. 😅

robertolopezlopez:
this belongs to cli-1826

> [**Missing Factory Registration**](https://github.com/snyk/go-application-framework/pull/723/changes#diff-36feaa65501342945cfb6a875563a8df7a8e4656fde322bb49ce32ab4f97cb1dR63-R89) 🟠 [major]
> In `NewLocalFindingsRenderer`, factories are registered for `NoneMimeType`, `DefaultMimeType`, and `ApplicationSarifMimeType`, but not for the newly added `ApplicationTOONMimeType`. When a consumer attempts to render `LocalFinding` models using the TOON format, `getImplementationFromMimeType` will find no entry in `p.templateImpl`, causing it to fall back to `NoneMimeType` which renders an empty string.
> 
> ```go
> templateImpl: map[string]TemplateImplFunction{
>     NoneMimeType: func() (*template.Template, template.FuncMap, error) {
>         localFindingsTemplate, err := template.New(NoneMimeType).Parse("")
>         if err != nil {
>             return nil, nil, err
>         }
>         return localFindingsTemplate, nil, nil
>     },
>     DefaultMimeType: func() (*template.Template, template.FuncMap, error) {
>         localFindingsTemplate, err := template.New(DefaultMimeType).Parse("")
>         if err != nil {
>             return nil, nil, err
>         }
> 
>         functionMapMimeType := getCliTemplateFuncMap(localFindingsTemplate)
>         return localFindingsTemplate, functionMapMimeType, nil
>     },
>     ApplicationSarifMimeType: func() (*template.Template, template.FuncMap, error) {
>         localFindingsTemplate, err := template.New(ApplicationSarifMimeType).Parse("")
>         if err != nil {
>             return nil, nil, err
>         }
> 
>         functionMapMimeType := getSarifTemplateFuncMap()
>         return localFindingsTemplate, functionMapMimeType, nil
>     },
> },
> ```

pkg/local_workflows/output_workflow/writer.go
robertolopezlopez: done

pkg/local_workflows/output_workflow/writer.go
```diff
@@ -107,7 +107,11 @@ func getDefaultWriter(config configuration.Configuration, outputDestination iUti
 	return writer
 }
 
-func GetWritersFromConfiguration(config configuration.Configuration, outputDestination iUtils.OutputDestination) WriterMap {
+func GetWritersFromConfiguration(config configuration.Configuration, outputDestination iUtils.OutputDestination) (WriterMap, error) {
```
octavian-snyk: Could we not modify the signature? This could be a breaking change for a consumer. We might need to do a breaking change eventually to hide --sarif, --toon, and --html from workflows that do not support UFM.

robertolopezlopez: it's either this or exporting the validator :-)

octavian-snyk (approved):

PR #765, opened 2026-09-10
feat: add TOON flags

octavian-snyk:
### Checklist

- [x] Tests added and all succeed (`make test`)
- [x] Regenerated mocks, etc. (`make generate`)
- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI
  1. Clone / pull the latest [CLI](https://github.com/snyk/cli) main.
  2. Run `go get github.com/snyk/go-application-framework@YOUR_LATEST_GAF_COMMIT` in the `cliv2` directory.
      - Tip: for local testing, you can uncomment the line near the bottom of the CLI's [`go.mod`](https://github.com/snyk/cli/blob/main/cliv2/go.mod) to point to your local GAF code.
  3. Run `go mod tidy` in the `cliv2` directory.
  4. Run the CLI tests and do any required manual testing.
  5. Open a PR in the CLI repo **now** with the `go.mod` and `go.sum` changes.
  - Once this PR is merged, repeat these steps, but pointing to the latest GAF commit on main and update your CLI PR.

Related PR: https://github.com/snyk/cli/pull/7255

---

> [!NOTE]
> **Low Risk**
> Flag registration and telemetry allowlist only; no changes to rendering or auth paths in this diff.
> 
> **Overview**
> Adds **TOON** as a first-class output option alongside JSON, SARIF, and HTML by registering `--toon` (stdout) and `--toon-file-output` (file) on the output workflow flag set.
> 
> Also adds `toon` and `toon-file-output` to the instrumentation **known flags** list so CLI usage of these options is categorized correctly. Init tests assert the new flags default to **off** and an **empty** file path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3266b3f8e1754a499d06b34cf3909d15261352e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #766, opened 2026-09-15
feat: TOON field selection

robertolopezlopez:
### Description

Adds default and `--full` TOON output for SCA and Secrets.

- context, feedback and scanner diagnostics
- `--full` adds SCA title, CVSS and upgrade details
- stdout and file output render the same TOON
- extraction errors and unfinished scans stop rendering
- to align with AXI behavior: replaced `results[]` with `sca[]` and `secrets[]`

### Checklist

- [x] Linted (`make lint`)
- [x] Test your changes work for the CLI -> https://github.com/snyk/cli/pull/7267

---

> [!NOTE]
> **Medium Risk**
> Changes CLI structured output format and flag semantics (`--toon`), which can break consumers expecting the old `results[]` envelope; issue grouping and primary-problem selection also affect SCA/license presentation.
> 
> **Overview**
> **Reworks UFM TOON output** from a raw `results[]` JSON dump to **compact, scanner-specific tables** (`sca[]`, `secrets[]`, generic `findings[]`) with org/project context, interaction ID, diagnostics (`*_error`, `*_hint`, `*_warning`), and a compact-mode hint to use `--toon=full`.
> 
> **`--toon` becomes a string flag** (`compact` / `full`, bare `--toon` → full) with validation; **full mode** renders SARIF-derived `runs[]` via a new `renderFullTOON` path instead of `toon.PrepareResults`. Rendering is **context-aware** through `RenderTemplateWithContext` and new template helpers (`mapTOONIssue`, `jsonFields`, issue conversion from test results).
> 
> **Issue extraction changes** in `testapi`: deterministic grouping order, safer keyless grouping, and **primary problem** preference (`snyk_vuln` over license/generic). Malformed SCA data and **unfinished scans** fail without writing partial output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0083347142365ba91217635662e9c895d1a569b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/presenters/templates/ufm.toon.tmpl
```diff
@@ -1,6 +1,180 @@
...
+    {{- $findings := getFindings (getContext) $result -}}
+    {{- template "toonDiagnostics" (dict "State" $state "Result" $result "Findings" $findings) -}}
+    {{- range $finding := $findings -}}
+      {{- with $finding.Attributes -}}
+        {{- if eq .FindingType "sca" -}}
+          {{- template "toonSCA" (dict "State" $state "Finding" $finding) -}}
+        {{- else if or (eq .FindingType "secret") (eq .FindingType "secrets") -}}
+          {{- template "toonSecrets" (dict "Envelope" $envelope "Attributes" .) -}}
+        {{- else -}}
+          {{- fail (printf "unsupported TOON finding type %q" .FindingType) -}}
```
PeterSchafer: Question: does it have to fail when there is a new findingstype? this seems too much limited on existing products.

robertolopezlopez: Thank you for pointing this out. I added a generic "toonFinding" template to render finding type, severity and findings.

https://github.com/snyk/go-application-framework/pull/766/commits/a292888e3ce1b5c681129359bb7cd243c3c9ad21

internal/presenters/templates/ufm.toon.tmpl
```diff
@@ -1,6 +1,195 @@
...
+    {{- $_ := set $versions "Values" (sortVersions $versions.Values) -}}
+  {{- end -}}
+  {{- $seen := dict -}}
+  {{- $rows := array -}}
+  {{- range $row := $state.SCA -}}
+    {{- if index $seen $row.id -}}{{- continue -}}{{- end -}}
+    {{- $_ := set $seen $row.id true -}}
+    {{- $_ := set $row "pkg" (print $row.pkg "@" (join (index $state.Versions $row.id).Values ",")) -}}
+    {{- if index $state.Fixable $row.id -}}{{- $_ := set $row "fixable" "yes" -}}{{- end -}}
+    {{- $rows = append $rows $row -}}
```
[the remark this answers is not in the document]
robertolopezlopez: this is just reproducing AXI behavior

robertolopezlopez:
@PeterSchafer ready for review again after your comments. I also have https://github.com/snyk/go-application-framework/pull/767 which depends on this change. I am starting work on the final CLI PR for TOON in parallel.

pkg/local_workflows/output_workflow/constants.go
```diff
@@ -23,6 +23,13 @@ const (
 	TOON_MIME_TYPE                       = presenters.ApplicationTOONMimeType
 )
 
+// OUTPUT_CONFIG_KEY_TOON_FULL expands SCA fields in both TOON stdout and file output.
+const OUTPUT_CONFIG_KEY_TOON_FULL = presenters.CONFIG_TOON_FULL
+
+// OUTPUT_CONFIG_KEY_TOON_FEEDBACK supplies the host's feedback invitation for both TOON writers.
+// The feedback field is an empty string when unset.
+const OUTPUT_CONFIG_KEY_TOON_FEEDBACK = presenters.CONFIG_TOON_FEEDBACK
```
PeterSchafer: Question: I think feedback is a feature of `snyk agent` do we really need this here?

robertolopezlopez: ok, I will drop that

pkg/local_workflows/output_workflow/constants.go
```diff
@@ -23,6 +23,13 @@ const (
 	TOON_MIME_TYPE                       = presenters.ApplicationTOONMimeType
 )
 
+// OUTPUT_CONFIG_KEY_TOON_FULL expands SCA fields in both TOON stdout and file output.
+const OUTPUT_CONFIG_KEY_TOON_FULL = presenters.CONFIG_TOON_FULL
```
PeterSchafer: Question: is `--full` the correct flag here? It is quite generic and makes sense in `snyk agent` but in this more general output case we might want `--toon-ful`

robertolopezlopez: checking with Matt together with the default + full output

internal/presenters/templates/ufm.toon.tmpl
```diff
@@ -1,6 +1,195 @@
...
+  {{- $file := "unknown" -}}{{- $line := 0 -}}
+  {{- if $attributes.Locations -}}
+    {{- $location := getSourceLocation (index $attributes.Locations 0) -}}
+    {{- $file = or $location.FilePath "unknown" -}}{{- $line = $location.FromLine -}}
+  {{- end -}}
+  {{- $row := dict "rule" (or $rule "secret") "severity" (or (toLowerCase (print $attributes.Rating.Severity)) "low") "file" $file "line" (number $line) -}}
+  {{- $_ := set .Envelope "secrets" (append (or .Envelope.secrets (array)) $row) -}}
+{{- end -}}
+
+{{- define "toonSCA" -}}
```
octavian-snyk: suggestion: consider moving SCA aggregation out of the template

Comparing this to `ufm.human.tmpl` in the same package: that template keeps almost no aggregation logic in the template itself — `getRemediationSummary`, `getSummariesFromIssues`, etc. do the dedup/grouping in Go, and the template just ranges over the already-shaped result. `toonSCA`/`$state` here take the opposite approach: `$state` threads a mutable dict (SCA rows, per-id versions, per-id fixability, fatal flag) through `toonDiagnostics`/`toonSCA` across a `range`, doing dedup and version-merging inline.

It works and is tested, but it's harder to unit test than a Go function would be (limited to golden-file diffs), and it's the only place in this template package using `dict` for cross-call mutable state rather than static single-purpose argument passing.

Would it be worth pulling the SCA dedup/version-merge/fixability computation into a small Go helper (mirroring `getRemediationSummary`'s shape) and having the template just format the result?

internal/presenters/templates/ufm.toon.tmpl
```diff
@@ -1,6 +1,195 @@
...
+{{- end -}}
+
+{{- define "toonFinding" -}}
+  {{- $id := "" -}}{{- with .Finding.Id -}}{{- $id = print . -}}{{- end -}}
+  {{- $attributes := .Finding.Attributes -}}
+  {{- $row := dict "id" $id "finding_type" (print $attributes.FindingType) "title" $attributes.Title "severity" (print $attributes.Rating.Severity) -}}
+  {{- $_ := set .Envelope "findings" (append (or .Envelope.findings (array)) $row) -}}
+{{- end -}}
+
+{{- define "toonSecrets" -}}
```
octavian-snyk: question: does secrets need its own template?

In `ufm.human.tmpl`, secrets findings get no special-casing — they flow through the same generic issue rendering as everything else (`GetSourceLocations`/`GetTitle` cover file/line/rule). Here `toonSecrets` is a fully separate template with its own field extraction (`getSecretsRule`, `getSourceLocation`).

Could secrets fold into the generic `toonFinding` path with `file`/`line` as optional fields, the way the human template handles it? Might simplify the finding-type dispatch in `main`.

internal/presenters/templates/ufm.toon.tmpl
robertolopezlopez: this would introduce test-specific logic in the go code, which we want to be agnostic. In the future if we introduce more commands to --toon, we want this support to be automatic.

Or you meant to move outside of the `{{- define "toonSCA" -}}`

internal/presenters/templates/ufm.toon.tmpl
robertolopezlopez: The problem is that the schema is different for sca and secrets. We display different fields and have different presentation logic. 

SCA: deduplicates by vulnerability ID, aggregates package versions and paths, and counts fixable IDs

Secrets: counts every finding and severity occurrence

internal/presenters/templates/ufm.toon.tmpl
robertolopezlopez: we do not want to move scan logic to helpers :-/

octavian-snyk: Comparing this to the other three renderers in this package: they all build on testapi.Issue (via getIssuesFromTestResult), not raw findings.

toonSCA instead works off getFindings (raw, ungrouped) and reimplements dedup/fixability itself via $state.

Was there a reason not to use `getIssuesFromTestResult` here? human/sarif/html renderers use this. why does TOON need special SCA and Secrets logic while all other renderers do not?

One real gap if we switch: issueBuilder currently keeps only the first package version it sees per issue, not the full merged list this template builds (pkg@4.17.4,4.17.10). If that multi-version detail matters for TOON, it'd need either a small addition to issueBuilder (accumulate distinct versions instead of first-wins) or a helper that ranges over Issue.GetFindings() to collect them — either way, on top of the shared grouping rather than reimplementing it.

pkg/local_workflows/output_workflow/writer.go
```diff
@@ -205,7 +205,7 @@ func getDefaultWriterMimeType(config configuration.Configuration) string {
 		return HTML_MIME_TYPE
 	}
 
-	if config.GetBool(OUTPUT_CONFIG_KEY_TOON) {
+	if config.GetString(OUTPUT_CONFIG_KEY_TOON) != "" {
```
octavian-snyk, line 208: Can we use `config.IsSet(OUTPUT_CONFIG_KEY_TOON)` instead?

internal/presenters/templates/ufm.toon.tmpl
robertolopezlopez: this is done, this morning

pkg/local_workflows/output_workflow/writer.go
robertolopezlopez: `IsSet` would still select TOON for `--toon=false` or `Set("toon", "")`, since it only reports that the key was touched. Instead I added a config default function (`toonDefaultValue` in `output_workflow.go`) that normalises the value to `""`, `compact` or `full` (handles `false`/`true`/`0`, rejects typos with an InvalidFlagOption error), so the `!= ""` check here is now the right test. Leaving this line as is.

robertolopezlopez:
> ## PR Reviewer Guide 🔍
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Breaking Flag Type Change**](https://github.com/snyk/go-application-framework/pull/766/changes#diff-34dd372fc437871957f1deb2bc8705325c3b72299fa9847525540d05b98e2cd3R29-R30) 🔴 [critical]
> `OUTPUT_CONFIG_KEY_TOON` was previously registered as a boolean flag (`outputConfig.Bool(...)`) and has been changed to a string flag (`outputConfig.String(...)` with `NoOptDefVal = "compact"`). Any downstream consumer (such as Snyk CLI extensions, IDE integrations, or external workflows) that queries `config.GetBool(output_workflow.OUTPUT_CONFIG_KEY_TOON)` will now always receive `false` when `--toon` or `--toon=full` is passed because `"compact"` and `"full"` cannot be parsed as booleans. Because this alters the public API contract under `pkg/`, the PR commit message should use `feat!:` or carry a `BREAKING CHANGE:` footer to ensure semantic-release issues a major version bump.
> 

as decided on slack

> 
> [**Faulty Falsy Flag Check**](https://github.com/snyk/go-application-framework/pull/766/changes#diff-bae3c9dcb4f60bdef6dfb9446977706cebaa5eb61be559e4b5a92d39358639e7R208-R210) 🟠 [major]
> `getDefaultWriterMimeType` checks `if config.GetString(OUTPUT_CONFIG_KEY_TOON) != ""` to select `TOON_MIME_TYPE`. If a consumer or user explicitly disables TOON output by configuring `config.Set(OUTPUT_CONFIG_KEY_TOON, false)`, setting `toon = "false"`, or passing `--toon=false`, `config.GetString(OUTPUT_CONFIG_KEY_TOON)` returns `"false"`. Because `"false" != ""` evaluates to true, `getDefaultWriterMimeType` enables the TOON writer instead of falling back to `DEFAULT_MIME_TYPE`. The condition should check that the string value is non-empty and not equal to `"false"`, or check `config.GetBool(OUTPUT_CONFIG_KEY_TOON)` alongside valid string modes (`"compact"`, `"full"`).
> 
> ```go
> if config.GetString(OUTPUT_CONFIG_KEY_TOON) != "" {
> 	return TOON_MIME_TYPE
> }
> ```

This is already addressed in https://github.com/snyk/go-application-framework/pull/765 so that --toon=false disables TOON and != "" is correct

pkg/local_workflows/output_workflow/writer.go
robertolopezlopez: `IsSet` would treat equally `--toon=false` and `--toon=0`. I do not think we want that

internal/presenters/templates/ufm.toon.tmpl
robertolopezlopez: refreshing this thread

- `toonSecrets` uses a different schema, giving one row per finding
- `toonFinding` gives one row per issue in findings

Putting them together would add more complexity

pkg/local_workflows/output_workflow/constants.go
robertolopezlopez: it is already out of toon output

robertolopezlopez:
added [an extra commit](https://github.com/snyk/go-application-framework/commit/bf5ea7d662232cfafc4183e7ba2bdcc4b090bdf6) to address 3 findings which AI found in my local machine:

> High — internal/presenters/templates/ufm.toon.tmpl:7: 🔴 bug: TOON sorts every issue but never applies severity-threshold or include-ignores; excluded and ignored findings remain visible. Use the existing issue filtering policy before rendering and cover both flags.
> Medium — internal/presenters/templates/ufm.toon.tmpl:6-7, internal/presenters/funcs.go:1089: 🟠 bug: TOON calls Findings twice; the second call uses context.Background() and its error is discarded, allowing a valid custom TestResult to return successful incomplete output. Build issues once with the invocation context and propagate the error.
> Medium — pkg/apiclients/testapi/issues.go:321-331: 🟠 bug: generated keys such as __finding_2 share the namespace with real finding keys, so a real key can merge with a keyless finding. Use a tagged comparable grouping key and add the collision regression.

internal/presenters/testdata/ufm/toon/mixed.json
```diff
@@ -38,7 +38,7 @@
               {
                 "id": "example-secret-rule",
                 "name": "Example secret rule",
-                "source": "snyk_secrets_rule"
```
PeterSchafer, line 41: Question: how comes this was changes?

robertolopezlopez: a mind lapse, removing change

internal/presenters/templates/ufm.toon.tmpl
```diff
@@ -1,6 +1,214 @@
...
+  {{- $attributes := .Attributes -}}
+  {{- $rule := or .Issue.GetProblemID $attributes.Title -}}
+  {{- $file := "unknown" -}}{{- $line := 0 -}}
+  {{- if $attributes.Locations -}}
+    {{- $location := getSourceLocation (index $attributes.Locations 0) -}}
+    {{- $file = or $location.FilePath "unknown" -}}{{- $line = $location.FromLine -}}
+  {{- end -}}
+  {{- $row := dict "rule" (or $rule "secret") "severity" (or (toLowerCase (print $attributes.Rating.Severity)) "low") "file" $file "line" (number $line) -}}
```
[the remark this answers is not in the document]
robertolopezlopez: dismiss: in secret findings, severity == Rating.Severity

internal/presenters/funcs.go
```diff
@@ -939,8 +992,67 @@ func applyInlineMarkdown(s string) string {
 
 func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinfo.RuntimeInfo) template.FuncMap {
 	defaultMap := template.FuncMap{}
+	defaultMap["dict"] = templateDict
```
PeterSchafer: Question: Is there a reason why this is added to the default function map? isn't this something only used for toon?

internal/presenters/funcs.go
```diff
@@ -939,8 +992,67 @@ func applyInlineMarkdown(s string) string {
...
+	defaultMap["getErrorTip"] = func(ctx context.Context) string {
+		if value, ok := ctx.Value(uitypes.ErrorTipKey).(string); ok {
+			return value
+		}
+		return ""
+	}
+	defaultMap["set"] = func(values map[string]any, key string, value any) map[string]any { values[key] = value; return values }
+	defaultMap["array"] = func(values ...any) []any { return values }
+	defaultMap["append"] = func(values []any, value any) []any { return append(values, value) }
+	defaultMap["sortVersions"] = func(values any) ([]string, error) {
```
PeterSchafer: Question: what is this used for? this is quite tricky logic to get right. I would argue we should not do it at all if we will most likely fail to do it correctly.

internal/presenters/funcs.go
```diff
@@ -972,6 +1084,13 @@ func getDefaultTemplateFuncMap(config configuration.Configuration, ri runtimeinf
 	defaultMap["getIssuesFromTestResult"] = func(testResults testapi.TestResult, findingType ...testapi.FindingType) []testapi.Issue {
 		return utils.ValueOf(testapi.GetIssuesFromTestResult(testResults, findingType))
 	}
+	defaultMap["getIssuesFromTestResultWithContext"] = func(ctx context.Context, result testapi.TestResult) ([]testapi.Issue, error) {
```
PeterSchafer: Question: Is this reasonable to have in a template? Isn't context an implementation concept of golang?

PeterSchafer: The context handling could easily be done in the background without exposing it to the template itself.

internal/presenters/funcs.go
```diff
@@ -1077,6 +1196,15 @@ func sortAndFilterIssues(config configuration.Configuration) func(issues []testa
 				}
 			}
 		}
+		if config.GetString(configuration.FLAG_SEVERITY_THRESHOLD) == "" {
```
PeterSchafer: Question: this seems like a duplication of the code above! What is the added value here?

internal/presenters/presenter_ufm.go
```diff
@@ -145,6 +146,10 @@ func (p *UfmPresenter) RegisterMimeType(mimeType string, implFactory TemplateImp
 }
 
 func (p *UfmPresenter) RenderTemplate(templateFiles []string, mimeType string) error {
+	return p.RenderTemplateWithContext(context.Background(), templateFiles, mimeType)
+}
+
+func (p *UfmPresenter) RenderTemplateWithContext(ctx context.Context, templateFiles []string, mimeType string) error {
```
PeterSchafer, line 152: Question: why not using the common pattern of functional options to provide the context?

internal/presenters/presenter_ufm.go
```diff
@@ -156,6 +161,8 @@ func (p *UfmPresenter) RenderTemplate(templateFiles []string, mimeType string) e
 		return err
 	}
 
+	localFindingsTemplate.Funcs(template.FuncMap{"getContext": func() context.Context { return ctx }})
```
PeterSchafer: Question: why is this here and not in defaultfunctions? but also why would a template require the context?

pkg/apiclients/testapi/issues.go
```diff
@@ -656,7 +674,7 @@ func (b *issueBuilder) processProblems(finding *FindingData) {
 			b.processCveProblem(&problem)
 		case "cwe":
 			b.processCweProblem(&problem)
-		case "secret":
+		case "secret", "snyk_secrets_rule":
```
PeterSchafer: Question: was there a change in the API?

pkg/apiclients/testapi/issues.go
```diff
@@ -486,6 +490,7 @@ type issueBuilder struct {
 	problemID            string
 	packageName          string
 	packageVersion       string
+	packageVersions      []string
```
PeterSchafer, line 501: Question: why is this required? does toon require data that wasn't shown in any of the existing renderers yet?

pkg/local_workflows/output_workflow.go
```diff
@@ -26,24 +28,45 @@ func InitOutputWorkflow(engine workflow.Engine) error {
 	outputConfig.String(output_workflow.OUTPUT_CONFIG_KEY_SARIF_FILE, "", "Write sarif output to file")
 	outputConfig.Bool(output_workflow.OUTPUT_CONFIG_KEY_HTML, false, "Print html output to console")
 	outputConfig.String(output_workflow.OUTPUT_CONFIG_KEY_HTML_FILE, "", "Write html output to file")
-	outputConfig.Bool(output_workflow.OUTPUT_CONFIG_KEY_TOON, false, "Print toon output to console")
+	outputConfig.String(output_workflow.OUTPUT_CONFIG_KEY_TOON, "", "Print toon output to console (compact or full)")
+	outputConfig.Lookup(output_workflow.OUTPUT_CONFIG_KEY_TOON).NoOptDefVal = "compact"
```
PeterSchafer, line 32: Question: why don't you set the default value when registering the string in the line above?

pkg/local_workflows/output_workflow.go
```diff
@@ -26,24 +28,45 @@ func InitOutputWorkflow(engine workflow.Engine) error {
...
+	}
+	switch strings.ToLower(strings.TrimSpace(fmt.Sprint(existingValue))) {
+	case "", "false", "0":
+		return "", nil
+	case "true", "compact":
+		return "compact", nil
+	case "full":
+		return "full", nil
+	}
+	return nil, fmt.Errorf("invalid value %v for --%s, expected compact or full", existingValue, output_workflow.OUTPUT_CONFIG_KEY_TOON)
```
PeterSchafer: Suggestion: Instead of using a generic error here, please re-use the existing invalid flag error https://docs.snyk.io/scan-fix-and-prevent/prevent/error-catalog#snyk-cli-0004 using error catalog errors

pkg/local_workflows/output_workflow.go
```diff
@@ -26,24 +28,45 @@ func InitOutputWorkflow(engine workflow.Engine) error {
...
+
 // outputWorkflowEntryPoint defines the output entry point
 // the entry point is called by the engine when the workflow is invoked
 func outputWorkflowEntryPoint(invocation workflow.InvocationContext, input []workflow.Data, outputDestination iUtils.OutputDestination) ([]workflow.Data, error) {
 	output := []workflow.Data{}
 
 	var finalError error
 	config := invocation.GetConfiguration()
+	if _, err := config.GetWithError(output_workflow.OUTPUT_CONFIG_KEY_TOON); err != nil {
+		return output, cli.NewInvalidFlagOptionError(err.Error())
```
PeterSchafer: See https://github.com/snyk/go-application-framework/pull/766/changes#r4038974381

internal/presenters/templates/ufm.toon.tmpl
PeterSchafer: I actually agree with @octavian-snyk on this topic. The toon renderer has other then other renderer a strong coupling to products rather then being data driven. This is problematic from a platform perspective.

And I don't see why this needs to be the case. UFM stands for "unified" and the renderer as of now removes the benefits of unification. With given timelines we might be forced to accept this as a first iteration but I think we are missing the overall goal for now.

PeterSchafer: > SCA: deduplicates by vulnerability ID, aggregates package versions and  paths, and counts fixable IDs
> Secrets: counts every finding and severity occurrence

This is something that should be covered by the `Issue` representation and not spill into the renderer

internal/presenters/funcs.go
robertolopezlopez: This would simplify quite much. The only problem is that, if we drop this, different reports for the same findings may differ. WDUT

robertolopezlopez: what about just using semver.Compare (if any) I was considering that actually

robertolopezlopez: I will just use slices.Sort for sake of simplicity

internal/presenters/presenter_ufm.go
robertolopezlopez: As the context is just used for the rendering and not the render state, I think keeping this func may be a good idea. Even though, I am moving the context out of the template. What do you think?

internal/presenters/funcs.go
robertolopezlopez: context moved out of the template

internal/presenters/funcs.go
robertolopezlopez: After digging, this was a confusion from my side. To avoid a regression in internal/presenters/presenter_ufm_toon_test.go (keeping unknown severity issues), but as TOON is not yet released this regression is not really valid. Thanks for pointing this out, I will remove this block

internal/presenters/presenter_ufm.go
robertolopezlopez: this was removed already

pkg/local_workflows/output_workflow.go
robertolopezlopez: adding value in the line above would add a not desired behavior. With `outputConfig.String(output_workflow.OUTPUT_CONFIG_KEY_TOON, "compact", ...)` if the user types `snyk test` -> the config would consider it as disabled and give the default value of "compact" -> the user would get TOON instead of what they wanted. `NoOptDefVal` is documented as "default value (as text); if the flag is on the command line without any options"

robertolopezlopez: example (vibe coded):

```go
package main

import (
	"fmt"

	"github.com/spf13/pflag"
)

const OUTPUT_CONFIG_KEY_TOON = "toon"

func parse(args ...string) {
	outputConfig := pflag.NewFlagSet("output", pflag.ContinueOnError)
	toon := outputConfig.String(OUTPUT_CONFIG_KEY_TOON, "compact", "Print toon output to console (compact or full)")
	outputConfig.Lookup(OUTPUT_CONFIG_KEY_TOON).NoOptDefVal = "compact"

	err := outputConfig.Parse(args)
	fmt.Printf("args=%q toon=%q err=%v\n", args, *toon, err)
}

func main() {
	parse()
	parse("--toon")
	parse("--toon=full")
}
```

output:

```bash
⋊> ~/g/go-application-framework on feat/CLI-1859 ◦ go run /tmp/pflag_toon_check.go        
args=[] toon="compact" err=<nil>
args=["--toon"] toon="compact" err=<nil>
args=["--toon=full"] toon="full" err=<nil>
⋊> ~/g/go-application-framework on feat/CLI-1859 ◦   
```

pkg/local_workflows/output_workflow.go
robertolopezlopez: good one, applying here and the next comment below

internal/presenters/templates/ufm.toon.tmpl
robertolopezlopez: with the last commits I have tried to change this

internal/presenters/funcs.go
robertolopezlopez: moved in 2eec477 to `getToonTemplateFuncMap`

pkg/apiclients/testapi/issues.go
robertolopezlopez: this is not up to date, removing it

pkg/apiclients/testapi/issues.go
robertolopezlopez: yes, this is required because we are grouping findings and showing all affected versions

robertolopezlopez: we=TOON

internal/presenters/templates/ufm.toon.tmpl
```diff
@@ -1,6 +1,108 @@
...
+    {{- $warningKey := print $scanner "_warning" -}}
+    {{- if eq $scanner "secrets" -}}{{- $warningKey = "secrets_hint" -}}{{- end -}}
+    {{- template "toonAppendMessage" (dict "Envelope" $envelope "Key" $warningKey "Message" $warnings.Message) -}}
+  {{- end -}}
+{{- end -}}
+
+{{- define "toonDocument" -}}
   {{- $first := true -}}
-  {{- range $key, $value := prepareUFMToon .TestResults -}}
+  {{- range $key, $value := . -}}
```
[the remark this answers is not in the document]
robertolopezlopez: dismiss

https://pkg.go.dev/text/template#hdr-Actions

`If the value is a map and the keys are of basic type with a defined order, the elements will be visited in sorted key order`

pkg/apiclients/testapi/issues.go
```diff
@@ -49,9 +49,19 @@ const (
 	// Value type: string
 	DataKeyRuleShortDescription = "rule-short-description"
 
+	// DataKeyIsUpgradable indicates whether an issue has an upgrade path.
+	// Value type: bool
+	DataKeyIsUpgradable = "is-upgradable"
+
+	// DataKeyUpgradeTarget is the first direct package upgrade target.
+	// Value type: Package
+	DataKeyUpgradeTarget = "upgrade-target"
```
octavian-snyk: Could we call ufm_helpers.GetFixAttributes + GetDirectPackageUpgradeTarget/GetDirectPackagePinTarget from toon_issues.go, the way SARIF does, instead of adding these fields to the shared Issue builder?

pkg/apiclients/testapi/issues.go
```diff
@@ -302,33 +323,24 @@ func (g *idBasedIssueGrouper) getUniqueKey(finding *FindingData) string {
...
-		groups[key] = append(groups[key], finding)
-	}
-
-	result := make([][]*FindingData, 0, len(groups))
-	for _, group := range groups {
-		result = append(result, group)
-	}
-
-	return result
+		// Keyless, ID-less findings must not be dropped; give each its own issue.
```
octavian-snyk: Can this ever happen? Is there a reason to add this in this PR?

pkg/apiclients/testapi/issues.go
```diff
@@ -649,8 +707,16 @@ func (b *issueBuilder) processProblems(finding *FindingData) {
 
 		switch discriminator {
 		case "snyk_vuln":
+			if !b.hasVulnerability {
```
octavian-snyk, line 655: why are we adding these changes?

pkg/apiclients/testapi/issues.go
robertolopezlopez: good idea, applied

pkg/apiclients/testapi/issues.go
robertolopezlopez: lines 710->719: as of now, toon supports just vulnerability or license problems.

`if !b.hasVulnerability {` -> select just the first vulnerability

further down: `if !b.hasVulnerability && !b.hasLicense {` -> a license problem can replace a general problem but not a vulnerability or the first license

Those changes give the precedence: vulnerability > license > generic problem

pkg/apiclients/testapi/issues.go
robertolopezlopez: kudos, dead code

pkg/apiclients/testapi/issues.go
```diff
@@ -52,6 +52,8 @@ const (
 	FindingTypeLicense = "license"
 )
 
+const dataKeyComponentVersions = "component-versions"
```
octavian-snyk: I noticed the string is also duplicated in `internal/presenters/toon_issues.go`. Would it make sense to move this logic entirely into `internal/presenters/toon_issues.go` so we can avoid introducing any new symbols here?

pkg/apiclients/testapi/issues.go
```diff
@@ -486,11 +489,14 @@ type issueBuilder struct {
 	problemID            string
 	packageName          string
 	packageVersion       string
+	packageVersions      []string
```
octavian-snyk: If we can move `dataKeyComponentVersions` to `internal/presenters/toon_issues.go`, I think we can also remove `packageVersions`. We can keep the shared `pkg/apiclients/testapi/issues.go` changes minimal and iterate on TOON separately as needed.

internal/presenters/funcs.go
```diff
@@ -939,8 +1003,34 @@ func applyInlineMarkdown(s string) string {
...
+	defaultMap["append"] = func(values []any, value any) []any { return append(values, value) }
+	defaultMap["runes"] = func(value string) []rune { return []rune(value) }
+	defaultMap["runeString"] = func(value []rune) string { return string(value) }
+	defaultMap["stringValue"] = func(value *string) string {
+		if value == nil {
+			return ""
+		}
+		return *value
+	}
+	defaultMap["number"] = func(value int) json.Number { return json.Number(strconv.Itoa(value)) }
```
octavian-snyk, line 1045: I am not entirely sure these are needed. Could we keep them in TOON for the time being, and move them to default once we confirm our last iteration requires these and would be useful to other consumers as well?

pkg/apiclients/testapi/issues.go
robertolopezlopez: the problem about moving `dataKeyComponentVersions` is that it is used here and I wouldn't see clean/logical having `issues.go` depending on `toon_issues.go` - don't you think?

pkg/apiclients/testapi/issues.go
robertolopezlopez: I wrote the same answer in `packageVersions` below:

The problem about moving `dataKeyComponentVersions` is that it is used here and I wouldn't see clean/logical having `issues.go` depending on `toon_issues.go` - don't you think?

internal/presenters/presenter_ufm_full_toon.go
```diff
@@ -0,0 +1,127 @@
...
+	delete(document, "$schema")
+	delete(document, "version")
+	if runs, ok := document["runs"].([]any); ok {
+		for _, run := range runs {
+			if run, ok := run.(map[string]any); ok {
+				delete(run, "automationDetails")
+			}
+		}
+	}
+	return document, nil
```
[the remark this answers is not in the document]
robertolopezlopez: yes this is in the contract

pkg/apiclients/testapi/issues.go
octavian-snyk: Agreed, issues.go shouldn't depend on toon_issues.go. That's not what I'm proposing.

issues.go drops packageVersions and dataKeyComponentVersions entirely. 

toon_issues.go builds the same list itself, reading each finding's own problem and package location, the way singleFindingComponent already does for a single finding. Neither package needs a new shared symbol.

Could this work?

internal/presenters/funcs.go
robertolopezlopez: moved to `getToonTemplateFuncMap`

pkg/apiclients/testapi/issues.go
octavian-snyk: You seem to already have this here: https://github.com/snyk/go-application-framework/pull/767/changes/bd51c8885c1edda0354fe5547b76a589cfade053#diff-923c25855547b91d56b9927487d06cc40313c54e63fe1b138f5b47911c676435

internal/presenters/funcs.go
```diff
@@ -1182,7 +1299,7 @@ func getFindingTypesFromTestResult(testResults testapi.TestResult) []testapi.Fin
 
 	// Add finding types derived from actual findings
 	// todo: this is potentially an expensive intermediate conversion to issues, which we could cache or optimize differently.
-	issues, err := testapi.NewIssuesFromTestResult(context.Background(), testResults)
+	issues, err := testapi.NewIssuesFromTestResult(ctx, testResults)
```
octavian-snyk, line 1302: Why not testapi.GetIssuesFromTestResult ? Could this also simplify the changes to pkg/apiclients/testapi/issues.go further?

robertolopezlopez: because GetIssuesFromTestResult creates a new context. And using NewIssuesFromTestResult we carry forward the context from the template rendering

robertolopezlopez: also, GetIssuesFromTestResult adds filtering and sorting -> we do not need that here

internal/presenters/presenter_ufm.go
```diff
@@ -156,6 +165,10 @@ func (p *UfmPresenter) RenderTemplate(templateFiles []string, mimeType string) e
 		return err
 	}
 
+	if mimeType == ApplicationTOONMimeType {
```
octavian-snyk, line 168: not necessarly an actionable comment for now; perhaps we should pick this in a later iteration.

just to keep in mind: it may be a good idea to split TOON Full MimeType from Compact TOON MimeType.

Perhaps:
`text/toon; schema=full`
`text/toon; schema=compact`

internal/presenters/funcs.go
```diff
@@ -1173,6 +1282,14 @@ func formatDatetime(input string, inputFormat string, outputFormat string) strin
...
 
 func getFindingTypesFromTestResult(testResults testapi.TestResult) []testapi.FindingType {
+	findingTypes, err := getFindingTypesFromTestResultWithContext(context.Background(), testResults)
+	if err != nil {
+		return findingTypes
+	}
+	return findingTypes
+}
+
+func getFindingTypesFromTestResultWithContext(ctx context.Context, testResults testapi.TestResult) ([]testapi.FindingType, error) {
```
octavian-snyk, line 1292: reading the body of the function before this, could we fold `getFindingTypesFromTestResultWithContext` directly into `getFindingTypesFromTestResult`? I'm not sure why there's a need to add the `getFindingTypesFromTestResultWithContext`.

internal/presenters/funcs.go
octavian-snyk: sorry, I don't think these were moved. 
I am not sure if `jsonStrings`, `jsonNumbers`, `getSourceLocation`, `runes`, `runeString`, and `number` have any consumer?

PR #770, opened 2026-09-16
fix: IANDT-402 contributors collection git extensions failure

jonnyowenpowell:
### Description

This PR implements a workaround for a bug in go-git where git extension casing that differs from the expectations hardcoded into the library cause it to fail to open git repositories. This problem was observed in the majority of tests in Azure Pipelines. This leads to a failure to emit contributor data.

A local build of the CLI has been tested in Azure Pipelines and confirmed to fix the issue.

### Checklist

- [x] Linted (`make lint`)


NikkiL-Snyk (approved):

internal/contributors/collect.go
```diff
@@ -199,14 +205,136 @@ func openRepositoryFast(path string) (*git.Repository, func() error, error) {
...
+func (s extensionTolerantStorage) Config() (*gitconfig.Config, error) {
+	cfg, err := s.Storage.Config()
+	if err != nil {
+		return nil, err
+	}
+	if cfg.Raw != nil {
+		cfg.Raw.RemoveSection("extensions")
+	}
+	return cfg, nil
+}
```
danskmt, line 272: Is it correct to just remove the whole "extensions" section here?

internal/contributors/collect.go
```diff
@@ -199,14 +205,136 @@ func openRepositoryFast(path string) (*git.Repository, func() error, error) {
 
 	st := filesystem.NewStorageWithOptions(repositoryFilesystem(gitDir), cache.NewObjectLRUDefault(), filesystem.Options{KeepDescriptors: true})
 
-	fast, err := git.Open(st, wtFs)
+	fast, err := git.Open(extensionTolerantStorage{st}, wtFs)
 	if err != nil {
 		_ = st.Close()
 		return repo, noopClose, nil //nolint:nilerr // use the slow repo if our fast open fails
 	}
 	return fast, st.Close, nil
```
danskmt, line 213: wrapping into **extensionTolerantStorage**, just an observation here: every repository scanned now has its extensions silently stripped, not just the Azure Pipelines ones.

internal/contributors/collect.go
```diff
@@ -199,14 +205,136 @@ func openRepositoryFast(path string) (*git.Repository, func() error, error) {
...
+// detectGitDir finds the git directory governing path the way
+// git.PlainOpenWithOptions with DetectDotGit does, returning it together with
+// the root of its worktree, which is empty for a bare repository. It is needed
+// because that discovery is what go-git's extension check rejects, so the git
+// directory has to be located without it.
+func detectGitDir(path string) (gitDir, worktreeRoot string, err error) {
+	dir, err := filepath.Abs(path)
+	if err != nil {
+		return "", "", err
+	}
```
danskmt, line 283: Since it reimplements go-git's git dir discovery logic, unit tests should have been added for this function, or even be defined somewhere else.

Please note:

- **No tilde expansion** - go-git calls path_util.ReplaceTildeWithHome(path) before resolving; this doesn't. If a scan path starts with ~, they'd behave differently.
- **No symlink handling for .git directories** - A .git that is a symlink to a directory would be followed by os.Stat (which follows symlinks), so this probably works, but the behavior is implicitly relying on os.Stat semantics rather than explicitly handling it.

PR #771, opened 2026-09-16
feat(machineid): add shared machine id resolution and persistence

basti-snyk:
### Description

Adds `pkg/machineid`, which resolves one machine identifier per machine so that the CLI and the Language Server report the same value. Today snyk-ls derives a per-process id from `machineid.ProtectedID("Snyk-LS")` that the CLI cannot reproduce, and the CLI only relays an environment variable without storing anything.

Resolution runs as a configuration default registered in `pkg/app.initConfiguration`, so every GAF consumer gets the same algorithm from one config lookup. Precedence: existing stored value, external channel (`INTERNAL_SNYK_CLIENT_MACHINE_ID`), OS identifier, opt-in legacy device-id file, then a generated lowercase UUIDv4. The winning source is stored next to the value. A stored value is never overwritten; `Reset` is the only way to replace one. The machine identifier has no defined format. No validation is applied. A value from any source is stored and reported exactly as supplied, whether that is a hardware serial, a hostname, a JAMF UDID or an Intune GUID.

Non-OS values are persisted to a product-neutral shared file and mirrored into `snyk.json`. OS-derived values are not written, since every product recomputes them identically. Writes take the storage lock, refresh, re-check, then set, so two cold starts converge on one value. Storage failures are non-fatal.

Eight commits: config keys, the package itself, app wiring, removal of format validation, trimming trailing whitespace from the legacy device-id file, and three follow-up fixes from review (crash-safe persistence ordering, shared-file source validation, and tighter shared-file permissions).

Design doc: https://snyksec.atlassian.net/wiki/spaces/EE/pages/5182357599

### Notes for the reviewer

Six things worth your attention:

1. Storage errors in the resolver are swallowed rather than logged. `DefaultValueFunction` has no logger parameter, so there is nowhere to log to without changing that signature.
2. `config.GetString` inside a default-value function recurses, because GAF re-runs a key's default on every lookup. The re-check goes through a scratch `configuration.NewInMemory()` instead. Worth a second opinion on whether that reads the storage you would expect.
3. `sharedFilePaths` and `osMachineID` are package-level vars purely as test seams.
4. The stale hostname and serial-number diagnostic from the design is not implemented here. This step is resolution only.
5. The legacy device-id file is the only source that gets any trimming. Trailing whitespace, including a trailing newline, is removed from the value it supplies before adoption, since a device-id file written by an earlier product typically ends with a newline as a file-format artifact rather than part of the identifier. This is parsing, not validation, and it does not imply the identifier has a format. Every other source, including the external channel, the shared file and the OS identifier, is still stored exactly as supplied, including any trailing whitespace it carries.

Emission of these fields on analytics events is not in this PR. It depends on the analytics-service API version in https://github.com/snyk/analytics-service/pull/181

### Checklist

- [x] Tests added and all succeed (`go test -race ./pkg/machineid/...` passes)
- [ ] Regenerated mocks, etc. (`make generate`) — no generated code affected
- [x] Linted (`make lint`) — 0 issues
- [ ] Test your changes work for the CLI — not done yet; CLI adoption is a follow-up


pkg/machineid/machineid.go
```diff
@@ -0,0 +1,226 @@
...
+// later lookups need no further I/O. Storage errors are swallowed; the resolved value is always
+// returned.
+//
+// The re-check after Refresh reads into a scratch in-memory Configuration rather than config
+// itself: config.GetString(configuration.MACHINE_ID) would re-invoke this very default value
+// function, since Configuration re-runs a key's default function on every lookup.
+func mirrorIntoStorage(config configuration.Configuration, id string, source Source) (string, Source) {
+	storage := config.GetStorage()
+	if storage == nil {
+		config.Set(configuration.MACHINE_ID, id)
```
basti-snyk, line 167: should this be persisted?

pkg/machineid/machineid.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if refreshedID := refreshed.GetString(configuration.MACHINE_ID); hasValue(refreshedID) {
+		id = refreshedID
+		source = Source(refreshed.GetString(configuration.MACHINE_ID_SOURCE))
+	} else {
+		//nolint:errcheck // a failed write here still leaves the correct value in config.Set below for this process
+		_ = storage.Set(configuration.MACHINE_ID, id)
+		//nolint:errcheck // a failed write here still leaves the correct value in config.Set below for this process
+		_ = storage.Set(configuration.MACHINE_ID_SOURCE, string(source))
+	}
```
basti-snyk, line 196: why?

pkg/machineid/machineid.go
basti-snyk: No, this only runs when there's no Storage behind the Configuration at all (see TestIntegration_NoStorageResolvesWithoutPersisting), so there's nothing to persist to. `config.Set` just updates this process's in-memory cache so repeat lookups return the same value instead of re-resolving every time.

Pulled this into a `persistInMemoryOnly` closure now, since the fix below needed the same in-memory-only fallback from the storage path too (3fdb37f).

pkg/machineid/machineid.go
basti-snyk: Lock-refresh-recheck so two processes starting cold at the same moment converge on one value instead of each writing its own generated id. `Lock` serializes writers, then `Refresh` re-reads storage under that lock in case another process already won and stored a value while this one was resolving its own candidate. If so, we adopt that value instead of overwriting it.

Also fixed as part of this review: write order here mattered more than I'd realized. `MACHINE_ID_SOURCE` is now written before `MACHINE_ID`, and a failed `MACHINE_ID_SOURCE` write now skips the `MACHINE_ID` write too, so a stored `MACHINE_ID` can never be observed without a matching source next to it (3fdb37f).

pkg/machineid/machineid.go
```diff
@@ -0,0 +1,256 @@
+// Package machineid resolves and persists a single machine identifier shared by every Snyk
```
PeterSchafer: Question: why is the machineid package in pkg and thereby exported ? It seems to be only useed as plumbing and could easily be internal for now so that we don't extend the public API contract more the necessary

PR #772, opened 2026-09-16
fix: IANDT-402 contributor capture large response body handling

jonnyowenpowell:
### Description

This PR modifies the method used for capturing data from requests/responses from obtaining a 64kb head of the body, which fails for large bodies, to streaming parsing the whole body until the desired key is found, avoiding loading the entire body into memory. This is important for cases like Open Source scans with large license policies, where the desired key is always after the large content.

### Checklist

- [x] Linted (`make lint`)


pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,236 @@
...
+// Close finishes the scan and waits for its result, so capture is complete for
+// every body a consumer closes, whether or not it read one.
+func (s *scanningReadCloser) Close() error {
+	s.drainForScan()
+	s.stopFeeding(nil)
+	err := s.underlying.Close()
+
+	select {
+	case <-s.done:
+	case <-time.After(scanSettleTimeout):
```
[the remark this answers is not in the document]
jonnyowenpowell: This is intentional, we don't want to wait forever.

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
 	"net/http"
+	"sync"
+	"time"
+
+	"github.com/snyk/go-application-framework/internal/contributors"
+)
+
+// maxScanBytes bounds how much of a body is fed to an extractor. Bytes past it
+// still reach the consumer, unscanned.
+const maxScanBytes = 8 << 20 // 8 MiB
```
NikkiL-Snyk, line 16: Seems a lot - is this based on examples you've seen? Would we want to scan something this big?

jonnyowenpowell: Snyk Code is sometimes very big - there's not a lot of penalty for us continuing to scan (i.e. if this was 50MB, we're not holding it in memory) it's just almost certainly useless CPU cycles at that point; we're not _aware_ of anything where the project ID is buried under that much other data. The reason for having it quite high is to avoid playing whack-a-mole with conservative limits that we keep having to increase.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -94,7 +94,7 @@ func (m *ContributorCaptureMiddleware) RoundTrip(req *http.Request) (*http.Respo
 		return res, err
 	}
 
-	m.completeRequestCapture(state, req, res)
+	m.startResponseCapture(state, req, res)
```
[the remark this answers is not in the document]
jonnyowenpowell: That is a danger of streaming parsing like this, but in reality the create response is necessary for polling to start, so there won't be a case where the polling starts before this response completes.

pkg/networking/middleware/contributor_capture/contributor_capture.go
```diff
@@ -112,10 +112,15 @@ func (m *ContributorCaptureMiddleware) beginRequestCapture(req *http.Request) (s
...
-		m.record(contributors.EntityTypeRevision, parseAIBomUploadRevisionID(m.requestBody(req)))
+		revisionID, err := scanRequestBody(req, maxScanBytes, extractAIBomUploadRevisionID)
+		m.recordResult(contributors.EntityTypeRevision, revisionID, err)
 		return state, false
 	case endpointTestCreate:
-		state.publishReport, state.publishReportKnown = parseCreateTestPublishReport(m.requestBody(req))
+		report, err := scanRequestBody(req, maxScanBytes, extractCreateTestReport)
+		if err != nil {
+			m.sink.RecordMiss(missReasonFor(err))
+		}
```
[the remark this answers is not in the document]
jonnyowenpowell: The sink is designed to handle this case, taking the first given miss.

pkg/networking/middleware/contributor_capture/body.go
danskmt: do we have any analytics about this? something that could be monitored, even if it surpasses the limit for example?

jonnyowenpowell: Yes in that case we'll emit a `capture_body_too_large` reason for contributor emit failure, which we have monitoring for.

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
-	}
+// runExtractor contains a panicking extractor, so it cannot take down the
+// request it was capturing from.
+func runExtractor[T any](extract func(r io.Reader) (T, error), r io.Reader) (result T, err error) {
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			var zero T
+			result, err = zero, errScanPanic
+		}
+	}()
```
danskmt, line 99: should we log / send instrumentation data about the error...?

jonnyowenpowell: `errScanPanic` will turn into a `capture_panic` reason for contributor emit failure. Do you think we should include a debug log with the error in the recover here? The reason I haven't used a lot of debug logs is that this process is invisible to the customer, so it's unlikely they'll ever know to provide us with the logs.

danskmt: I'm not sure, just thought that it might be a flow where we would like to have some logs in case of failures

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
+
+	"github.com/snyk/go-application-framework/internal/contributors"
+)
+
+// maxScanBytes bounds how much of a body is fed to an extractor. Bytes past it
+// still reach the consumer, unscanned.
+const maxScanBytes = 8 << 20 // 8 MiB
+
+// scanSettleTimeout bounds how long closing a body waits for its scan.
+const scanSettleTimeout = 5 * time.Second
```
danskmt, line 25: I wonder if we should have these "magic values" configurable somehow - if possible

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
+// wedged, and costs that capture rather than the consumer's Close.
+func (s *scanningReadCloser) Close() error {
+	s.drainForScan()
+	s.stopFeeding(nil)
+	err := s.underlying.Close()
+
+	select {
+	case <-s.done:
+	case <-time.After(scanSettleTimeout):
 	}
```
danskmt, line 138: From my AI reviewer:
"""
This comment overstates the guarantee: a wedged extractor is only bounded after drainForScan finishes, not on the live Read path (line 109-118), where feed can block a Read indefinitely on an unbuffered pipe write.
"""

pkg/networking/middleware/contributor_capture/body.go
jonnyowenpowell: The problem with putting them in configuration is that customers could change them, and we don't want that because it would make diagnosing problems harder!

pkg/networking/middleware/contributor_capture/body.go
danskmt, line 138: and also (from AI):
"""
`drainForScan()` runs unbounded/untimed before the settle timeout applies, so Close() on an abandoned large body can synchronously block on up to 8 MiB of network reads with no deadline.
"""

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
-	defer reader.Close()
+	s.remaining -= int64(len(b))
 
-	decoded, err := io.ReadAll(reader)
-	if err != nil && complete {
-		return nil, err
+	if overBudget || s.remaining == 0 {
+		s.stop(errScanBudget)
 	}
```
danskmt: From AI:
"""
`off-by-one`: a body exactly at budget drives remaining to 0 and gets reported as errScanBudget even though nothing was truncated. Should only fire on overBudget.
"""

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
+	reader    io.Reader
+	remaining int64
 }
 
-func (b *stitchedReadCloser) Close() error {
-	return b.underlying.Close()
+func (b *budgetReader) Read(p []byte) (int, error) {
+	if b.remaining <= 0 {
+		return 0, errScanBudget
+	}
```
danskmt, line 252: From AI:
"""
same off-by-one on budgetReader, used by scanRequestBody.
"""
that is:
"""
`off-by-one`: a body exactly at budget drives remaining to 0 and gets reported as errScanBudget even though nothing was truncated. Should only fire on overBudget.
"""

pkg/networking/middleware/contributor_capture/body.go
```diff
@@ -1,87 +1,249 @@
...
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+
+		id, err := runExtractor(extract, pipeReader)
+		// Make any further feeding fail fast rather than block.
+		_ = pipeReader.CloseWithError(io.EOF)
+		onResult(id, err)
+	}()
```
danskmt, line 63: From AI:
"""
`goroutine-per-response` spawned here having no lifecycle tied to whether the consumer ever calls Close() — leaks a goroutine+pipe per abandoned body.
"""

pkg/networking/middleware/contributor_capture/body.go
jonnyowenpowell: There's two ways to have something get blocked here. The underlying read blocking on a slow network, which would exist in the consumer without this middleware, and we deliberately don't impose a read deadline of our own, since some networks are legitimately slow and the HTTP transport's timeouts will handle that.

The other case, is if the extractor stops consuming, which blocks the consumer rather than the network. It's correct that scanSettleTimeout doesn't cover that, but we don't want to because it would truncate the response for the consumer. That would require an extractor that neither reads nor returns, but they're all using json.Decoder to repeatedly read, i.e. they never loop on things that aren't reads.

jonnyowenpowell: This is another place where we delegate to the transport's timeouts rather than impose separate timeouts on top.

pkg/networking/middleware/contributor_capture/body.go
jonnyowenpowell: Yeah this and the comment below are correct. They're obviously very unlikely to be hit, because it requires a request/response body that is _exactly_ 8MB, but it's still a bug so I'll fix it.

pkg/networking/middleware/contributor_capture/body.go
jonnyowenpowell: See above.

pkg/networking/middleware/contributor_capture/body.go
jonnyowenpowell: The justification for not adding more machinery here to handle that is that it's overly defensive, because we control the "consumer" (the extractor) here, so the contract is that those need to read to EOF or close, and they do.

danskmt (approved):

